PR #16, opened 2017-07-12
Move tab bar to the bottom

rmevans9:
This moves the tab bar to the bottom on Android.


housseindjirdeh:
I owe you one mate, thank you :tada:

PR #27, opened 2017-07-12
📝 Document development API keys

JasonEtco:
Make it clear (in the code) that these API keys are for development and not for the real app.


housseindjirdeh:
Cheers - I have it in the README so definitely makes sense to have it commented as well 🙌 

Don't forget to add yourself as a contributor to the README! :)

PR #33, opened 2017-07-12
Adding user bio into user profile. Closes #17

### LINKED ISSUE #17: User profile description is missing
When you open your profile, the user description is missing.

I'd love to work on this feature, let me know if I can do it.

Also, thanks for making this project, it's awesome!!!

alejandronanez:
# Adding the user bio to user profiles.

This was kinda tricky, basically what happens with profile components is that it has a fix height of 50% of the window but user bios can have different lengths (up to 160 chars). I think this is fine as an MVP approach and without going into _complicated_ calculations (checking in how many lines does the bio needs)

Here are a couple of screenshots of the logged in user + other users with bio and without bio.

# Updated Screens
## Logged in user
screen shot 2017-07-13 at 11 59 08 am

## User with no bio
screen shot 2017-07-13 at 11 57 12 am

## User with long bio
screen shot 2017-07-13 at 11 57 28 am


src/components/user-profile.component.js
```diff
@@ -32,6 +33,9 @@ export const UserProfile = ({
       />
       <Text style={styles.title}>{user.name}</Text>
       <Text style={styles.subtitle}>{initialUser.login}</Text>
+      <Text style={styles.bio}>
+        {get(initialUser, ['bio', 'length'], 0) ? initialUser.bio : user.bio}
```
alejandronanez: I added `lodash.get` in here, it's easier to check deeply into an object for an specific key by doing that.

alejandronanez: Also I think we should _merge_ `initialUser` and `user`, I think a _dumb_ user profile component should have no logic at all and should work for any user no matter if it's logged in or not. We can discuss this in another issue, I just didn't want to forget about this.

housseindjirdeh: I like `lodash.get`. We can consider adding the entire library if we need more utilities but for now `.get` is definitely useful :)

housseindjirdeh: So `initialUser` is the first user object that gets passed in as a prop when viewing a list of users (like the Followers screen for example). However that doesn't contain all the information needed for the user so another API call is made to get the full user data once the user profile screen is mounted. I pass `initialUser` in order to have the avatar and name show immediately while the call is still fetching.

Name probably is quite misleading - will definitely be more than happy to discuss renaming or restructuring this in a separate ticket <3

alejandronanez:
@housseindjirdeh yeah my solution felt kinda hacky tbh, the way the panels are made makes difficult to set a dynamic height on the first panel (you'd have to calculate the # of lines and set the height based on that) I really like to have the user bio under its picture but I think for the sake of progress we can create the bio in the second panel and move it to the top in another iteration :) sounds good?

alejandronanez:
Also, since we're not using user bio here, what do you think about making the black panel 40%/45% of the total height of the screen?

housseindjirdeh:
Agreed I do think having it underneath the picture would be _ideal_ - will love to discuss that further in a separate ticket. I'm definitely on board with keeping things simpler for now and having it in the second section :)

housseindjirdeh:
40%/45% instead of half the screen? I think that should be good - will love to see a screenshot of how it looks first :raised_hands:

alejandronanez:
Sweet, will work on it. 👌🏼 I'll send a PR tomorrow.

alejandronanez:
@housseindjirdeh about the 40% I think it doesn't look that good. Let's stick with 50% (at least for now)

housseindjirdeh:
Perfect that works for me, appreciate you giving it a shot 🙌

housseindjirdeh:
:tada: looks good to me!

housseindjirdeh:
Woops forgot to mention and it slipped - `react-native-elements` exposes a `hideChevron` prop to hide the right icon for list items :). Also I think its better to hide the bio entirely if not present so I just put up a [commit](https://github.com/gitpoint/git-point/commit/d6fdcbf8814fff63c22b68f4737cecdd82fe9be2) for that

PR #47, opened 2017-07-13
Fixing broken link to issue for Android version

swapnilmj:


housseindjirdeh:
Oooh thank you good catch, that was from before I migrated to the repo to the organization.

Thank you so much for reviewing @tomscholz 🙌

Issue #2, opened 2017-07-05
Android Version


housseindjirdeh:
Since this is built with React Native, it only makes sense to include an Android version.

Unfortunately, I haven't _once_ tried running the application on a Android emulator or device so there is definitely going to be quite a bit that needs to be done to have it working smoothly across both platforms.

rmevans9:
I love to look of this app but can't use it since I run Android... so I spent 5ish minutes to see if there is anything that is a major showstopper for this to run on Android. My findings:

- You need to hook up deep linking on Android.  Add this to your main activity in `AndroidManifest`
```
<intent-filter>
    <action android:name="android.intent.action.VIEW" />
    <category android:name="android.intent.category.DEFAULT" />
    <category android:name="android.intent.category.BROWSABLE" />
    <data android:scheme="gitpoint" />
</intent-filter>
```

- `react-native-search-bar` is iOS only. Would need to either switch to something more cross platform or provide an Android alternative.

So far that is all i have as I haven't had a chance to work around the search bar issue. The screens I have seen working do look good on Android though so thumbs up there!

housseindjirdeh:
@rmevans9 I can't thank you enough for this. I'm so happy to hear that screens are looking quite good on Android 🙌 . From the sounds of it, it doesn't look like there is _too much_ to do 😃 

I'm going to tackle this in the coming week and hopefully have it up and running in around a week or so. Will post my updates here (and of course: any kind of contribution is always welcome!!)

rmevans9:
I should probably point out I only saw like two screens because if the second issue pointed out.

housseindjirdeh:
Thank you, in that case you're right there are quite a few screens in the app so there may definitely be things to do in many of them (off the top of my head for example - I use `ActionSheetIOS` in a number of screens so will need to find a cross platform solution for that)

Two screens looking decent is still good news nevertheless! :)

rmevans9:
After commenting out `SearchBar` in the 4 screen files I have the app up and working. I am sure that if I triggered anything that used `ActionSheetIOS` it would fail miserably.

![image](https://user-images.githubusercontent.com/14151327/28098776-18e5c1ec-666d-11e7-85ab-9061ae6eaec4.png)

![image](https://user-images.githubusercontent.com/14151327/28098781-2074ebf4-666d-11e7-9cb2-6504961ae568.png)

[...]

PR #49, opened 2017-07-14
Android Version (beta)

patw0929:
As mentioned at https://github.com/gitpoint/git-point/issues/2#issuecomment-315032689, I've made a Android version. What I do were replacing the components `ActionSheetIOS` & `react-native-search-bar`, because which are only compatible with iOS.

I think there might be some bugs, and I need other contributors help to check the features are works like expected (and before) on Android & iOS.

Thanks for reviewing this :)
Very looking forward to see the Android version in the Google Play.


housseindjirdeh:
This is an **outstanding** PR 🎉 🎉 

I can't thank you enough for this. I'm going to take a good look at this tomorrow morning - test it on an Android device and we'll get it merged in. Once it's in, I'll spend some time going through the app  to make sure things are smooth (there may definitely be some more things here and there) as well as check for regression. I'm so excited that Android looks like it'll be ready *a lot* sooner than I imagined and it's all thanks to you and everyone else assisting <3

App.js
```diff
@@ -40,8 +46,10 @@ class App extends Component {
       secretKey: md5(DeviceInfo.getUniqueID())
     });
 
+    store = configureStore();
```
alejandronanez: I'm curious, any specific reason to do this in here instead inside `persistStore(...)`?

package.json
```diff
@@ -21,10 +22,12 @@
     "query-string": "^4.3.1",
     "react": "16.0.0-alpha.6",
     "react-native": "0.44.2",
+    "react-native-actionsheet": "^2.2.0",
     "react-native-communications": "^2.2.1",
     "react-native-device-info": "^0.10.2",
     "react-native-elements": "^0.9.3",
     "react-native-htmlview": "^0.9.0",
+    "react-native-material-design-searchbar": "^1.1.4",
```
alejandronanez: Any reason to add this dep in here? We already have a search bar in `react-native-elements` https://react-native-training.github.io/react-native-elements/API/searchbar/

src/components/search-bar.component.ios.js
```diff
@@ -0,0 +1,46 @@
+import React, { Component, PropTypes } from 'react';
+import RNSearchBar from 'react-native-search-bar';
```
alejandronanez: I have the same question, since we're using `react-native-elements` I think we should stick with that instead of start adding more deps into the project. You can find the component of `react-native-elements` in here react-native-search-bar

src/issue/screens/pull-merge.screen.js
```diff
@@ -68,9 +68,22 @@ class PullMerge extends Component {
...
   }
 
+  mergeMethodMessages = ['Create a merge commit', 'Squash and merge'];
+
+  showActionSheet = () => {
+    this.ActionSheet.show();
+  };
+
+  handlePress = (i) => {
+    if (i !== this.mergeMethodMessages.length) {
```
alejandronanez: What does `i` stand out for? I think it doesn't hurt to add a more descriptive name in here.

src/user/screens/profile.screen.js
```diff
@@ -84,10 +80,11 @@ class Profile extends Component {
       isPendingUser,
       isPendingOrgs,
       isPendingCheckFollowing,
-      navigation
+      navigation,
```
alejandronanez: Remove dangling coma here please.

.eslintrc
```diff
@@ -12,6 +12,10 @@
         "plugin:react/recommended"
     ],
     "rules": {
-        // overrides
+        "no-console": 0
```
alejandronanez: What about leaving `no-console` on and disable the line only when needed?

src/issue/screens/pull-merge.screen.js
alejandronanez: Something like `index` or `issue` for example.

package.json
housseindjirdeh: Yep we've discussed this in #2. So you'll notice that I already use `react-native-search-bar` instead of the RNE solution. That library provides a wrapper around a native iOS search bar instead of relying on a non-native solution.

We've discussed that it probably makes sense to use a separate library for Android and keep the iOS logic as native as possible. However I have two questions:

1. Is there any sort of actual native Android search bar with a react-native bridge we can use instead?
2. If 1 doesn't seem to exist - then that's okay. The last time I tried using the search bar with RNE - it wasn't the simplest to modify and use. I don't know if that's changed however. I think it's worth a quick check and if we think that using this imported library than I'm okay with that

src/components/search-bar.component.ios.js
housseindjirdeh: See my comment above too!

root.store.js
```diff
@@ -22,7 +31,20 @@ const getEnhancers = () => {
...
 
-export const configureStore = createStore(
-  rootReducer,
-  compose(getMiddleware(), ...getEnhancers())
-);
+
+export default function configureStore() {
+  const store = createStore(rootReducer, compose(getMiddleware(), ...getEnhancers()));
+
+  if (module.hot) {
```
housseindjirdeh: Just curious, what exactly is this `if` condition for?

root.store.js
```diff
@@ -22,7 +31,20 @@ const getEnhancers = () => {
...
+    module.hot.accept(() => {
+      store.replaceReducer(require('./root.reducer').default); // eslint-disable-line global-require
+    });
+  }
+
+  if (__DEV__) {
```
housseindjirdeh: Do we need to save the `store` and `reducer` to the global object?

package.json
alejandronanez: Is there something like an universal solution for search bar? I don't think it's a good idea to have split specific components for each platform that makes our code less mantainable. Also we don't want our search bar to depend on 2 different projects.
Hopefully that makes sense.

src/user/screens/profile.screen.js
housseindjirdeh: I'm not too picky about dangling comments. On the contrary - I [prefer them](https://medium.com/@nikgraf/why-you-should-enforce-dangling-commas-for-multiline-statements-d034c98e36f8).

I think it would be good to come to a consensus and enforce whichever way we want in the linter

.eslintrc
housseindjirdeh: I'm assuming this allows `console.logs` in the code? Agreed, I would prefer to see it enforced actually

.eslintrc
```diff
@@ -12,6 +12,10 @@
...
+    },
+    "globals": {
+        "__DEV__": true,
```
housseindjirdeh: 👍

housseindjirdeh: 👍

src/user/screens/profile.screen.js
alejandronanez: Oh wow those are some good reasons to do dangling commas. I'm down with that :D

root.store.js
alejandronanez: +1 on that question? What are you trying to do with this? I assume it is related with redux devtools but not 100% sure though.

alejandronanez:
I'm ok with what ever we choose for the search bar, the only restriction is that we should have only one implementation for iOS **and** Android, maintaining 2 components that do the same less maintainable and could ended up in a headache later down the road.

src/user/screens/profile.screen.js
patw0929: I think maybe we can set up eslint & check lint rule before commit in next step to keep the coding style quality 😄

root.store.js
patw0929: I copied these redux devtools related code from my own project, but it seems unnecessary here, so I removed.

root.store.js
patw0929: same reason, so I removed.

patw0929:
Hey folks, thanks for reviewing & discussing this PR!
Sorry for the lately reply, I just woke up 😅
I had rebased master branch commits, fixed some conflicts, and fixed the code review advices above.

After thinking, about the iOS & Android SearchBar implementation, I agree with @alejandronanez, the maintaining cost might be a problem in the future.

src/user/screens/profile.screen.js
alejandronanez: That's right, I was thinking about adding CircleCI or Travis in here, at least checking for eslint would be nice!

App.js
```diff
@@ -26,6 +26,10 @@ import DeviceInfo from 'react-native-device-info';
 // md5
 import md5 from 'md5';
 
+if (console) {
+  console.disableYellowBox = true; // eslint-disable-line no-console
```
alejandronanez: 👍

README.md
```diff
@@ -58,6 +58,7 @@ Feel free to send me feedback on [twitter](https://twitter.com/hdjirdeh) or [fil
 - `yarn run link` to link react-native dependencies
 - `yarn start` in one terminal tab to start the packager
 - `yarn run ios` in another tab to run the app in the iOS simulator
+- `yarn run android` in another tab to run the app in the Android device/emulator
```
alejandronanez: ❤️

ios/GitPoint.xcodeproj/project.pbxproj
```diff
@@ -36,6 +35,7 @@
 		2D02E4C91E0B4AEC006451C7 /* libReact.a in Frameworks */ = {isa = PBXBuildFile; fileRef = 3DAD3EA31DF850E9000B6D8A /* libReact.a */; };
 		2DCD954D1E0B4F2C00145EB5 /* GitPointTests.m in Sources */ = {isa = PBXBuildFile; fileRef = 00E356F21AD99517003FC87E /* GitPointTests.m */; };
 		3B4B2C15526143C288A991B2 /* libRNDeviceInfo.a in Frameworks */ = {isa = PBXBuildFile; fileRef = 651F5DC3BF71489BB193A32B /* libRNDeviceInfo.a */; };
+		48F49F6A1F19028D0012FAD6 /* libRNSearchBar.a in Frameworks */ = {isa = PBXBuildFile; fileRef = 48F49F671F19026D0012FAD6 /* libRNSearchBar.a */; };
```
alejandronanez: I'm just curious about all changes in this file, are all this changes auto generated? if so, what triggered them?

package.json
alejandronanez: Hey @patw0929! thanks for the quick changes!
So according to your (comment)[https://github.com/gitpoint/git-point/pull/49#issuecomment-315505647] I'm glad we should use une search bar solution that we can use in both iOS and Android. Do you have any suggestions?

ios/GitPoint.xcodeproj/project.pbxproj
patw0929: Because I added back the react-native-search-bar, which need added some files in Xcode.

package.json
patw0929: I think I will try react-native-elements' search bar later. Another choice might be the react-native-search-bar.

ios/GitPoint.xcodeproj/project.pbxproj
housseindjirdeh: Yep adding a native-bridged library will cause more than a few changes in native code base

src/user/screens/profile.screen.js
housseindjirdeh: Opened an issue to discuss #66

patw0929:
Oops, stuck in some replacing problems:

1. When replacing with `react-native-elements`' SearchBar, I have no idea how to call `onCancel` event, and how to invoke the `unFocus` behavior.

2. If using `react-native-searchbar`, the SearchBar seems has absolute position, and behind the tabs.

alejandronanez:
> Okay in that case I'm okay with splitting the logic to two separate components for now - we can try consolidating them to a single component in a later ticket, don't want you to spend too much time trying to make it work with a single component that isn't rendering too well

Yeah agree, priority should be release android app ASAP. I created an issue for it https://github.com/gitpoint/git-point/issues/72 @housseindjirdeh I'm not sure where should it live in our project, can you please put it in there?

Also @patw0929 thanks for the awesome work and sorry for all these iterations!

housseindjirdeh:
@patw0929 really sorry if you had some back and forths with this - and I can't thank you enough for this PR. Just want to ask if you're okay with keeping the search bar components separate (I definitely am if the RNE solution/`react-native-searchbar` isn't working too well) and we can merge this beauty in :raised_hands:

patw0929:
@housseindjirdeh Yes, I am ok with keeping search bar separated, and I think we can refine it in the next step. If everything works fine, please help to merge it 👍 Thanks!

ios/GitPoint.xcodeproj/project.pbxproj
alejandronanez: Thanks for clarification =]

housseindjirdeh:
Awesome @patw0929 :) 

Unfortunately some conflicts are present since we just merged #73, if you can please resolve them and I'll merge this baby in. (Don't hesitate to let me know if you would rather have me resolve them if you don't have the time :raised_hands: : )

alejandronanez:
Approving this since we already agreed on using 2 searchbars, keep console rules in eslint and better named variables. Just resolve merge conflicts @patw0929 and I think we should be good to merge.

PR #50, opened 2017-07-14
Forking

### LINKED ISSUE #41: Add forking
Add "fork" to Repository Actions

mikekavouras:
Introducing forking 🍴 (closes #41)

forking

## Behavior
* The "fork" option will always be present on someone else's repo.
* The repo will be forked to a user's personal account (can be expanded to support a user's organizations in another PR)

# Notes (of disappointment) 
GitHub API does not provide enough information to infer whether or not a user has forked a repo based on some origin repo.

### Attempt 1
**Repository Search**
`https://api.github.com/search/repositories?q=react-native%20user%3Amikekavouras%20fork%3Aonly%20in%3Aname`

The response object does not contain _any_ reference to the origin repo, so there's no way to tell exactly where this repository began.

### Attempt 2 
**User Repositories**  
`https://api.github.com/repos/${username}/${repo.name}`  

This solution assumes that the repo names are the same, which is an inaccurate assumption. If a user has a conflicting repo name, GitHub will automatically append a `-n` to the forked repo (e.g. `react-native-1`).

If there is something I overlooked, please feel free to collaborate.

## One more thing...

Issue #3 is more pronounced with this PR because of navigation from repo screen to repo screen. Navigating back will *not* reload the previous repository state. This is outside the scope of this PR.


src/repository/screens/repository.screen.js
```diff
@@ -266,6 +302,16 @@ const styles = StyleSheet.create({
   listTitle: {
     color: colors.black,
     fontFamily: 'AvenirNext-Medium'
+  },
+  overlay: {
```
housseindjirdeh: looks like this isn't being used at all, most likely forgot to remove it when you created a separate component :)

housseindjirdeh:
With regards to Issue #3, definitely out of the scope of this PR :)

src/repository/screens/repository.screen.js
mikekavouras: yeah. removed.

mikekavouras:
>"is there no simple way to know whether a user has already forked an existent repository?"

As far as I can tell, the only reliable way to know if a user fork exists is to use `/repos/:owner/:repo/forks`. However, a repo like react-native has 11k+ forks meaning **350+ paged responses**. _Very impractical_. The best idea I can come up with is some idea of, "where owner is ${user}". Something like `/repos/:owner/:repo/forks?owner=username`. Sadly, nothing like that exists.


>"- does the call pass and the user is navigated to the repository normally?"

Yes it does. GH essentially ignores the request and sends us back the already forked repo JSON. The server response is the same in both cases and the resolution on the front end is consistent.

housseindjirdeh:
> Something like /repos/:owner/:repo/forks?owner=username. Sadly, nothing like that exists.

Yeah that's _exactly_ what I was hoping for, but it is what it is :)

Cheers thank you so much mate, I'm so sorry I just merged another PR and there's a conflict in the README with the contributor section. If you can resolve that I'll merge this right in 🎉

mikekavouras:
@housseindjirdeh should be 👍 to merge

housseindjirdeh:
🎉 🎉 🎉 🎉 🎉

Issue #66, opened 2017-07-15
add precommit hook + enforce eslint rules


housseindjirdeh:
As mentioned, let's narrow down our eslint/formatting preferences in the codebase. We can use this thread for discussion.

I'm a huge fan of `prettier` - we can do a [`prettier-eslint`](https://gist.github.com/cadebward/c26e218220d653385d876f9a81308140) pre-commit hook? Thoughts? CC @alejandronanez @patw0929

patw0929:
`prettier` is good.
I used [`lint-staged`](https://github.com/okonet/lint-staged) + [`prettier`](https://github.com/prettier/prettier) before.

And which eslint config base should we use? Or just custom rules?

patw0929:
Just a test, I made a lint-staged + prettier branch:
https://github.com/patw0929/git-point/tree/feature/prettier-n-eslint

I added some eslint rules. Including the import lint, so I think that removing babel resolve module alias should be better. Although it's convenient, but sometimes it confuse me, "Is this import a packages or actual path?".

Btw, it might be large commits, because we need fix many lint warnings & errors after enabling eslint.

alejandronanez:
@housseindjirdeh @patw0929 
+1 on travis, that's great.
About prettier I have never used but I want to, so, go for it, I remember on my last a project had it and prettier was enforcing something like:
`<Foo foo={1} bar={2} baz={3} />`
That's not so good for diffing/reading purposes, we should enforce:
```js
<Foo
    foo={1}
    bar={2}
    baz={3}
/>
```
that being said, go ahead with prettier =]

Is it possible to not format all files at once? Maybe it would be better to do it incrementally, let's say, I modify `path/to/foo.js` so I apply `prettier` to that particular file, it would be easier to track down changes. Thoughts?

housseindjirdeh:
@alejandronanez so with a prettier plugin in your editor you should be able to format single files at a time

Thanks again @patw0929 for adding this to project 🙌

PR #73, opened 2017-07-15
Prettier, ESLint and Travis CI

patw0929:
Related: #66 

- Added prettier & eslint hook via `lint-staged` & `pre-commit`.
- Based on `eslint-config-airbnb`, customize the eslint rules.
- Fixed all the current files which cannot pass the eslint rules.
- Added travis.yml to run eslint on Travis CI. (Need to set travis up by getpoint's github account)

Any suggestion?
About the eslint rules, I pick up rules from my projects, and it might not suitable for everyone's habits. Need to be discuss :)


App.js
```diff
@@ -1,68 +1,60 @@
...
   render() {
-    if (!this.state.rehydrated)
+    if (!this.state.rehydrated) {
       return (
         <View style={styles.logoContainer}>
-          <Image
-            style={styles.logo}
-            source={require('./src/assets/logo-black.png')}
-          />
+          <Image style={styles.logo} source={require('./src/assets/logo-black.png')} />
```
alejandronanez: I'd like to enforce 
```js
<Image
    style={styles.logo}
    source={require(....)}
/>
```
instead of a _one liner_ do you think we can do that with prettier?

App.js
```diff
@@ -1,68 +1,60 @@
...
-import { colors } from 'config';
-
-//Routes
-import { GitPoint } from './routes';
-
-// Redux Store
-import { configureStore } from './root.store';
 import { Provider } from 'react-redux';
-
+import { AppRegistry, AsyncStorage, Image, StyleSheet, View, LayoutAnimation } from 'react-native';
```
alejandronanez: Same comment for the multi line vs one liner here.

alejandronanez:
The only thing I think we should enforce (at least for now) is the multi line in imports and react components. If you have a component with more than 3 props in a one liner it could ended up being difficult to read + the diff will be harder.
Besides that, looks good! Thanks for your contribution @patw0929 :)

patw0929:
@alejandronanez Yes, it can do it. It based on `print-width`, default width is `80`.

Ran prettier again, and also rebased master branch.

.babelrc
```diff
@@ -4,27 +4,4 @@
     "flow"
   ],
   "retainLines": true,
-  "plugins": [
-    [
-      "module-resolver",
-      {
-        "root": [
```
alejandronanez: Is there a way to keep this? relative imports is a PITA in a big project like this.

housseindjirdeh: Agreed - however if it's bothering you @patw0929 because of the linter I'm happy to compromise and remove it.

patw0929:
Thanks for reviewing this! 👍 
Already added back the module resolver alias, including the import path.

I hope that I am not messing up the functions & behaviors. 😀

housseindjirdeh:
Thanks so much homies :star: We can't thank you enough for this @patw0929!

package.json
```diff
@@ -47,14 +62,20 @@
...
+    "eslint-config-airbnb": "~15.0.1",
+    "eslint-config-airbnb-base": "~11.2.0",
+    "eslint-import-resolver-babel-module": "^3.0.0",
+    "eslint-plugin-flowtype": "^2.35.0",
     "eslint-plugin-import": "^2.2.0",
-    "eslint-plugin-jsx-a11y": "^4.0.0",
-    "eslint-plugin-react": "^6.10.0",
-    "eslint-plugin-react-native": "^2.2.1",
+    "eslint-plugin-jsx-a11y": "^5.0.1",
+    "eslint-plugin-react": "^7.1.0",
```
Antoine38660: @patw0929 did you choose packages "fixed" versions for:
- `eslint-config-airbnb`
- `eslint-plugin-jsx-a11y`
- `eslint-plugin-react`

for any good reason ?

I'm using Atom 1.18.0 and the ESLinter returns an error each time I try to modify a file. After updating these packages versions : https://github.com/Antoine38660/git-point/commit/cfdd16d8786ec8ae7ce32f31a496644b5c0c0060 linter in Atom works again.

I wanted your opinion before submitting a new PR 😄

alejandronanez: That's a good question. I think fixed versions doesn't really matter since we're using yarn. Maybe it's an Atom bug?

Antoine38660: Yes, probably an Atom bug with these specifics versions (I have exactly this error https://github.com/airbnb/javascript/issues/1432 but inside the Atom console).

Waiting for @patw0929 answer. If there is no contraindication, then I would submit the update.

patw0929: Sorry to keep you waiting. I just woke up.
Not tested, but I think might be ok to use newer version. 👍

PR #74, opened 2017-07-15
update contributors badge + deps

alejandronanez:
Auto update for contributors were broken because we were using an `` instead of markdown, this should fix it + I ran `yarn install` and got some updates.

The readme will ended up looking like this
screen shot 2017-07-15 at 11 50 16 am I think there's no way to update automatically the contributors badge if we stick with HTML

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,3 +1,6 @@
+[![All Contributors](https://img.shields.io/badge/all_contributors-12-orange.svg?style=flat-square)](#contributors)
+[![PRs Welcome](https://img.shields.io/badge/PRs-welcome-brightgreen.svg?style=flat-square)](http://makeapullrequest.com)
+
 <p align="center">
   <a href="https://gitpoint.co/">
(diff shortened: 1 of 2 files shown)


housseindjirdeh:
Beautiful - thanks so much for pinpointing the issue. We can live with left aligned badges :joy:

README.md
```diff
@@ -1,3 +1,6 @@
+[![All Contributors](https://img.shields.io/badge/all_contributors-12-orange.svg?style=flat-square)](#contributors)
+[![PRs Welcome](https://img.shields.io/badge/PRs-welcome-brightgreen.svg?style=flat-square)](http://makeapullrequest.com)
```
housseindjirdeh: I think we can link to `CONTRIBUTING.md` here. Happy to have a snippet in that file to link to `makeapullrequest` for people who would like to know how to make a pull request

PR #79, opened 2017-07-15
Center "Sign In" button on welcome screen

RolfKoenders:
Hi! First of all, thanks for this awesome app! Im using it to get started with react-native.

When i started up the project for the first time i noticed that the "Sign in" button is not centered on the welcome screen. 

Not centered button

I fixed it by removing the `right` offset.

PR #60, opened 2017-07-15
Syntax highlighting. Closes #26

Antoine38660:
As discuted in #26 I used the [react-native-syntax-highlighter](https://github.com/conorhastings/react-native-syntax-highlighter) lib.

Result:

(I finally fixed the bug with scrolling)

Cheers 😄

PR #80, opened 2017-07-16
Syntax highlighting in PR Diff

Antoine38660:
> Will be great to have syntax highlighting on `PullDiffScreen` for pull requests

As @housseindjirdeh said here : https://github.com/gitpoint/git-point/pull/60#issuecomment-315512222 

So this PR is about this integration.  
Screenshots:
Screenshots...

Because of how each line are generated (separately), there is a "Syntax Highlighter" on each line. I did some performance tests to show you that it's not very slower than before (but, yes it is! Especially in case of a lot of line diff).

| | Before | Now | |
|:-:|:-:|:-:|:-:|
| PR #49 | 2480 ms | 4090 ms | +64% |
| PR #79 | 700 ms | 1010 ms | +27% |

_Tested on my iPhone 6S and a "manual chronometer" (human imprecision)_

### DIFF
--- a/src/components/code-line.component.js
+++ b/src/components/code-line.component.js
@@ -1,12 +1,12 @@
 // @flow
 /* eslint-disable no-nested-ternary */
 
-import React from 'react';
+import React, { Component } from 'react';
 import { View, Text, StyleSheet, Dimensions } from 'react-native';
-
+import SyntaxHighlighter from 'react-native-syntax-highlighter';
+import { getLanguage } from 'lowlight';
+import { github as GithubStyle } from 'react-syntax-highlighter/dist/styles';
 import { colors, normalize } from 'config';
-
-type Props = { newChunk: boolean, change: Object };
 
 const styles = StyleSheet.create({
   container: {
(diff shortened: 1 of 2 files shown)


Antoine38660:
Once you merge #81 and once I rebase `master` CI should pass 😃

housseindjirdeh:
This is awesome @Antoine38660 🎉 

Unfortunately the diff screen can already be kind of slow with large diff files since we currently render line by line. PR #49 slows numbers down significantly because we introduced `remote-redux-devtools` there (opened an issue for this #82). Comment out [these lines](https://github.com/gitpoint/git-point/blob/master/root.store.js#L11-L22) and you'll notice more normal speeds (and I assume the difference with including SyntaxHighlighter shouldn't be too bad).

src/components/code-line.component.js
```diff
@@ -63,44 +63,79 @@ const styles = StyleSheet.create({
...
+
+          <View
+            style={[
+              styles.codeLineContainer,
+              newChunk && styles.newChunkLineContainer,
+              change.type === 'add' && styles.addLine,
+              change.type === 'del' && styles.delLine,
+            ]}
+          >
+            {(newChunk || !this.isKnownType(language)) &&
```
housseindjirdeh: should this block of content show with `newChunk` _or_ it's an unknown language? Is there a possibility that line 123 will render as well 134 if `newChunk === true` and `this.isKnownType(language) === true`? I might be wrong here so please correct me if I am.

Antoine38660: `isKnownType(language)` function checks if it isn't a `newChunk` (line 74). So if `newChunk === true`, `this.isKnownType(language)` can't be `true` !

housseindjirdeh: Ah perfectttt - awesome thanks so much matey 👊 

This is such a sweet PR 🎉

PR #81, opened 2017-07-16
Fixed merge bugs

patw0929:
Hey @housseindjirdeh , really sorry for the late reply. 
I found you've merged #49, however there have several eslint errors & javascript errors. 
So I open this PR, trying to solve them. Please help to review this again, thanks!

### DIFF
--- a/package.json
+++ b/package.json
@@ -56,6 +56,8 @@
     "redux-logger": "^2.7.4",
     "redux-persist": "^4.3.1",
     "redux-persist-transform-encrypt": "^1.0.2",
+    "remote-redux-devtools": "^0.5.12",
+    "remote-redux-devtools-on-debugger": "^0.7.1",
     "redux-thunk": "^2.2.0"
   },
   "devDependencies": {
@@ -76,8 +78,6 @@
     "eslint-plugin-react-native": "^2.3.2",
     "flow-bin": "^0.47.0",
     "jest": "18.1.0",
-    "remote-redux-devtools": "^0.5.12",
-    "remote-redux-devtools-on-debugger": "^0.7.1",
     "lint-staged": "^3.2.6",
     "pre-commit": "^1.2.2",
     "prettier": "^1.3.1",
--- a/package.json
+++ b/package.json
@@ -56,8 +56,6 @@
     "redux-logger": "^2.7.4",
     "redux-persist": "^4.3.1",
     "redux-persist-transform-encrypt": "^1.0.2",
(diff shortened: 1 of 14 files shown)


package.json
```diff
@@ -56,6 +56,8 @@
     "redux-logger": "^2.7.4",
     "redux-persist": "^4.3.1",
     "redux-persist-transform-encrypt": "^1.0.2",
+    "remote-redux-devtools": "^0.5.12",
```
housseindjirdeh: Although the linter is yelling to have these dependencies - this really shouldn't be in `dep` and should be in `devDep`. We can do `// eslint-disable-line import/no-extraneous-dependencies` as a workaround where they're being imported

src/components/search-bar.component.ios.js
```diff
@@ -12,6 +12,13 @@ export class SearchBar extends Component {
     onSearchButtonPress: PropTypes.func.isRequired,
   };
 
+  static defaultProps = {
```
housseindjirdeh: Just realized you're using the `proptypes` package to define props in these components. The rest of the application uses flow so I think it makes sense to do the same here. You can do this by adding a simple `Props` object like so:

```
type Props = {
  textColor: string,
  textFieldBackgroundColor: string,
  showsCancelButton: bool,
  placeholder: string,
  onFocus: Function,
  onCancelButtonPress: Function,
  onSearchButtonPress: Function,
};
```

and then adding it as a type:

```
export const Searchbar = ({
  textColor,
  textFieldBackgroundColor,
  showsCancelButton,
  placeholder,
  onFocus,
  onCancelButtonPress,
  onSearchButtonPress,
}: Props) => {
```

This should save us the need to define defaultProps as well.

Also each of the components in the `components` folder are stateless functional. Although everyone has their own preferences, I think it makes sense to follow the same format here as well. Will prefer if you can make `search-bar.ios` and `search-bar.android` the same if that's okay :)

package.json
patw0929: 👍

src/components/search-bar.component.ios.js
patw0929: Cool, I am not familiar with flow. Looks better.
Fixed in search-bar ios & android.

housseindjirdeh:
Thanks a million once again <3 - it looks good now! Please me know if you're done with updates and I'll merge this in

patw0929:
Done! 🚀🚀🚀
Thanks!

PR #83, opened 2017-07-16
Separate dev commands. Closes #82

### LINKED ISSUE #82: Logging middleware is slow
So `redux-logger` has already been included in this app and it does slow the simulator down. The introduction of `remote-redux-devtools` is now making things significantly slower.

I tried setting up environment variables using [babel-plugin-transform-inline-environment-variables](https://github.com/babel/babili/tree/master/packages/babel-plugin-transform-inline-environment-variables) and separate script commands (`yarn start ios:logger` and `yarn start ios:devtools`) which will include the middleware separately and only when those commands were run. However `process.env` only seemed defined for `NODE_ENV` and nothing else.

Will really appreciate if someone can help find a way to conditionally add separate logging middleware instead of having them included everytime we run `yarn run ios` or `yarn run android`.

lex111:


lex111:
@housseindjirdeh I did a rebase, the logger is running on a separate command.

housseindjirdeh:
Beautiful ⭐️ thanks again for this :)

housseindjirdeh:
CI is failing because of the postinstall script - otherwise this is good

lex111:
Already should not fall :)

Issue #87, opened 2017-07-16
Apply Android fonts


housseindjirdeh:
So the font we use throughout the app is iOS specific (`AvenirNext`). We'll need to make sure a suitable font is applied for Android.

Instead of checking for `Platform` everywhere in the app - I think this will be a good time to add some global fonts to default `Text` component `fontFamily`. A possible solution is [react-native-global-props](https://github.com/Ajackster/react-native-global-props) but I think having a top level root file that exposes different font objects based on the platform will probably be better. Giving it a shot now 🙌

PR #90, opened 2017-07-16
add fonts config file, for #87

housseindjirdeh:
For #87 

Added a separate `fonts` config file to expose separate font objects for iOS/Android. This PR uses the built-in Android font `sans-serif`:

image

Looks decent I think but also thinking of importing a custom font if we can find one that's better (Android [default font list](https://github.com/react-native-training/react-native-fonts) isn't as large as iOS)

Open to all kinds of suggestions and tips 🙌

### DIFF
--- a/src/auth/screens/auth-profile.screen.js
+++ b/src/auth/screens/auth-profile.screen.js
@@ -12,7 +12,7 @@ import {
   ParallaxScroll,
   UserListItem,
 } from 'components';
-import { colors } from 'config';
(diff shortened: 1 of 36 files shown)


lex111:
In my opinion, everything is fine here!

housseindjirdeh:
Thank you mate: appreciate the review :)

housseindjirdeh:
Definitely open to having further discussion/exploration if we can find a better font

Issue #40, opened 2017-07-13
add logout functionality


housseindjirdeh:
So we can't logout of the app right now :). Currently users have to delete the app and redownload in order to sign in with another account.

We can just navigate the user to the initial welcome screen (might be worth clearing our persisted store too).

alejandronanez:
Any steps for the new people to the repo to follow? Also, any thoughts about the design? I think a logout link in the profile page would be enough.

housseindjirdeh:
Thank you @alejandronanez, I had it all in mind and just realized I didn't actually write it down 😞 .

So the `UserProfile` screen has a menu icon on the top right that lets you follow/unfollow users. We don't have this for the `AuthUserProfile` screen and I think we can just add this to open an Action Sheet with a Logout button.

2 caveats:

1. `ActionSheetIOS` is being used and will be changed soon: https://github.com/gitpoint/git-point/issues/2#issuecomment-315032689 to allow for Android functionality. Will appreciate if we use `react-native-actionsheet` here as well
2. There's another ticket open (#8) to allow editing user profile details. If that gets merged in prior --> it probably will make more sense to have `Log Out` included in that screen. But in the meantime the approach mentioned above will work fine

RolfKoenders:
I started to play with logging out a bit and at the moment i have the following.

<a href="http://ibb.co/mZLdSv"><img src="http://thumb.ibb.co/mZLdSv/Screen_Shot_2017_07_17_at_20_31_15.png" alt="Screen_Shot_2017_07_17_at_20_31_15" border="0"></a> <a href="http://ibb.co/c9fpZa"><img src="http://thumb.ibb.co/c9fpZa/Screen_Shot_2017_07_17_at_20_31_25.png" alt="Screen_Shot_2017_07_17_at_20_31_25" border="0"></a>

By clicking yes you would navigate to the 'Login' screen. At this moment i'm trying to figure out how to clear the userData (log the user out).
Till there will be an editing button (Maybe a cogwheel icon next to the profile image will do?) this will do the job i think?

Let me know if i should push my branch and create a pr or its better to combine this with a possible editing feature.

housseindjirdeh:
@RolfKoenders that looks amazing 😍 

[...]

PR #99, opened 2017-07-17
WIP Feature: Logout

RolfKoenders:
### Description
Adding a 'Sign Out' button on the bottom of the profile page.

### Reference
#40


housseindjirdeh:
❤️ ❤️ ❤️ ❤️ 

Will pull this down and find out why it doesn't seem to clear your store data (in a few hours hopefully :))

src/auth/screens/auth-profile.screen.js
```diff
@@ -14,6 +13,8 @@ import {
 } from 'components';
 import { colors, fonts } from 'config';
 import { getUser, getOrgs } from 'auth';
+import { persistStore } from 'redux-persist';
+import { configureStore } from '../../../root.store';
```
alejandronanez: We should use absolute imports in here instead of relative imports.

alejandronanez:
Hey @RolfKoenders, thanks for starting working on this! Let us know if you need something!

src/auth/screens/auth-profile.screen.js
```diff
@@ -36,6 +37,9 @@ const styles = StyleSheet.create({
     color: colors.greyDark,
     ...fonts.fontPrimary,
   },
+  logoutTitle: {
+    color: colors.red,
```
housseindjirdeh: The primary font, `...fonts.fontPrimary` should also be applied here as well! :)

src/auth/screens/auth-profile.screen.js
```diff
@@ -143,6 +147,38 @@ class AuthProfile extends Component {
                 />
               )}
             </SectionList>
+
+            <SectionList title="Sign Out">
+              <ListItem
+                title="Sign Out"
+                titleStyle={styles.logoutTitle}
+                leftIcon={{
```
housseindjirdeh: So sorry for the back and forth here, but having the icon not aligned with the organization avatars above seem a little off. Let's try removing it so it looks like this:

image

Similarly let's add a `hideChevron` prop to hide the right icon

src/auth/screens/auth-profile.screen.js
```diff
@@ -143,6 +147,38 @@ class AuthProfile extends Component {
...
+                  name: 'sign-out',
+                  color: colors.grey,
+                  type: 'octicon',
+                }}
+                onPress={() => {
+                  Alert.alert(
```
housseindjirdeh: Would like to have this in a separate function like so: `onPress={() => this.signOut}` and the logic can be abstracted within the `signOut` method (like how I've abstracted `getUserBlog` in a separate method for example)

src/auth/screens/auth-profile.screen.js
```diff
@@ -143,6 +147,38 @@ class AuthProfile extends Component {
...
+                    'Are you sure you want to sign out?',
+                    [
+                      {
+                        text: 'No',
+                      },
+                      {
+                        text: 'Yes',
+                        style: 'destructive',
+                        onPress: () => {
+                          // persistStore(configureStore).purge();
```
housseindjirdeh: Similarly will like to have this abstracted in a separate function (`clearStoreAndSignOut` for example)

So this is a little trickier then I thought it would be. So if we try purging the store - that still doesn't mean our local Redux state will be cleared. The same with `AsyncStorage.clear()`

First and foremost, AsyncStorage.clear should return a promise so I thought something like this would work:

```
clearStoreAndSignOut = () => {
  AsyncStorage.clear(() => {
    this.props.navigation.navigate('Login');
  });
}
```

However because that only clears AsyncStorage and not the actual current data, you'll notice that it navigates to `Login` but gets navigated right back to the `Main` tab screens because of how we [check our authentication state](https://github.com/gitpoint/git-point/blob/master/src/auth/screens/login.screen.js#L88) when that component mounts.

So what I think we can do is create a new action creator in `auth.action.js` for `LOGOUT`. In here we can do two things:

1. Clear async storage
2.  Reset the state of the current store [just like this](https://stackoverflow.com/questions/35622588/how-to-reset-the-state-of-a-redux-store/35641992#35641992)

Let me know what you think (and don't hesitate to ask for help in the gitter channel - it's what I'm here for :))

andrewda:
For all user-facing parts of the app, the "Sign In/Sign Out" terminology is consistent, but internally it seems like it's being mixed with "Login/Logout". Probably not something that's in the scope of this PR, but I think it'd eventually be nice to straighten that out a little.

housseindjirdeh:
@andrewda agreed, I can fix where it's mentioned in the app to be consistent.

As discussed, thanks so much for this @RolfKoenders, will be taking it on from here and will hopefully get to this today or tomorrow

src/auth/auth.action.js
```diff
@@ -26,6 +33,19 @@ export const auth = (code, state) => {
   };
 };
 
+export const signOut = callback => {
```
alejandronanez: Hey! Is there a way we can omit this callback in this action? It looks like a code smell to me.

alejandronanez: Maybe we can return a promise in here and call the next action later inside the component or something like that.
What do you think @housseindjirdeh

andrewda: Yea that'd probably work and make it a bit nicer :+1:

housseindjirdeh: Agreed 👍

RolfKoenders: Ye true, agree, srry :)

src/auth/auth.reducer.js
```diff
@@ -1,4 +1,10 @@
-import { LOGIN, GET_AUTH_USER, GET_AUTH_ORGS, GET_EVENTS } from './auth.type';
+import {
+  LOGIN,
+  LOGOUT,
+  GET_AUTH_USER,
+  GET_AUTH_ORGS,
+  GET_EVENTS,
+} from './auth.type';
```
alejandronanez: We're using absolute paths here :)

src/auth/auth.action.js
```diff
@@ -1,10 +1,17 @@
...
   fetchUserEvents,
 } from 'api';
-import { LOGIN, GET_AUTH_USER, GET_AUTH_ORGS, GET_EVENTS } from './auth.type';
+import {
+  LOGIN,
+  LOGOUT,
+  GET_AUTH_USER,
+  GET_AUTH_ORGS,
+  GET_EVENTS,
+} from './auth.type';
```
alejandronanez: Use absolute imports here too.

housseindjirdeh: Don't think we really need to here since it's calling it within the same directory right? Or do you still prefer having global imports here as well?

housseindjirdeh:
Wish I had known I can update this PR directly, sorry @RolfKoenders I created a separate one for this at #131

Issue #102, opened 2017-07-17
Searching user repositories don't show forked repos


xuezhma:
1. Repo list is including all repos I have push access to
2. Search results are not including forks.
~~3. Going back to person A's repo list screen from person B's repo list screen will show person B's repo list screen~~ duplicate #3

xuezhma:
Search function is making a request every time. It seems unnecessary as we already fetched all repos belong to that user, we could simply filter that data.

housseindjirdeh:
Cheers thanks for logging this 👍 

1. Interesting, so the list includes all your repos + other user repos you have push access to? So in the app currently, I'm just getting all the repositories for a public user (which is incorrect and has been [fixed since](https://github.com/gitpoint/git-point/commit/cf898e77ac50ae98fd20dcd0d62f4dc3503cb784#diff-62d84b949fccbdc8190dd54246bc8819L129) so not sure why you're seeing that. _However_ with this change I don't think you'll see anything else besides (please feel free to pull the app down and run it on a simulator or device to see the latest changes)
2. Is this in the global search screen or specifically in searching in a user's repository list?
3. Unfortunately it is necessary at the moment because we only show 50 of the user's repositories - if the user has more then filtering wouldn't be sufficient. Open issue to address scrolling and obtaining more data --> #95

xuezhma:
Thanks for the fast response.
I am running the latest code on an iOS simulator.

1. It works correctly for public users. But for the account I logged in, I am able to see other user repos I have push access to. I'm opening a pr in one sec.

2. Searching in a user's repository list. I guess the filtering part is a duplicate to #95. But the results of current search is still wrong.

housseindjirdeh:
Thanks so much for the PR - Just merged it ❤️ 

And you're right, searching doesn't show forks for user repos! What we'll need to do is just add a [`fork` flag](https://developer.github.com/v3/search/#search-repositories) and set to true for the [search call](https://github.com/gitpoint/git-point/blob/master/src/user/user.action.js#L200)

Just updated this issue title to reflect that. Thanks again my friend 🙌

RolfKoenders:
Hmm, im facing a different issue. My profile page shows 29 repositories. It shows 35 repositories in the list view. (Without searching) And all of them are repo's where i have push access to, not my own 'created' repositories, which i would like to see in that screen.

xuezhma:
That sounds like what #105 is trying to fix. With it merged, you should only see your own 'created' repositories now.

[...]

PR #105, opened 2017-07-17
Don't include other user repos that I have push access to

xuezhma:
Fixes first bug in #102


src/user/user.action.js
```diff
@@ -135,7 +135,7 @@ export const getRepositories = user => {
       .then(data => {
         dispatch({
           type: GET_REPOSITORIES.SUCCESS,
-          payload: data,
+          payload: data.filter(repo => repo.permissions.admin),
```
housseindjirdeh: 🎉 🎉

PR #108, opened 2017-07-18
Code push integration

junedomingo:
- Able to update the app instantly
- Manually check for an update by tapping "Check for update" button in profile screen


.env
```diff
@@ -0,0 +1,2 @@
+CODEPUSH_ANDROID_DEPLOYMENT_KEY=
+CODEPUSH_IOS_DEPLOYMENT_KEY=
```
housseindjirdeh: Do we need to put the staging keys here? Or is that something we don't want to push up to the repo?

PR #115, opened 2017-07-18
Show all organizations on auth profile page. Closes #96

### LINKED ISSUE #96: I have some organizations, but not appeared
There are no any organizations on `ORGANIZATIONS` section on my profile page. But I have some orgs. 

It seems a bug? right?

lex111:


src/auth/auth.action.js
```diff
@@ -51,14 +53,24 @@ export const getUser = () => {
...
+        const orgs = [];
+
+        data.forEach(part => part.map(org => orgs.push(org)));
+
         dispatch({
           type: GET_AUTH_ORGS.SUCCESS,
-          payload: data,
+          payload: uniqby(orgs, 'id').sort(
+            (org1, org2) => org1.login > org2.login
+          ),
```
andrewda: GitHub sorts orgs on profiles by the creation date of the organization. I like alphabetically more, but I'm not sure if anyone feels strongly about trying to mimic GitHub?

src/auth/auth.action.js
```diff
@@ -51,14 +53,24 @@ export const getUser = () => {
...
 
-    fetchAuthUserOrgs(accessToken)
+    Promise.all([
+      fetchAuthUserOrgs(accessToken),
+      fetchUserOrgs(login, accessToken),
+    ])
       .then(data => {
+        const orgs = [];
+
+        data.forEach(part => part.map(org => orgs.push(org)));
```
housseindjirdeh: Probably simpler to just do `const orgs = data[0].concat(data[1])`

src/auth/auth.action.js
housseindjirdeh: I don't mind alphabetically either actually 👍

src/auth/auth.action.js
lex111: Exactly, I completely forgot :confused:

lex111:
All fixed, also added a note:
image

P.S. And can it be better to take `yarn.lock` in .gitignore file?

andrewda:
@lex111 I think the `yarn.lock` isn't supposed to be ignored, but could be wrong about that.

housseindjirdeh:
Yep you definitely want to [check in yarn.lock files](https://yarnpkg.com/blog/2016/11/24/lockfiles-for-all/)

Thanks @lex111 this looks amazing 🎉

lex111:
@andrewda @housseindjirdeh ok, thanks for the links with the explanations.

Issue #92, opened 2017-07-17
Numbered list count bug


Antoine38660:
Just found this weird one : 

`2.` on numbered list seems to be skipped...

Screenshots:
<img src="https://user-images.githubusercontent.com/9287184/28253035-b7282c0e-6a74-11e7-8a7c-3774af66e447.jpg" alt="Gitpoint" width="300"/>
Original from [Github.com](https://github.com) :
<img width="782" alt="Github" src="https://user-images.githubusercontent.com/9287184/28253036-b72886ae-6a74-11e7-833a-40759401046a.png">

housseindjirdeh:
Interesting - not sure why this is happening

Good catch however! I do a bunch of conditional checks on the type of DOM element in `CommentListItem` for the comment HTML retrieved, so I may be doing something unnecessary with ordered lists.

Antoine38660:
**Some more information after another test**
Initial markdown:
> 1. One
> 2. Two
> 3. Three
> 4. Four
> 5. Five
> 6. ...
<img src="https://user-images.githubusercontent.com/9287184/28279384-b01dcee6-6af6-11e7-8871-f6aded88352f.png" width="300" alt="screenshot" />

The bug exist for each _even number_.

Antoine38660:
Found this one: https://github.com/jsdf/react-native-htmlview/issues/92 

I tried to update `react-native-htmlview` to 0.11.1 but it doesn't seem to be fixed... I'll reopen an issue there.
I also checked the `commentBodyAdjusted` (https://github.com/gitpoint/git-point/blob/master/src/components/comment-list-item.component.js#L283) but there is no problem with this generated DOM.

Seems definitely to be an `HTMLView` issue 😉

housseindjirdeh:
Ah thanks a million for diving into this, looks like it's definitely something with the package.

Cheers let us know when you create an issue there and we'll link it here - in the meantime we can leave this ticket open until that gets resolved.

Antoine38660:
Update: Just submitted a PR https://github.com/jsdf/react-native-htmlview/pull/131 to fix this one :smiley:

PR #118, opened 2017-07-18
Fix <li> prefix number on ordered list (<ol>) #92

Antoine38660:
I also submitted a PR to `react-native-htmlview` but it didn't pass the CI test and there are some old ones not merged...
So this quick fix will works for us 👍

### DIFF
--- a/src/components/comment-list-item.component.js
+++ b/src/components/comment-list-item.component.js
@@ -229,7 +229,7 @@ export class CommentListItem extends Component {
       } else if (node.name === 'li' && parent.name === 'ol') {
         return (
           <Text key={index} style={textStyle}>
-            {`${Math.ceil((index + 1) / 2)}.`}{' '}
+            {`${Math.ceil((index + 1) / 2)}. `}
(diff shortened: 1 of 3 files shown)


src/components/comment-list-item.component.js
```diff
@@ -226,6 +226,13 @@ export class CommentListItem extends Component {
             {defaultRenderer(node.children, parent)}
           </Text>
         );
+      } else if (node.name === 'li' && parent.name === 'ol') {
+        return (
+          <Text key={index} style={textStyle}>
+            {`${Math.ceil((index + 1) / 2)}.`}{' '}
```
andrewda: Would ``{`${Math.ceil((index + 1) / 2)}. `}`` work here to keep things simple?

Antoine38660: Oh yes of course ! I didn't put this `{' '}`, it was probably added by the pre-commit hook.

I push an update now

andrewda: Huh, wonder if that's `prettier` making things ugly :joy:

PR #123, opened 2017-07-19
Add npm scripts for quick develop

lex111:
As an alternative to the two commands: `yarn start` + `yarn run android|ios`, instead just run `yarn start: android` or `yarn start:ios` - you do not need to run two terminals. Quickly and convenient.

Can I add the appropriate commands with `logger`?

### DIFF
--- a/package.json
+++ b/package.json
@@ -3,11 +3,15 @@
   "version": "0.0.1",
   "private": true,
   "scripts": {
-    "start": "node node_modules/react-native/local-cli/cli.js start",
-    "start:logger": "LOGGER_ENABLED=true node node_modules/react-native/local-cli/cli.js start --reset-cache",
-    "link": "./node_modules/.bin/react-native link",
-    "ios": "./node_modules/.bin/react-native run-ios",
-    "android": "./node_modules/.bin/react-native run-android",
+    "start": "react-native start",
+    "start:logger": "LOGGER_ENABLED=true react-native start --reset-cache",
+    "start:android": "concurrently -r 'react-native start' 'react-native run-android'",
+    "start:android:logger": "LOGGER_ENABLED=true concurrently -r 'react-native start --reset-cache' 'react-native run-android'",
+    "start:ios": "concurrently -r 'react-native start' 'react-native run-ios'",
+    "start:ios:logger": "LOGGER_ENABLED=true concurrently -r 'react-native start --reset-cache' 'react-native run-ios'",
+    "link": "react-native link",
+    "ios": "react-native run-ios",
+    "android": "react-native run-android",
     "lint-staged": "lint-staged",
     "lint-pass": "echo '\\033[42mCOMMIT SUCCEEDED\\033[0m 🎉 🎉 🎉\n'",
     "eslint": "eslint .",
@@ -90,7 +94,8 @@
     "lint-staged": "^3.2.6",
     "pre-commit": "^1.2.2",
     "prettier": "^1.3.1",
-    "react-test-renderer": "16.0.0-alpha.6"
+    "react-test-renderer": "16.0.0-alpha.6",
+    "concurrently": "^3.5.0"
   },
   "jest": {
     "preset": "react-native"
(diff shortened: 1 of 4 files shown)


package.json
```diff
@@ -5,6 +5,8 @@
   "scripts": {
     "start": "node node_modules/react-native/local-cli/cli.js start",
     "start:logger": "LOGGER_ENABLED=true node node_modules/react-native/local-cli/cli.js start --reset-cache",
+    "start:android": "concurrently -r 'node node_modules/react-native/local-cli/cli.js start' './node_modules/.bin/react-native run-android'",
```
alejandronanez: You can do `react-native run-android` instead of `./node_modules/.bin/react-native run-android`

andrewda: We can do that for a lot of things here if I'm not mistaken (e.g. `start` could just be `react-native start`)

alejandronanez: That's true. If you don't mind you can do it too here :)

lex111: Yes, I also thought about it. But I often saw that usually write  the commands with full paths (it seems, for greater compatibility)

alejandronanez: hmmm I haven't had any issue with that. Basically what node does is searching inside the `node_modules/.bin` folder for those executables (that should be installed after running `yarn`)

alejandronanez:
About

> Can I add the appropriate commands with logger?

Can you clarify a little please?

package.json
lex111: Perhaps the reason is that you will not have globally `react-native`?

alejandronanez: Your local modules will have precedence over the globally installed ones, so it's safe to remove the whole path.

alejandronanez: As long as you're using something that's inside `node_modules/.bin` you'll be ok.

lex111:
> Can you clarify a little please?

@alejandronanez We still have a [command to run with a logger](https://github.com/gitpoint/git-point/pull/123/files#diff-b9cfc7f2cdf78a7f4b91a753d10865a2R7), can add the same commands (with `concurrently`), but with the run of the logger?

alejandronanez:
I think running the logger should be optional, if you want to add concurrently to it we should have the option for running it with `yarn start:ios:logger` for example.
It's a good use case, go for it. Can you please update the README.md file too with the new commands?

Thanks!

package.json
```diff
@@ -3,11 +3,15 @@
...
-    "start:logger": "LOGGER_ENABLED=true node node_modules/react-native/local-cli/cli.js start --reset-cache",
-    "link": "./node_modules/.bin/react-native link",
-    "ios": "./node_modules/.bin/react-native run-ios",
-    "android": "./node_modules/.bin/react-native run-android",
+    "start": "react-native start",
+    "start:logger": "LOGGER_ENABLED=true react-native start --reset-cache",
+    "start:android": "concurrently -r 'react-native start' 'react-native run-android'",
+    "start:android:logger": "LOGGER_ENABLED=true concurrently -r 'react-native start' 'react-native run-android'",
+    "start:ios": "concurrently -r 'react-native start' 'react-native run-ios'",
+    "start:ios:logger": "LOGGER_ENABLED=true concurrently -r 'react-native start' 'react-native run-ios'",
```
andrewda: Do these need `react-native start --reset-cache`? That's what the `start:logger` script has, not sure if it's important.

lex111: Of course, fixed :+1:

lex111:
@alejandronanez the scripts updated, with the update README.md could you help me? I'm afraid there will be a lot of duplication of the same words.

lex111:
@housseindjirdeh great, I generally thought of leaving the old scripts, but it's probably better. And how to write a description for the logger - `yarn start:ios:logger`, do not duplicate the same?

housseindjirdeh:
@lex111 we can probably just add it to the same line. For example:

- `yarn start:ios` to start the packager and run the app in the iOS simulator (`yarn start:ios:logger` will boot the application with [redux-logger](https://github.com/evgenyrodionov/redux-logger))
- `yarn start:android` to start the packager and run the app in the the Android device/emulator (`yarn start:android:logger` will boot the application with [redux-logger](https://github.com/evgenyrodionov/redux-logger))

Let me know if that works mate :raised_hands:

lex111:
@housseindjirdeh these commands work, updated the description of the build process in all required files :heavy_check_mark:

PR #126, opened 2017-07-19
Remove listContainer border while it's loading

junedomingo:
a

### DIFF
--- a/src/search/screens/search.screen.js
+++ b/src/search/screens/search.screen.js
@@ -181,6 +181,9 @@ class Search extends Component {
       users.length === 0 &&
       searchType === 1;
 
+    const isPending = isPendingSearchUsers || isPendingSearchRepos;
+    const noResults = !noUsersFound && !noReposFound;
+
     return (
       <ViewContainer>
         <View>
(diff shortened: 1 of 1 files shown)


src/search/screens/search.screen.js
```diff
@@ -227,7 +230,13 @@ class Search extends Component {
         {searchStart &&
           !noUsersFound &&
           !noReposFound &&
-          <View style={styles.listContainer}>
+          <View
+            style={[
+              styles.listContainer,
+              (isPendingSearchUsers || isPendingSearchRepos) &&
```
alejandronanez: I like this. Maybe you can do `const isLoading = isPendingSearchUsers || isPendingSearchRepos` and later add that constant in here (it's a little more explicit)

alejandronanez: What do you think @junedomingo ?

alejandronanez:
Let us know @junedomingo if you think that change is necessary and if not, let us know so we can merge your changes!

src/search/screens/search.screen.js
junedomingo: That works for me :-)

junedomingo:
I also assign noResult to new constant for consistency, what do you think?

andrewda:
Very clean! Looks good to me!! 👍

PR #130, opened 2017-07-19
Main language in Repository view. Close #119

### LINKED ISSUE #119: Display repository language
The repository's primary programming language string can be found at `repository.language`.

Antoine38660:
Finally positioned top-centered 😄  
Android and iOS preview:

 screenshot_1500504737 
----
simulator screen shot 19 juil 2017 a 17 50 42

### DIFF
--- a/src/components/repository-profile.component.js
+++ b/src/components/repository-profile.component.js
@@ -81,9 +81,9 @@ const styles = StyleSheet.create({
   },
   languageInfo: {
     flexDirection: 'row',
+    marginTop: 35,
   },
   languageInfoTitle: {
-    // flex: 1,
     color: colors.white,
     paddingLeft: 3,
     fontSize: normalize(10),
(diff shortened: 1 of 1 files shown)


src/components/repository-profile.component.js
```diff
@@ -79,10 +79,33 @@ const styles = StyleSheet.create({
...
+    color: colors.white,
+    paddingLeft: 3,
+    fontSize: normalize(10),
+    ...fonts.fontPrimary,
+  },
 });
 
 export const RepositoryProfile = ({ repository, starred, navigation }: Props) =>
   <View style={styles.container}>
+    {repository.language !== null &&
```
andrewda: Have you tried this on a repository without a language? I'm pretty sure GitHub responds with an empty string and not `null`, but we may be doing all that conversion somewhere else.

andrewda: Have you tried this on a repository without a language? I'm pretty sure GitHub responds with an empty string and not null, but we may be doing all that conversion somewhere else.

Antoine38660: I haven't tried with an empty repo (_without a language_). But I was inspired by https://github.com/gitpoint/git-point/blob/master/src/components/repository-list-item.component.js#L86 (Moreover, the condition in this file is only for the icon node).

Antoine38660: From the doc (https://developer.github.com/v3/repos/#get) language can be `null` :)

housseindjirdeh: Yep they return null (I had also assumed they would be empty string but that doesn't seem to be the case). Here's an example of an empty repo (no language) 
https://api.github.com/repos/housseindjirdeh/test-repo

PR #131, opened 2017-07-20
add logout

housseindjirdeh:
Logging out is a little tricky, I can't thank you enough @RolfKoenders for getting it started and apologize you didn't have the easiest thing to start working on.

We use the `Oauth` web flow in the application and not `Basic Authentication`. In the API, revoking access tokens and deleting authorizations [only work](https://developer.github.com/v3/oauth_authorizations/) for `Basic Authentication`.

This PR does the following:

1. Some minor styling in the `AuthProfileScreen` (Check for updates tap target is a bit bigger now, centered and darkened note on requesting access on repos, made sure `fontSizes` were normalized.

<p align="center">

</p>

2. When the user logs out, the `auth` reducer portion of the Redux store is cleared. That's where we store the access token and `isAuthenticated` logic
3.  `AsyncStorage` is also cleared because the Redux store is persisted there
4. User is navigated to the Login Screen

All this works just fine. The tricky part is when the user clicks login once again. This is how it would work:

justnavigate

Notice how I didn't have to type in my username and password again. This is because I'm still logged in to Safari hence why I'm just redirected right back with a new access token.

So here's what I did: open `github.com/logout` using the Safari View to allow the user to log out through the browser. Unfortunately they have to click `Done` to be able to navigated back to the app (I don't know if it's possible to redirect automatically after a Safari event)

completelogout

Let me know what you folks think, and please let me know if there's anything I should change in this PR :raised_hands:

### DIFF
--- a/src/auth/screens/auth-profile.screen.js
+++ b/src/auth/screens/auth-profile.screen.js
@@ -8,6 +8,7 @@ import {
   Linking,
 } from 'react-native';
 import { ListItem } from 'react-native-elements';
+import { StyleSheet, Alert, Text, TouchableOpacity, View } from 'react-native';
 import codePush from 'react-native-code-push';
 
 import {
@@ -20,7 +21,7 @@ import {
   EntityInfo,
 } from 'components';
 import { colors, fonts } from 'config';
-import { getUser, getOrgs } from 'auth';
+import { getUser, getOrgs, signOut } from 'auth';
 
 const mapStateToProps = state => ({
   user: state.auth.user,
@@ -32,6 +33,7 @@ const mapStateToProps = state => ({
 const mapDispatchToProps = dispatch => ({
   getUser: () => dispatch(getUser()),
   getOrgs: () => dispatch(getOrgs()),
+  signOut: callback => dispatch(signOut(callback)),
 });
 
 const styles = StyleSheet.create({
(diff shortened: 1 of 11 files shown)


src/auth/screens/auth-profile.screen.js
```diff
@@ -113,6 +121,34 @@ class AuthProfile extends Component {
...
+  openLink = url => {
+    if (Platform.OS === 'ios') {
+      SafariView.show({
+        url,
+        fromBottom: true,
+      });
+    } else {
+      Linking.openURL(url);
+    }
+  };
```
andrewda: I feel like openLink is a fairly complex method that will (and should) likely be used in other places. Not sure where it should be moved to, though, maybe `utils`?

housseindjirdeh: Agreed it's already being used elsewhere, will update it in a bit 🙌

housseindjirdeh: Done - moved it to a helper method 👍

andrewda:
I rdon't eally understand the inner-workings of how the login session is maintained, but would something like https://github.com/reactnativecn/react-native-http-cache/ work?

housseindjirdeh:
Hmm thanks will take a look at that tomorrow for sure. The logout flow is relatively straightforward but it's just that the password/username persists in Safari hence it logs you in automatically when a user logs out and tries to sign in again (as opposed to having you type your details once more)

andrewda:
Ah ok. Yea clearing the cache seems like it'd probably work (right?) but not sure if the package I linked above will do that.

PR #133, opened 2017-07-20
Add auto-completion of usernames

### LINKED ISSUE #110: Auto complete usernames in @ mentions
It's difficult to type out long usernames everytime.

andrewda:
Fixes #110.

Demo

(or slightly higher quality YouTube video: https://www.youtube.com/watch?v=7jnJddP0aRA)

Anyways, the basic features are all laid out and it works fine, but would like to add a couple more things before merging:

- [x] ~~Popup the mention box by typing an "@" anywhere in the sentence~~ (can't be done easily... maybe change the logic around a bit in the future to overcome this as it's not _super_ important)
- [x] ~~Make it work with issue/PR `#`'s~~ (another PR)
- [x] ~~Add normalized name if available (e.g. "@johndoe62 - John Doe")~~ (Not possible ATM, maybe after switching to GraphQL/GitHub API v4)
- [x] Add fuzzy search
- [x] Make it look nicer
- [x] ~~Maybe add profile images?~~ (I kind of like it how it is)
- [x] Take some dope pictures/videos for this PR 😉


src/components/mention-area.component.js
```diff
@@ -0,0 +1,134 @@
...
+  }
+
+  onSuggestionTap(user, close) {
+    const { text, trigger } = this.props;
+
+    this.props.updateText(
+      `${text.slice(0, text.lastIndexOf(trigger))} @${user} `
+    );
+
+    if (close) this.closeSuggestionsPanel();
```
andrewda: Change this to `this.stopTracking()`

housseindjirdeh:
I can't even begin to describe how awesome of a job this is @andrewda, wayy exceeded my expectations and you're not even done!

Does clicking any name in the popup mentions box automatically insert the name into the comment text input? If it doesn't that's fine, something we can work on later because even without it this is solid.

andrewda:
@housseindjirdeh Yep! It only works at the end of the current text at the moment but I'm working on making it work anywhere.

src/components/mention-area.component.js
```diff
@@ -0,0 +1,169 @@
...
+        duration: 100,
+        toValue: this.props.height,
+        friction: 4,
+      }).start();
+    }
+  }
+
+  renderSuggestionsRow(users) {
+    return users.map(user =>
+      <TouchableHighlight
```
andrewda: Let's use TouchableOpacity instead.

andrewda:
I think I'll hold off on the issue and PR suggestions for now so I can implement it a little differently in another PR. Just gotta resolve a couple issues and merge conflicts and should be good to go! :)

src/components/mention-area.component.js
```diff
@@ -0,0 +1,177 @@
...
+    if (newValue < this.props.height) {
+      Animated.timing(this.state.height, {
+        duration: 100,
+        toValue: newValue,
+      }).start();
+    } else {
+      Animated.spring(this.state.height, {
+        duration: 100,
+        toValue: this.props.height,
+        friction: 4,
```
housseindjirdeh: I feel like the bounce is a tiny but too much when the mentions box first shows. Tried a friction of `8` or `10` and it seems a little smoother.

andrewda: Alright sounds good 👍

andrewda:
@housseindjirdeh Thanks so much!

1. Will do that in a sec 😄 
2. This happens because I'm checking to see if the last character in the text field is an `@` symbol. Not sure how to check what the last character that was actually typed is...
3. Weird not sure why that's happening. I'll look into this but doesn't seem like a huge deal, probably just open an issue ticket and we can check in on it when anyone gets a chance.

Haven't tested on Android, would much appreciate it if u could check it out <3

src/components/comment-input.component.js
```diff
@@ -90,6 +92,13 @@ export class CommentInput extends Component {
 
     return (
       <View style={styles.container}>
+        <MentionArea
+          trigger="@"
+          text={this.state.text}
+          updateText={text => this.setState({ text })}
```
alejandronanez: Hey @andrewda! Can you move this arrow function to a class method instead, if we do it that way we will use the function reference instead of creating a function every time we update the text.

src/components/mention-area.component.js
```diff
@@ -0,0 +1,177 @@
...
+      toValue: 0,
+    }).start();
+  }
+
+  updateHeight(num) {
+    const newValue = num * 50;
+
+    if (newValue < this.props.height) {
+      Animated.timing(this.state.height, {
+        duration: 100,
```
alejandronanez: I think we can move this `duration` to a constant so we can use it later in another components that need animation === 100. What do you think?

src/components/mention-area.component.js
```diff
@@ -0,0 +1,177 @@
...
+        friction: 10,
+      }).start();
+    }
+  }
+
+  renderSuggestionsRow(users) {
+    return users.map(user =>
+      <TouchableOpacity
+        key={user}
+        onPress={() => this.onSuggestionTap(user, true)}
```
alejandronanez: Same comment about not using arrow functions.

src/components/comment-input.component.js
andrewda: Thanks for the feedback! What's the benefit of doing it this way?

src/components/mention-area.component.js
andrewda: Good idea. Would that constant fit best in a config file??

alejandronanez: Yes sir!

src/components/comment-input.component.js
alejandronanez: Long story short, it hurts performance. You will find more info here =] https://stackoverflow.com/a/36677798

andrewda: Never knew that, thanks for the info!

andrewda: Actually now that I look at it we use arrow functions like this in a lot of places throughout the code. It's probably best to update this all together in a separate PR and open an issue.

housseindjirdeh: That's a solid point @alejandronanez, agree with @andrewda we could try tackling that everywhere in a separate ticket

alejandronanez: I'm ok with this!

src/config/animations.js
```diff
@@ -0,0 +1,2 @@
+export const speed = 100;
+export const friction = 10;
```
housseindjirdeh: <p align="center">

</p>

I like that `animations` constants are in their own file! But two things I noticed:

1. `duration` isn't in the `animations.js` config file
2. You probably want to export it as an object called `animations` so you can access it with `animations.property` like you did in `mention-area`.

```
export const animations = {
   speed: 100,
   ...
}
```

andrewda: 😅  Totally my bad, did this quickly completely without testing. Should be fixed now, could you test again?

andrewda:
Will begin working on issue implementation tomorrow!

PR #135, opened 2017-07-20
Load issue on issue.screen; Closes #124

### LINKED ISSUE #124: Issues take a long time to load from the notifications screen
The transition isn't super smooth. It could just be my slow internet, but it would be nice for the issue screen with the loading indicator to appear immediately after clicking on the issue.

Antondomashnev:
Hi there ✋ 
This PR reflects the issue #124. The changes I've made:
* `notifications.screen` passes `issueURL` instead of `issue` to `issue.screen`
* `getIssueCommentsByDispatch` from `issue.actions` expects not issue but issueCommentsURL from now on. This change allows to parallel two requests on `issue.screen`: `getIssueFromUrlByDispatch` and `getIssueCommentsByDispatch`
* `LoadingContainer` has the text now as it maybe shown for a longer period as it was shown before.

Overall this is my first javascript PR ever, so thanks for the pre-commit hooks and I'm looking forward to hearing your feedback 😄


src/issue/screens/issue.screen.js
```diff
@@ -40,7 +40,8 @@ const mapStateToProps = state => ({
 });
 
 const mapDispatchToProps = dispatch => ({
-  getIssueCommentsByDispatch: url => dispatch(getIssueComments(url)),
+  getIssueCommentsByDispatch: issueCommentsURL =>
+    dispatch(getIssueComments(issueCommentsURL)),
```
andrewda: Might be nice to keep this as `url => dispatch(getIssueComment(url)),` to keep things on one line, but either way works for me

src/issue/screens/issue.screen.js
```diff
@@ -214,14 +220,25 @@ class Issue extends Component {
...
 
     return (
       <ViewContainer>
-        {isPendingComments &&
-          <LoadingContainer animating={isPendingComments} center />}
+        {(isPendingComments || isPendingIssue) &&
+          <LoadingContainer
+            animating={isPendingComments || isPendingIssue}
+            center
+            text={'Retrieving issue'}
```
andrewda: Not sure if we want the text here (yet, at least), it would be the only place in the app besides searches where we have loading text AFAIK. If we add it here I'd also want to add it to users, code and repositories. It might fit better in another PR.

src/issue/screens/issue.screen.js
housseindjirdeh: If I'm not mistaken `prettier` does this with longer lines but I could be wrong. I'm fine with it either way

src/issue/screens/issue.screen.js
housseindjirdeh: Agreed - I really don't mind having text in all the loading icons but we can just leave it out to match the other screens for now.

I've been meaning to create a LoadingIssueScreen as a loading placeholder (just like the `LoadingRepositoryListScreen` for example) but I didn't have the time before first release and was like meh that can come later :P. Ideally we would have that instead of a spinner so I'm going to open an issue for that as a separate ticket.

src/issue/screens/issue.screen.js
andrewda: Yea, that's why I'm thinking of shortening the line to use `url` instead of `issueCommentsURL` because `prettier` would keep it all on the same line 👍

src/issue/screens/issue.screen.js
Antondomashnev: @housseindjirdeh @andrewda I was using the `notifications.screen` as a reference and it contains the text message, but sure thing, the UX should be consistent. I think it's a good idea to put texts in a separate PR. Removed in 5dd58c6 😉

src/issue/screens/issue.screen.js
Antondomashnev: Good point 👍 updated in 96ad24e

Antondomashnev:
> First JavaScript PR??? @Antondomashnev this is just outstanding 🎉 Could barely find anything that I would change amazing work.

> Awesome, thanks so much @Antondomashnev!

Thanks @housseindjirdeh and @andrewda  😄

Antondomashnev:
@housseindjirdeh done 🚀 
Looking forward to take another one 😄

Issue #136, opened 2017-07-21
Star count display


RogerAbyss:
It`s just a small improve.
I think that star count display counld be improve like github do on website.
star:81220 shows 8.1k may look better, just like thist:
![dc8b5680-a6e1-4893-a16a-c0ec75143e18](https://user-images.githubusercontent.com/8122587/28445870-b9542b8a-6df9-11e7-847b-07badea2d50e.png)

housseindjirdeh:
Agreed, it's a small detail but I like it 😊

PR #138, opened 2017-07-21
Stardisplay

RogerAbyss:
dc8b5680-a6e1-4893-a16a-c0ec75143e18


src/utils/text-helper.js
```diff
@@ -1,5 +1,11 @@
 import emoji from 'node-emoji';
 
+const thousandUnit = 'k';
+
 export const emojifyText = text => {
   return emoji.emojify(text);
 };
+
+export const numbersText = count => {
+  return count > 999 ? (count / 1000).toFixed(1) + thousandUnit : count;
```
andrewda: Tiny nit-pick but I would say `count >= 1000` just to make it more friendly.

RogerAbyss: fine.

RogerAbyss:
Needs close #136

RogerAbyss:
okey,i fix another small bug that user-page when repositories is 0.
8a91087f-bb43-4c37-8867-744e9e20e27a
By the way, i find that user-page has some bug.I`ll open another iusse.
Hope this project will be better and better! It look really pretty good!

src/components/user-profile.component.js
```diff
@@ -109,7 +109,7 @@ export const UserProfile = ({
           })}
       >
         <Text style={styles.unitNumber}>
-          {user.public_repos + (user.total_private_repos || 0) || ''}
+          {user.public_repos + (user.total_private_repos || 0)}
```
housseindjirdeh: Thanks so much for catching this :tada:

src/components/repository-list-item.component.js
```diff
@@ -80,7 +81,9 @@ const renderTitle = repository =>
       />
 
       <Text style={[styles.extraInfoSubject, { paddingTop: 2 }]}>
-        {repository.stargazers_count}
+        {repository.stargazers_count >= 1000
```
housseindjirdeh: Did you decide to not have this in `utils`? I think it might actually be better the way you had it before (separate method in `utils` for `getStartNumber` for example. This way we can import the method and use it here _and_ in `repository-profile.component.js` where the stars show on the repository screen.

Let me know if that's alright with you!

RogerAbyss: I think put it in utils will be better.I think  i miss understanding you meanning before.😃😃

PR #139, opened 2017-07-21
[WIP] Opening the profile avatar in full size by click. Closes #7

### LINKED ISSUE #7: Clicking profile image navigates to full screen
Would be nice if clicking the image on a user profile screen opens a modal screen that shows the image in full screen mode :)

lex111:
When clicking on the avatar, the full size of the picture opens in the modal window (closes #7).

image

Important note: I tested the work only on Android, **please see how it works on iOS**.

Also considered the use of the library:
[react-native-lightbox](https://github.com/oblador/react-native-lightbox) - slows opening on Android
[react-native-zoom-image](https://github.com/Tinysymphony/react-native-zoom-image ) - works fine, but does not apply to miniature styles (`borderRadius` in particular)

Therefore, I had to make my own component, using the built-in component `Modal` and PhotoView.

### DIFF
--- a/android/app/build.gradle
+++ b/android/app/build.gradle
@@ -135,6 +135,7 @@ android {
 }
 
 dependencies {
+    compile project(':react-native-photo-view')
     compile project(':react-native-config')
     compile project(':react-native-code-push')
     compile project(':react-native-device-info')
--- a/android/app/src/main/java/com/gitpoint/MainApplication.java
+++ b/android/app/src/main/java/com/gitpoint/MainApplication.java
@@ -3,6 +3,7 @@ package com.gitpoint;
 import android.app.Application;
 
 import com.facebook.react.ReactApplication;
+import com.reactnative.photoview.PhotoViewPackage;
 import com.lugg.ReactNativeConfig.ReactNativeConfigPackage;
 import com.microsoft.codepush.react.CodePush;
 import com.learnium.RNDeviceInfo.RNDeviceInfo;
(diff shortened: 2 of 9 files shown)


android/app/build.gradle
```diff
@@ -135,6 +135,7 @@ android {
 }
 
 dependencies {
+    compile project(':react-native-photo-view')
```
andrewda: Why does this need a native dependency? I feel like it'd probably be best to avoid extraneous native dependencies if possible.

housseindjirdeh:
This looks soo nice 👍🏾 I'll test it on my iPhone today and let you know it renders.

android/app/build.gradle
lex111: @andrewda In this dependence there is the "pinch to zoom" feature, it's convenient. This may come in handy in the future (to display other images with zoom). Although I agree with you, as long as this is not necessary, why can use the usual component `Image`?

andrewda: Ah okay yea pinch to zoom is nice, and that makes sense! Either way works for me :)

lex111: @andrewda great :smile: 

If not against @housseindjirdeh, then I would leave this dependency, it does not seem heavy and does not interfere with the application. 

BTW, I now thought, how much the application will increase in size after connection of any dependency (or it becomes more slow to work for example)? What are thoughts?

housseindjirdeh: i was thinking the same thing these past few weeks. It's React Native - so it's inevitable that more and more dependencies are going to be added from here on out - that's one of the reasons why I love RN so much. 

But just like you said, we all usually try limiting dependencies in our applications generally to reduce bloat. I really _don't know_ how much application load times/performance is effected in a native application to be honest. However I of course would probably agree with most people and limit dependencies to only those we particulary need, even if a native application isn't as affected as much.

Issue #114, opened 2017-07-18
Some width problems at repository title


jstrobel:
![img_8114](https://user-images.githubusercontent.com/1428396/28333670-edf3eb12-6bf8-11e7-9d59-da816625c9eb.PNG)

As you can see at the screenshot there are some width issues with the repository title.

andrewda:
One option is to move the repo description down near the owner (either above or below). Not sure how that would look.

jstrobel:
isn't there a way to set a limit for the width?

RolfKoenders:
I think the location of the description is good. A solution could be to set a padding of like 10 to 15 pixels on both sides and expand the text view down?

andrewda:
@RolfKoenders :+1: Yep that'd work nicely! Is there a character limit for repository descriptions?

RolfKoenders:
@andrewda After trying it with different characters, there is no actual limit on how much you can insert/save. Github is only showing **347** characters before it cuts it of "Loru..."

andrewda:
Alright, then does it make sense to have the text view expand dynamically?

RolfKoenders:
I think we should decide on a limit. 1/3 of my iPhone 6(s) Plus screen is filled with the repository description with 347 characters.. Thats to much i think?

andrewda:
Yep I agree (though I don't think there are many repos with huge descriptions). Maybe add an ellipsis after the word before the text reaches ~150 characters?

housseindjirdeh:
So we can definitely add some padding on the sides (and we probably should) but super long descriptions can still look funky.

In the list screen long descriptions look fine in my opinion:

![image](https://user-images.githubusercontent.com/12476932/28342829-409c0752-6be8-11e7-81a9-8bc36609a9d7.png)

But in the repo screen this can be a problem:

![image](https://user-images.githubusercontent.com/12476932/28342843-4fd1ee44-6be8-11e7-8b30-0398dfa8cd6d.png)

So I think we should:

1. Add a _little_ padding (or width) so long descriptions aren't so close to the left and right edges of the screen
2. Definitely cut out descriptions after a certain point with ellipsis (~150 characters sounds fine)

jstrobel:
Looks good for me, and cutting after some kind of characters seems fine.

RolfKoenders:
@housseindjirdeh That sounds like a decent solution!

Antondomashnev:
[...]

PR #142, opened 2017-07-21
Better UI for long repository description;

Antondomashnev:
These are small bits of UI improvement related to #114 😄 

This is how the long title looks like with changes:

I've changed the margin left & right to `30` and instead of truncating string programatically decided to use `numberOfLines` props of `Text`.

### DIFF
--- a/src/components/repository-profile.component.js
+++ b/src/components/repository-profile.component.js
@@ -122,7 +122,6 @@ export const RepositoryProfile = ({ repository, starred, navigation }: Props) =>
       </Text>
 
       <Text
-        /* @see https://github.com/gitpoint/git-point/issues/114 */
         numberOfLines={5}
         style={[
           styles.subtitle,
--- a/src/components/repository-profile.component.js
+++ b/src/components/repository-profile.component.js
@@ -36,8 +36,8 @@ const styles = StyleSheet.create({
     color: colors.white,
     ...fonts.fontPrimary,
     fontSize: normalize(12),
-    paddingLeft: 15,
-    paddingRight: 15,
+    paddingLeft: 30,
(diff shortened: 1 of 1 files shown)


src/components/repository-profile.component.js
```diff
@@ -123,6 +123,8 @@ export const RepositoryProfile = ({ repository, starred, navigation }: Props) =>
       </Text>
 
       <Text
+        /* @see https://github.com/gitpoint/git-point/issues/114 */
```
andrewda: Probably don't need this comment

src/components/repository-profile.component.js
```diff
@@ -36,8 +36,8 @@ const styles = StyleSheet.create({
     color: colors.white,
     ...fonts.fontPrimary,
     fontSize: normalize(12),
-    paddingLeft: 15,
-    paddingRight: 15,
+    paddingLeft: 30,
+    paddingRight: 30,
```
andrewda: _Maybe_ a little bit of top padding?

junedomingo: for padding left and right, just use `paddingHorizontal: 30`

src/components/repository-profile.component.js
Antondomashnev: I initially wanted to use if we truncate the string, but than found the `numberOfLines` option. I think yeah, we can remove it. Done in 4b55d88

src/components/repository-profile.component.js
Antondomashnev: @junedomingo thanks, didn't know about that before. Used in d2ce96f

> Maybe a little bit of top padding?
For me currently look good, don't you like how the UI looks now?

PR #145, opened 2017-07-22
Small UI fixes

junedomingo:
- Fixed issues and pull requests header
group

- Hid language dot icon while it's loading
language-header

### DIFF
--- a/src/components/repository-section-title.component.js
+++ b/src/components/repository-section-title.component.js
@@ -19,6 +19,9 @@ const styles = StyleSheet.create({
     color: colors.black,
     ...fonts.fontPrimaryBold,
   },
+  badgeContainer: {
+    flexDirection: 'row',
+  },
   badge: {
     marginLeft: 10,
   },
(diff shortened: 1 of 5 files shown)


src/components/repository-section-title.component.js
```diff
@@ -33,8 +35,11 @@ export const RepositorySectionTitle = ({
       <Text style={styles.titleText}>
         {text}
       </Text>
-      <StateBadge type="open" text={openCount} style={styles.badge} />
-      <StateBadge type="closed" text={closedCount} style={styles.badge} />
+      {!loading &&
+        <View style={styles.title}>
```
housseindjirdeh: `<View style={styles.title}>` looks duplicated here and in line 34. I'm assuming we need `flexDirection: 'row'` applied to both - if that's the case could we add a new style here for `styles.badgeContainer` or something along those lines and rename `styles.title` to `styles.container` where both can just be `flexDirection: 'row'`?

junedomingo: Yeah! sure thing :-)

PR #153, opened 2017-07-23
Show zero when there are no repositories

lex111:
A small fix, because now shows an empty string:

Before | After 
| - | -
image | image

In this regard, I had this question: if there are no elements (repositories, followers, etc.), then we need to disable `onPress` handler, because we go to the [screen with an empty result](https://user-images.githubusercontent.com/4408379/28499729-4463f8a0-6fc5-11e7-8228-6cde0c5d32c8.png)?


lex111:
I just noticed that this will be fixed in #138.

PR #154, opened 2017-07-23
Fix displaying the list of organization's repositories

lex111:
Quick fix: when moving to the organization's repositories there is nothing (it broke after the changes in [this commit](https://github.com/gitpoint/git-point/commit/0eb5fd9429e1865447278b4640c1aebb33729487))

Strangely, I do not see that API returns `permissions` for repos.

1. image
2. image


lex111:
@housseindjirdeh I seem to understand, fixed, now it's true?

PR #160, opened 2017-07-24
Add number and type notification to title

lex111:
Now the notification item have this title:
image

There is no number and type (issue or PR). So I thought it would be nice to add this:
imageimage

### DIFF
--- a/routes.js
+++ b/routes.js
@@ -115,11 +115,15 @@ const sharedRoutes = {
   },
   Issue: {
     screen: IssueScreen,
-    navigationOptions: ({ navigation }) => ({
-      title: `#${navigation.state.params.issue
-        ? navigation.state.params.issue.number
-        : 'Issue'}`,
-    }),
+    navigationOptions: ({ navigation }) => {
+      const issueNumberRegex = /issues\/([0-9]+)$/;
+      const { issue, issueURL, isPR } = navigation.state.params;
+      const number = issue ? issue.number : issueURL.match(issueNumberRegex)[1];
+
+      return {
+        title: isPR ? `Pull Request #${number}` : `Issue #${number}`,
+      };
+    },
   },
   IssueSettings: {
     screen: IssueSettingsScreen,
--- a/src/notifications/screens/notifications.screen.js
+++ b/src/notifications/screens/notifications.screen.js
@@ -187,6 +187,7 @@ class Notifications extends Component {
     markAsReadByDispatch(notification.id);
     navigation.navigate('Issue', {
       issueURL: notification.subject.url.replace('pulls', 'issues'),
+      isPR: notification.subject.type === 'PullRequest',
     });
   }
 


routes.js
```diff
@@ -115,11 +115,15 @@ const sharedRoutes = {
...
-        ? navigation.state.params.issue.number
-        : 'Issue'}`,
-    }),
+    navigationOptions: ({ navigation }) => {
+      const issueNumberRegex = /issues\/([0-9]+)$/;
+      const { issue, issueURL, isPR } = navigation.state.params;
+      const number = issue ? issue.number : issueURL.match(issueNumberRegex)[1];
+
+      return {
+        title: isPR ? `Pull request #${number}` : `Issue #${number}`,
```
andrewda: I'd say `Pull Request` should be capitalized, but it depends on how it's done other places throughout the code.

lex111: @andrewda I looked at the merge button, yeah, it is capitalized, therefore did likewise :+1:

src/notifications/screens/notifications.screen.js
```diff
@@ -187,6 +187,7 @@ class Notifications extends Component {
     markAsReadByDispatch(notification.id);
     navigation.navigate('Issue', {
       issueURL: notification.subject.url.replace('pulls', 'issues'),
+      isPR: notification.subject.type === 'PullRequest',
```
housseindjirdeh: This is awesome, but the notifications screen isn't the only place to navigate to the issue screen. There are two other places in the app (`IssueListItem` component and in the `EventsScreen`). Could you update all the places in the app to have this state param? (you can most probably check for `issue.pull_request` like I do in other places of the app to identify whether an issue is actually a pull request)

lex111: @housseindjirdeh Hmm.. it's good to do, although I did not plan to do it in this PR, that's why it's called that.

I noticed that in events PR is not clickable (there is no handler `onPress`). It will also navigate to issue, although they will not have an `issue` key, then you need to pass the `pull_request` key. There will be something like this:

```js
this.props.navigation.navigate('Issue', {
  issue: userEvent.payload.issue || userEvent.payload.pull_request,
  isPR: typeof userEvent.payload.pull_request !== undefined,
});
```

This is ok?

And I get crash when I want to go to issue, although I have not done this before (or probably not on all issues):
image

andrewda: @lex111 Yea that crash is a pain in the butt to debug. I was experiencing it for a while, and eventually I think I found out that it was because I had something like

```
{ foo && <ReactElement /> }
```

where `foo` was a string instead of a boolean. IIRC changing `foo` to `!!foo` worked, but I'm not positive.

housseindjirdeh: Yeah that error message popping up gave me some grief some time ago 😢 

@andrewda is right, not sure why you're seeing the message all of a sudden as I made sure it wasn't happening in all the screens (someone may have merged in something that changes things). That workaround will actually solve it. Please feel free to fix it on this PR @lex111 I really appreciate it

lex111: @housseindjirdeh well, I rolled back and look in the master branch, this problem is there.

Remind me what I need to do? Just now PR is not clickable, and fix the title associated with it is better (in my opinion) in PR, which will do it (it will be more understandable)

housseindjirdeh: @lex111 as discussed, this PR takes care of navigating to issues from the notifications screen and you'll prefer to keep it as such. I'm okay with that, we'll just need another PR to add the `isPR` flag to other places in the app that navigates to the `Issue` screen

Regarding that bug you're seeing, let's please open a separate bug for that if we're seeing the issue screen failing because of that message

PR #166, opened 2017-07-25
Submit new issue (closes #58)

### LINKED ISSUE #58: Add submit new issue functionality
Thank you @FlorianTaut for suggesting this :)

Will be awesome to be able to do this. Here's what I'm envisioning on the `IssueListScreen`:

<p align="center">

</p>

Only in the that screen --> add a + icon to the top right corner and when clicked it takes you to a form type screen that looks just like `PullDiffScreen`. An issue title input and an issue description input and a simple list item for `Submit new issue`. 

Please take a look at how I added a `cog` icon to the top right in the `PullDiffScreen` to get an idea of how I added a top right icon with actions in the navbar.

Once the API is hit and successful - we can use the response to navigate directly to the newly submitted issue. We can rely on the awesome work done by `@mikekavouras` and show the `loading-modal` until the call is complete similar to how he handled forking in #50.

Antoine38660:
Here is the **submit new issue** (#58) PR.
> Currently `WIP`.

### Screenshots

simulator screen shot 25 juil 2017 a 15 57 33

### DIFF
--- a/routes.js
+++ b/routes.js
@@ -140,7 +140,7 @@ const sharedRoutes = {
   NewIssue: {
     screen: NewIssueScreen,
     navigationOptions: {
-      title: 'New issue',
+      title: 'New Issue',
     },
   },
   PullDiff: {
--- a/routes.js
+++ b/routes.js
@@ -49,7 +49,12 @@ import {
 } from 'repository';
 
 // Issue
-import { IssueScreen, IssueSettingsScreen, PullMergeScreen } from 'issue';
+import {
+  IssueScreen,
+  IssueSettingsScreen,
+  NewIssueScreen,
+  PullMergeScreen,
+} from 'issue';
 
 const sharedRoutes = {
   RepositoryList: {
@@ -130,6 +135,12 @@ const sharedRoutes = {
     screen: IssueSettingsScreen,
     navigationOptions: {
       title: 'Settings',
+    },
+  },
+  NewIssue: {
+    screen: NewIssueScreen,
+    navigationOptions: {
+      title: 'New issue',
     },
   },
   PullDiff: {
--- a/src/api/index.js
+++ b/src/api/index.js
@@ -316,3 +316,21 @@ export const fetchMergePullRequest = (
     })
   );
 };
+
+export const fetchSubmitNewIssue = (
+  owner,
+  repo,
+  issueTitle,
+  issueComment,
+  accessToken
+) => {
+  const ENDPOINT = `https://api.github.com/repos/${owner}/${repo}/issues`;
+
+  return fetch(
+    ENDPOINT,
+    accessTokenParametersPOST(accessToken, {
+      title: issueTitle,
+      body: issueComment,
+    }).then(response => response.json())
+  );
+};
--- a/src/api/index.js
+++ b/src/api/index.js
@@ -331,6 +331,5 @@ export const fetchSubmitNewIssue = (
     accessTokenParametersPOST(accessToken, {
       title: issueTitle,
       body: issueComment,
-    }).then(response => response.json())
-  );
+    })).then(response => response.json());
 };
--- a/src/issue/issue.action.js
+++ b/src/issue/issue.action.js
@@ -6,6 +6,7 @@ import {
   fetchEditIssue,
   fetchChangeIssueLockStatus,
   fetchMergePullRequest,
+  fetchSubmitNewIssue,
 } from 'api';
 import {
   GET_ISSUE_COMMENTS,
@@ -16,6 +17,7 @@ import {
   GET_ISSUE_MERGE_STATUS,
   MERGE_PULL_REQUEST,
   GET_ISSUE_FROM_URL,
+  SUBMIT_NEW_ISSUE,
 } from './issue.type';
 
 const getDiff = url => {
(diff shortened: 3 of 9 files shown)


Antoine38660:
@andrewda as you can see, I created a new screen and new route. I also tried to add some code inside the `api` and `actions` files but I did some copy/paste : 
1. `api`: should I put the `.then(...)` ?
2. `actions` & `types`: I copied/pasted the `MERGE_PULL_REQUEST` ones.

src/issue/issue.reducer.js
```diff
@@ -165,6 +168,23 @@ export const issueReducer = (state = initialState, action = {}) => {
...
       };
+    case SUBMIT_NEW_ISSUE.PENDING:
+      return {
+        ...state,
+        isPendingSubmitting: true,
+      };
+    case SUBMIT_NEW_ISSUE.SUCCESS:
+      return {
+        ...state,
+        isSubmitted: action.payload,
```
housseindjirdeh: This is so awesome. The only thing I can think of is I don't think we need to store `isSubmitted` into our redux store. I think the `SUCCESS` handler can just set `isSubmitted` back to false but at the same time also assign the `issue` parameter to our new issue object (creating an issue [returns the new issue object](https://developer.github.com/v3/issues/#response-3) as the response)

src/issue/issue.action.js
```diff
@@ -242,3 +244,36 @@ export const mergePullRequest = (
...
+
+    return fetchSubmitNewIssue(
+      owner,
+      repo,
+      issueTitle,
+      issueComment,
+      accessToken
+    )
+      .then(data => {
+        dispatch({
```
housseindjirdeh: This is where you can potentially navigate to the issue screen with the `issue` response :)

src/issue/screens/new-issue.screen.js
```diff
@@ -0,0 +1,119 @@
...
+    ...fonts.fontPrimary,
+  },
+});
+
+const mapStateToProps = state => ({
+  repository: state.repository.repository,
+});
+
+const mapDispatchToProps = dispatch => ({
+  submitNewIssueByDispatch: (
```
housseindjirdeh: Looks like all you really have left to do is map this action to the screen :)

Similarly to `PullMergeScreen`, what we can do is just have a `ListItem` after the `Section Lists` called `Create New Issue` and clicking it will either:

1. Call `submitNewIssueByDispatch` and after it's completed (similarly to [mergePull](https://github.com/gitpoint/git-point/blob/master/src/issue/screens/pull-merge.screen.js#L128-L130), call navigation.navigate to `Issue` with the issue response returned from the `submitNewIssue` call.
2. Call `submitNewIssueByDispatch` and **within**  `submitNewIssue` right after the fetch call is made, do the navigation

I would prefer to have the navigation logic in our components and outside of our action files (option 1) _but_ if it's easier to have the navigation logic within the action method itself I'm okay with it :)

Antoine38660:
Thank you so much @housseindjirdeh <3

simulator screen shot 27 juil 2017 a 00 32 13

src/issue/screens/new-issue.screen.js
```diff
@@ -0,0 +1,164 @@
...
+              placeholderTextColor={colors.grey}
+              style={[
+                styles.textInput,
+                { height: Math.max(60, this.state.issueTitleHeight) },
+              ]}
+              value={issueTitle}
+            />
+          </SectionList>
+
+          <SectionList title="Issue comment">
```
andrewda: Capitalize the "C" for title case.

src/repository/screens/repository.screen.js
```diff
@@ -259,13 +260,18 @@ class Repository extends Component {
...
             }
-            showButton={pureIssues.length > 0}
-            buttonTitle="View All"
-            buttonAction={() =>
-              navigation.navigate('IssueList', {
-                type: 'issue',
-                issues: pureIssues,
-              })}
+            showButton
+            buttonTitle={pureIssues.length > 0 ? 'View All' : 'New issue'}
```
andrewda: Capitalize the "I" in "issue" for title case.

routes.js
```diff
@@ -132,6 +137,12 @@ const sharedRoutes = {
       title: 'Settings',
     },
   },
+  NewIssue: {
+    screen: NewIssueScreen,
+    navigationOptions: {
+      title: 'New issue',
```
andrewda: Capitalize "i" in "issue" for title case.

src/issue/screens/new-issue.screen.js
```diff
@@ -0,0 +1,164 @@
...
+              placeholderTextColor={colors.grey}
+              style={[
+                styles.textInput,
+                { height: Math.max(60, this.state.issueCommentHeight) },
+              ]}
+              value={issueComment}
+            />
+          </SectionList>
+
+          <SectionList title="Create new issue">
```
andrewda: I think the "Submit" button is pretty self-explanatory, so the "Create New Issue" text doesn't need to be repeated.

src/api/index.js
```diff
@@ -316,3 +316,20 @@ export const fetchMergePullRequest = (
     })
   );
 };
+
+export const fetchSubmitNewIssue = (
```
andrewda: Hmm, this method's name is a little funky. `submitNewIssue` might make more sense.

Antoine38660: I agree with you but I was inspired by the `fetchPostIssueComment` ([L182](https://github.com/gitpoint/git-point/blob/master/src/api/index.js#L182)) function name. For consistency questions I chose to keep the `fetch` keyword :)

andrewda: Yep good point, you're right!

Antoine38660:
Finally: 
simulator screen shot 28 juil 2017 a 08 52 08

PR #173, opened 2017-07-26
Feature: Share repository

RolfKoenders:
## Changes
- Added a 'share' action to the actionsheet on the repository screenshot. This uses the native share window, so you have all your available options shown. See the screenshots. (My simulator only has twitter) 
I added the actual repository link to the message, so its visible to the user that it also shares the url. If the `url` key is used in the share method it doesn't show the url. I thought that it might be strange for the user. 
- Removed unicode star. Since we don't have the octicons in the actionsheet yet, i removed the star since it doesn't look good otherwise.

## Screenshots
Share-screen1
Share-screen2
Share-screen3

### DIFF
--- a/src/repository/screens/repository.screen.js
+++ b/src/repository/screens/repository.screen.js
@@ -161,15 +161,13 @@ class Repository extends Component {
     const openIssues = pureIssues.filter(issue => issue.state === 'open');
     const closedIssues = pureIssues.filter(issue => issue.state === 'closed');
 
-    const repositoryActions = [starred ? 'Unstar' : 'Star'];
     const showFork =
       repository && repository.owner && repository.owner.login !== username;
-
-    if (showFork) {
-      repositoryActions.push('Fork');
-    }
-
-    repositoryActions.push('Share');
+    const repositoryActions = [
+      starred ? 'Unstar' : 'Star',
+      showFork && 'Fork',
+      'Share',
+    ];
 
     const loader = isPendingFork ? <LoadingModal /> : null;
 
--- a/src/repository/screens/repository.screen.js
+++ b/src/repository/screens/repository.screen.js
@@ -1,6 +1,6 @@
 import React, { Component } from 'react';
 import { connect } from 'react-redux';
-import { StyleSheet } from 'react-native';
+import { StyleSheet, Share } from 'react-native';
 import { ListItem } from 'react-native-elements';
 import ActionSheet from 'react-native-actionsheet';
 
@@ -99,11 +99,15 @@ class Repository extends Component {
       changeStarStatusRepoByDispatch,
       forkRepoByDispatch,
       navigation,
-      username,
     } = this.props;
-    const showFork = repository.owner.login !== username;
 
     if (index === 0) {
+      // Share
+      this.shareRepository(repository);
+    }
+
+    if (index === 1) {
+      // Star
       changeStarStatusRepoByDispatch(
         repository.owner.login,
         repository.name,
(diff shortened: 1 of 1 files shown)


src/repository/screens/repository.screen.js
```diff
@@ -145,7 +166,7 @@ class Repository extends Component {
     const openIssues = pureIssues.filter(issue => issue.state === 'open');
     const closedIssues = pureIssues.filter(issue => issue.state === 'closed');
 
-    const repositoryActions = [starred ? '★ Unstar' : '★ Star'];
+    const repositoryActions = ['Share', starred ? 'Unstar' : 'Star'];
```
andrewda: I think it might be more natural for "Share" to be at the bottom of the list, that's usually where I'm used to seeing it.

andrewda:
This is a really good idea!

src/repository/screens/repository.screen.js
```diff
@@ -99,25 +99,46 @@ class Repository extends Component {
...
+    if (index === 1) {
+      // Star
       changeStarStatusRepoByDispatch(
         repository.owner.login,
         repository.name,
         starred
       );
     }
 
-    if (index === 1 && showFork) {
```
housseindjirdeh: I think we still need to check for `showFork` so that the fork button doesn't show for any repository of the current authenticated user

src/repository/screens/repository.screen.js
```diff
@@ -1,6 +1,6 @@
 import React, { Component } from 'react';
 import { connect } from 'react-redux';
-import { StyleSheet } from 'react-native';
+import { StyleSheet, Share } from 'react-native';
```
housseindjirdeh: I didn't know there was a Share API! Amazing :tada:

src/repository/screens/repository.screen.js
```diff
@@ -99,25 +99,46 @@ class Repository extends Component {
...
-      username,
     } = this.props;
-    const showFork = repository.owner.login !== username;
 
     if (index === 0) {
+      // Share
+      this.shareRepository(repository);
+    }
+
+    if (index === 1) {
```
housseindjirdeh: I know this was already here in this format, but any chance you can just change it to `if`, `else if` blocks instead of multiple separate `if` conditions :O

src/repository/screens/repository.screen.js
housseindjirdeh: Oops nvm, there's already a check to not have it included in line 173, we're good :+1:

src/repository/screens/repository.screen.js
housseindjirdeh: Same, nothing extremely important however but I also tend to see it in the bottom.

src/repository/screens/repository.screen.js
RolfKoenders: 😃

src/repository/screens/repository.screen.js
```diff
@@ -145,14 +161,16 @@ class Repository extends Component {
...
-    const repositoryActions = [starred ? '★ Unstar' : '★ Star'];
+    const repositoryActions = [starred ? 'Unstar' : 'Star'];
     const showFork =
       repository && repository.owner && repository.owner.login !== username;
 
     if (showFork) {
       repositoryActions.push('Fork');
     }
 
+    repositoryActions.push('Share');
```
andrewda: Could this all be simplified to:

```javascript
const showFork = repository && repository.owner && repository.owner.login !== username;
const repositoryActions = [starred ? 'Unstar' : 'Star', showFork && 'Fork', 'Share'];
```

Have no idea if that would work but might be worth trying it because it's a little hard to read as-is.

RolfKoenders: Yea true, i was for now keeping it as it was. But this looks good! Will update it.

RolfKoenders:
@andrewda Yea right!? I actually wanted to share a repository i found but there was no way to easily copy the link or share it. So i thought, lets add that! 😄

PR #174, opened 2017-07-27
Onboarding swiper

### LINKED ISSUE #51: Add onboarding swiper as welcome screen
Our current welcome screen before a user logs in:

<p align="center">

</p>

I quite like it, but I think having a welcome swiper explaining the different things the app could do would be even better 🤓 

For example: the Slack mobile app

image

The Product Hunt app

<p align="center">

</p>

Doesn't have to be as anything as complicated as those options - they're just for reference.

Some points however:

1. [react-native-swiper](https://github.com/leecade/react-native-swiper) seems to be an excellent library that can do just this
2. I like the idea of having the Sign In button at the bottom always so even though the user can continue to swipe, they can always just click the sign in button
3. I'll leave the floor open for design suggestions. Something simple with a top title (`Control your notifications` for example, a GitHub [octicon](https://octicons.github.com/) in the middle and a mini blurb below (`View and control all of your unread and participating notifications.` for example is what I'm envisioning. Feel free to just lift copy from [the landing page](https://gitpoint.co)

Always open to further ideas and suggestions of course <3

housseindjirdeh:
Adds onboarding swiper. Closes #51.

onboarding

### DIFF
--- a/package.json
+++ b/package.json
@@ -64,6 +64,7 @@
     "react-native-photo-view": "^1.3.0",
     "react-native-safari-view": "^2.0.0",
     "react-native-search-bar": "^3.0.0",
+    "react-native-swiper": "^1.5.4",
     "react-native-syntax-highlighter": "^1.2.1",
     "react-native-vector-icons": "^4.0.0",
     "react-navigation": "^1.0.0-beta.11",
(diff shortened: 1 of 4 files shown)


housseindjirdeh:
<3

PR #183, opened 2017-07-29
update welcome screen, change onboarding swiper colors slightly

### LINKED ISSUE #181: Welcome screen change
After the introduction of onboarding swiper in PR #174, the question arises, do you really need a welcome screen at all? After entering the application, you see this screen:

image

But what's the point in it? I understand, so we welcome the user, I'm not against it, but maybe then let's make it automatically enter the application (e.g. in 5 seconds),  and not by pressing the enter button?
It's not very convenient, you have to press it every time after login, especially considering that except to enter, nothing more can be done.

I would also suggest somehow redesign of this screen, for example add an animation background or something like that, it would be cool! And we can change the text, eg "Welcome back to GitPoint!", if the user has previously been authorized with app (optional).

What thoughts?

housseindjirdeh:
Closes #181. Updates welcome screen so user is automatically navigated when request finishes firing:

welcome-screen-update

Also:
* improved colors used in onboarding swiper
* created a `resetNavigationTo` helper method to simplify things

### DIFF
--- a/src/auth/auth.action.js
+++ b/src/auth/auth.action.js
@@ -1,5 +1,7 @@
 import { AsyncStorage } from 'react-native';
 import uniqby from 'lodash.uniqby';
+import { resetNavigationTo } from 'utils';
+
 import {
   fetchAccessToken,
   fetchAuthUser,
@@ -15,7 +17,7 @@ import {
   GET_EVENTS,
 } from './auth.type';
 
-export const auth = (code, state) => {
+export const auth = (code, state, navigation) => {
   return dispatch => {
     dispatch({ type: LOGIN.PENDING });
 
@@ -25,6 +27,8 @@ export const auth = (code, state) => {
           type: LOGIN.SUCCESS,
           payload: data.access_token,
         });
+
+        resetNavigationTo('Main', navigation);
       })
       .catch(error => {
         dispatch({
--- a/src/auth/auth.action.js
+++ b/src/auth/auth.action.js
@@ -1,6 +1,6 @@
 import { AsyncStorage } from 'react-native';
 import uniqby from 'lodash.uniqby';
-import { resetNavigationTo } from 'utils';
+import { delay, resetNavigationTo } from 'utils';
 
 import {
   fetchAccessToken,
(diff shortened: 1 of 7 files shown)


housseindjirdeh:
@andrewda yeah I was thinking the same thing, considered just removing the title message entirely and only having a loading spinner but I think having a minimum timer could be okay as well.

Not sure if it's possible to fade away, agreed that would be nicer. Will investigate.

lex111:
@housseindjirdeh I like the idea with the spinner, so it will be clear.

src/auth/auth.action.js
```diff
@@ -21,7 +21,7 @@ export const auth = (code, state, navigation) => {
   return dispatch => {
     dispatch({ type: LOGIN.PENDING });
 
-    fetchAccessToken(code, state)
+    delay(fetchAccessToken(code, state), 2000)
```
andrewda: Should this be a constant or just keep in hard coded?

housseindjirdeh: Hmm are you thinking of having `2s` as a specific constant like `loadDelay`? Personally feel with the `delay` method in `utils`, might be better to have the freedom to use any number we like if we wish to delay other async actions.

What do you think?

housseindjirdeh:
Updated: added a `delay` util method to allow for a _whichever comes first_ promise wrapper.

welcome-update-2

Unfortunately I don't think [`NavigationActions`](https://reactnavigation.org/docs/navigators/navigation-actions), allows for anything aside from a _stack_ style of navigation so don't know if making it fade would be possible :thinking:

src/auth/auth.action.js
andrewda: Yep, sounds good!

andrewda:
@housseindjirdeh Not sure if this would work but: https://github.com/react-community/react-navigation/issues/1490#issuecomment-306009554

housseindjirdeh:
Nice I'll definitely give that a shot. Seems to be an iOS fade solution so would probably need to dig deeper to see if that could be applied to Android as well.

Definitely not going to prioritize that today/tomorrow however. I think the way it currently is should suffice for the welcome screen until we get the time to look into integrating fade animations :raised_hands:

PR #184, opened 2017-07-29
Watch repos #56

### LINKED ISSUE #56: Watch repos
I see that right now is not possible to watch any repos (only star repos is available)

I'd like to work on this feature too.

Thanks!

alejandronanez:
closes #56


src/repository/screens/repository.screen.js
```diff
@@ -95,12 +104,14 @@ class Repository extends Component {
   handlePress = index => {
     const {
       starred,
+      subscribed,
```
alejandronanez: I've found this `handlePress` method too fragile - depend on the index is not a good thing at all. We should improve this code.

housseindjirdeh: I very very very much agree that we should improve this, not the cleanest way to do things to be honest _especially_ when adding conditional items like `fork`

PR #186, opened 2017-07-30
add localization

housseindjirdeh:
Adds localization. Still need to:

1. Replace all copy in app with translation strings

EDIT:

PR up and ready for review. Adds a user profile screen where user can now modify their language (for every language file we add, we'll need to add its respective item to the list of languages in that screen).

languagegif


lex111:
We need to come up with a convention for the names of keys. (e.g. `events.noneTitle`)

housseindjirdeh:
@lex111 agreed. I was envisioning having it layered two fold based on the module + screen. For example:

```
export const en = {
  auth: {
    events: {
      //...keys for events screen
    },
    profile: {
      //...keys for auth profile screen
    },
    //...
  },
  issue: {
    settings: {
      // ... keys for issue settings screen
    },
    main: {
      // ... keys for issue screen
    }
  },
  // ...
};
```

src/notifications/screens/notifications.screen.js
```diff
@@ -344,19 +353,27 @@ class Notifications extends Component {
...
           {!this.isLoading() &&
             this.notifications().length === 0 &&
             <View style={styles.textContainer}>
               <Text style={styles.noneTitle}>
-                {`You don't have any${type === 0
-                  ? ' unread'
-                  : type === 1 ? ' pending' : ''} notifications!`}
+                {translate('notifications.main.noneMessage', language, {
+                  notificationType:
+                    type === 0 ? ' unread' : type === 1 ? ' participating' : '',
```
lex111: This also probably needs to be move to a local file?

src/notifications/screens/notifications.screen.js
```diff
@@ -344,19 +353,27 @@ class Notifications extends Component {
...
-                ? 'unread'
-                : type === 1 ? 'pending' : 'all'} notifications`}
+              text={translate(
+                'notifications.main.retrievingMessage',
+                language,
+                {
+                  notificationType:
+                    type === 0
+                      ? 'unread'
+                      : type === 1 ? 'participating' : 'all',
```
lex111: This also probably needs to be move to a local file?

src/issue/screens/issue-settings.screen.js
```diff
@@ -265,30 +287,39 @@ class IssueSettings extends Component {
           ref={o => {
             this.IssueActionSheet = o;
           }}
-          title={`Are you sure you want to ${issue.state === 'open'
-            ? 'close'
-            : 'reopen'} this?`}
-          options={['Yes', 'Cancel']}
+          title={translate('issue.settings.issueActionPrompt', language, {
+            action: issue.state === 'open' ? 'close' : 'reopen',
```
lex111: And here?

src/issue/screens/issue-settings.screen.js
```diff
@@ -265,30 +287,39 @@ class IssueSettings extends Component {
...
         <ActionSheet
           ref={o => {
             this.LockIssueActionSheet = o;
           }}
-          title={`Are you sure you want to ${issue.locked
-            ? 'unlock'
-            : 'lock'} this conversation?`}
-          options={['Yes', 'Cancel']}
+          title={translate('issue.settings.lockIssuePrompt', language, {
+            action: issue.locked ? 'unlock' : 'lock',
```
lex111: And here too?

src/utils/method-helpers.js
```diff
@@ -28,3 +29,6 @@ const sleep = ms => new Promise(resolve => setTimeout(resolve, ms));
 
 export const delay = (delayed, ms) =>
   Promise.all([delayed, sleep(ms)]).then(([data]) => data);
+
+export const translate = (key, lang, interpolation = null) =>
```
lex111: Can you avoid passing the parameter of the current language (`lang`)? Calculate the language that the user chose or which is specified in the settings and use it?

src/notifications/screens/notifications.screen.js
housseindjirdeh: Cheers, fixed it by just simplifying the message and removing the interpolation.

src/notifications/screens/notifications.screen.js
housseindjirdeh: Fixed (removed entirely)

src/issue/screens/issue-settings.screen.js
housseindjirdeh: Fixed

src/issue/screens/issue-settings.screen.js
housseindjirdeh: Fixed

src/utils/method-helpers.js
housseindjirdeh: Yes thanks for bringing this up. So there were two things I considered:

1. Find the language locale from the user's device and use it directly. Don't allow the user to change the language in the app and if they want to they would have to change it on their device.
2. Allow the user to change the language on their device.

The pros of 1) is that we wouldn't have to pass in `lang` the way I do now in every copy in the entire app. Just calling the translate function should suffice _I think_.

The issue with this however is that the user may not even know different languages if their device language isn't supported. I kinda think it's nicer to allow the user to change the language on the app directly.

So to allow for this, I added a `lang` parameter to `authReducer`. Throughout the entire app, I just pass this parameter down to this method and when the language is changed in the `userProfileScreen` - this gets reflected in all the copy in the app immediately.

What do you think, open to all kinds of suggestions and advice. Thanks for the help reviewing this PR mate :raised_hands:

lex111: @housseindjirdeh well, as I see: we determine the language of the device, if such is supported by the application, we use it. Further, if the user has changed the language in the application settings (we save it via AsyncStorage), then we use it (ie the language in the application settings has priority over the device language). 

If this language is not supported by the application, then English is set by default. When the user opens  the application again, we look whether there is a selected language in the application settings (via AsyncStorage), if there is one we use it.

> Just calling the translate function should suffice I think.

This is what I had in mind in order to discard this parameter.

housseindjirdeh: Ah I see what you mean. I agree it would definitely be nicer not having to pass it down as a parameter _every time_. The only thing is I have it as part of the redux store just like every other attribute in the app so it makes it easy to see how a language is changed and so forth. As you mentioned, I think we could manually set it to `AsyncStorage` on language change and then manually retrieve that parameter from there in the translate method but that sort of abstracts things away from the redux store (and I kinda like keeping everything attached to the store if possible).

Will you be okay with having the way it is now and opening a separate refactor ticket to discuss? Definitely think I may be overthinking things and your approach or something similar but be nicer but I'm not sure.

lex111: Hmm, I'm not exactly sure, but maybe it will help somehow, like an example https://www.npmjs.com/package/react-localize-redux

housseindjirdeh: Hmmm that might come in handy, will investigate using something like that to simplify things this week hopefully

PR #191, opened 2017-08-01
Hide tabbar when scrolling down on events screen

lex111:
Sorry, could not make the gif, so did a [demo video](https://files.gitter.im/git-point/Lobby/QBGj/demo1.mp4). The idea was to hide the tabbar when scrolling down on the events screen. When scrolling back up, it is shown again.

With animation there are problems, it is not so smooth now. Maybe someone knows why? Also need to check at it on iOS. I will be grateful for any help, thank you!

### DIFF
--- a/src/auth/screens/events.screen.js
+++ b/src/auth/screens/events.screen.js
@@ -8,7 +8,7 @@ import moment from 'moment';
 import { LoadingUserListItem, UserListItem, ViewContainer } from 'components';
 import { colors, fonts, normalize } from 'config';
 import { emojifyText, translate } from 'utils';
-import { getUserEvents } from '../auth.action';
+import { getUserEvents, changeTabBarVisibility } from '../auth.action';
 
 moment.updateLocale('en', {
   relativeTime: {
@@ -32,10 +32,13 @@ const mapStateToProps = state => ({
   userEvents: state.auth.events,
   language: state.auth.language,
   isPendingEvents: state.auth.isPendingEvents,
+  isTabBarVisible: state.auth.isTabBarVisible,
 });
 
 const mapDispatchToProps = dispatch => ({
   getUserEvents: user => dispatch(getUserEvents(user)),
+  changeTabBarVisibility: isTabBarVisible =>
+    dispatch(changeTabBarVisibility(isTabBarVisible)),
 });
 
 const styles = StyleSheet.create({
@@ -85,6 +88,13 @@ const styles = StyleSheet.create({
 });
 
 class Events extends Component {
+  constructor(props) {
+    super(props);
+    this.state = {
+      scrollOffset: 0,
+    };
+  }
+
   componentDidMount() {
     if (this.props.user.login) {
       this.getUserEvents(this.props.user);
@@ -475,6 +485,17 @@ class Events extends Component {
     return item.id;
   };
 
+  changeTabBarVisibility = event => {
+    const scrollOffset = event.nativeEvent.contentOffset.y;
+    const isScrollDown = scrollOffset > this.state.scrollOffset;
+
+    this.setState({
+      scrollOffset,
+    });
+
+    this.props.changeTabBarVisibility(!isScrollDown);
+  };
+
   renderDescription(userEvent) {
     return (
       <Text style={styles.descriptionContainer}>
@@ -526,6 +547,8 @@ class Events extends Component {
           onRefresh={this.getUserEvents}
           refreshing={isPendingEvents}
           keyExtractor={this.keyExtractor}
+          onScroll={this.changeTabBarVisibility}
+          onEndReachedThreshold={1200}
           renderItem={({ item }) =>
             <View>
               <UserListItem
--- a/src/components/tab-bar.component.js
+++ b/src/components/tab-bar.component.js
@@ -1,0 +1,113 @@
+/* eslint-disable no-unused-expressions */
+
+import React, { Component } from 'react';
+import { connect } from 'react-redux';
+import { Animated, LayoutAnimation } from 'react-native';
+import { TabBarBottom, NavigationActions } from 'react-navigation';
+
+import { animations } from 'config';
+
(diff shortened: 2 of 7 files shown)


RolfKoenders:
Cool! Really like the idea to hide the tabbar when scrolling.

Antoine38660:
Looks nice!

This is the render on iOS:
hide tabbar ios


As you can see, the tabbar disappear sometimes with no reason... I have no idea why...

_FYI: I use this tool https://ezgif.com/video-to-gif to convert video to gif (and also to optimize it, when too large) : I recommend it._

andrewda:
Hmm... Interesting. I think this has the potential to look really good, but also be pretty annoying.

Personally, I like the tab bar always being there. It feels more native to me allows me to navigate somewhere else without any extra hassle. The only other iOS app that I know does this is Safari, and even then the bar at the bottom of Safari has a much different use case and is used much less frequently than this bar. Our navigation here is a lot more like the navigation in the Twitter, Instagram, Facebook and Reddit apps, all of which have a static tab bar. In a way, a dynamic bar kinda feels more webapp-like, similar to when the header disappears as you start to scroll - not sure how much I like that.

lex111:
The bottom line is to focus on content, events in this case. You've probably seen how the site hides a header, when scrolling down to free up more space, and then reappears when you scroll up.

@Antoine38660 cool, thanks for the animated screenshot! Well, to understand why this is happening :disappointed: 

@andrewda We can put this in the settings, for example. By default, it will be enabled, but it will be possible to set the tabbar never to hide, how do you do it?

housseindjirdeh:
I agree with @andrewda in that having the tabbar disappearing on any screen on scroll down could be more of a problem then something useful. _However_, if it's only limited to the events screen and works fluidly I'm okay with that.

Will take a look at the code soon and let you know if I can spot why that issue is happening @lex111  (and will also try on my device 👍 )

andrewda:
@lex111 Sure that sounds good! Though I'd vouch for _not_ hiding by default because it's the more standard style.

housseindjirdeh:
Agreed we could have a check option in the auth user settings screen to turn it on. Do you think it would be worthwhile to have three options: None, Events Screen only and Everywhere?

housseindjirdeh:
Bah we really need to allow comment editing 😭. Wanted to tag both you @andrewda @lex111 to get your opinion on that ^.

lex111:
Hiding the tabbar was supposed _only_ on the screen of events, on other screens, I think it is not necessary. Therefore, in the settings, I would just leave only two options: None and Events Screen. More precisely, then, the title is "Hide tabbar when scrolling down on events screen" - and options: Yes and No (default)

It would be good to solve the problem described by @Antoine38660, I hope for your help :sos:

henrikra:
I think the TabBar should not be hidden. It is good as it is now :)

src/auth/screens/events.screen.js
```diff
@@ -85,6 +88,13 @@ const styles = StyleSheet.create({
 });
 
 class Events extends Component {
+  constructor(props) {
```
alejandronanez: You don't need props in here =]

src/auth/screens/events.screen.js
```diff
@@ -475,6 +485,17 @@ class Events extends Component {
     return item.id;
   };
 
+  changeTabBarVisibility = event => {
+    const scrollOffset = event.nativeEvent.contentOffset.y;
```
alejandronanez: Are we sure we will not get any undefined in this object? Maybe we can use `lodash.get` just to be sure :)
If you're 100% sure, disregard my comment

src/auth/screens/events.screen.js
```diff
@@ -526,6 +547,8 @@ class Events extends Component {
           onRefresh={this.getUserEvents}
           refreshing={isPendingEvents}
           keyExtractor={this.keyExtractor}
+          onScroll={this.changeTabBarVisibility}
+          onEndReachedThreshold={1200}
```
alejandronanez: Can you move this `1200` to a constant with a meaningful name? Also why `1200` instead of `1000` for example?

src/components/tab-bar.component.js
```diff
@@ -0,0 +1,113 @@
...
+import { animations } from 'config';
+
+const mapStateToProps = state => {
+  return {
+    isTabBarVisible: state.auth.isTabBarVisible,
+  };
+};
+
+class TabBarComponent extends Component {
+  props: {
```
alejandronanez: Sorry about asking this but what is the purpose of this?
I mean, I don't see any `@flow` on top of the file and I don't see this in `propTypes`

src/components/tab-bar.component.js
```diff
@@ -0,0 +1,113 @@
...
+      const expandedHeight = event.nativeEvent.layout.height;
+
+      this.state = {
+        animatedHeight: new Animated.Value(expandedHeight),
+        expandedHeight,
+      };
+    }
+  }
+
+  changeHeight(from, to) {
```
alejandronanez: You should change this to `changeHeight = (from, to) => { ... }` and avoid calling arrow functions in the render methods.

src/components/tab-bar.component.js
```diff
@@ -0,0 +1,113 @@
...
+  render() {
+    return (
+      <Animated.View
+        style={{ height: this.state.animatedHeight }}
+        onLayout={event => this.setExpandedHeight(event)}
+      >
+        <TabBarBottom
+          {...this.props}
+          jumpToIndex={index => {
+            const { dispatch, state } = this.props.navigation;
```
alejandronanez: Please, move this into a class method instead.

lex111:
@alejandronanez yes, you're right, it's still WIP, so there are imperfections, but there is a problem with smooth animation (see [comment](https://github.com/gitpoint/git-point/pull/191#issuecomment-319570952)), can you help? Thank you!

PR #193, opened 2017-08-02
fix(localization): various localization improvements

lex111:
- [x] Add `searchType` to localization
- [x] Localization `updateText` when change language + minor fixes
- [x] Update title on options screen when change language
- [x] Add `{{payload}}` (`action` and `ref_type`,) to localization file for events screen (eg, "created", "edited", "moved" or "deleted", "commented", etc). See details [here](https://developer.github.com/v3/activity/events/types/). (Important)
- [x] Add support for localization for `moment` (**done, but needs review!**)
- [x] Add dynamic language output ([view code](https://github.com/gitpoint/git-point/blob/localization-fixes/src/auth/screens/user-options.screen.js#L61-L68)). (i.e. so that we do not add it manually each time when a new language is supported).
Possible solution: export all available languages (locale/languages/index.js). A new key is added: `originalName` (or simply `title`) - this is the national name of the language (it will be displayed). Or something like this (information about the language at the beginning of each language file):
- [x] Localize `Members` and `Description` on organization screen
- [x] Localize `Stars`, `Forks`, `forked from`, `View All` (PR list) on repository screen
- [x] Localize `Open`, `Close` on PR screen
- [x] Localize `Post` (comment input component)
- [x] Localize title for PR's and Issues

Do not hesitate to update this PR, if you want!


lex111:
I did dynamic output of languages (from file `language-settings.js`), look, please.

Also I suggest to show native names of languages, instead of in English, i.e. Deutsch instead of German, for example, how do you like this proposal?

src/auth/screens/events.screen.js
```diff
@@ -1,30 +1,20 @@
...
-    h: '%dh',
-    hh: '%dh',
-    d: '%dd',
-    dd: '%dd',
-    M: '%dmo',
-    MM: '%dmo',
-    y: '%dy',
-    yy: '%dy',
-  },
+const language = I18n.locale.substr(0, 2);
```
lex111: I have done so for the time being, but this is essentially the initialization of the `moment`, and it should not be executed here (where can it be moved?), it first. And secondly, only the device language is taken into account here, but if the language was changed in the settings, then the date/time language (ie `moment`) will remain unchanged (we must also take into account the language in the storage). What can we do about it?

lex111:
> Update title on options screen when change language

How can we re-render the title when changing the language?
image

lex111:
Also, look at [this commit](https://github.com/gitpoint/git-point/pull/193/commits/8e1e310ffbaa9f8665af2da68c4026a70e3f659b), I did support the localization of events better, it looks like it worked out well, but I want to know your opinion :relieved:

Antoine38660:
@lex111 you said:
>> Update title on options screen when change language
>
>I need help! How can we re-render the title when changing the language?
>image


Two "solutions":
- Wait for this one https://github.com/react-community/react-navigation/issues/2330 (can be logged in another issue)
- I have never tested it yet, but the in the issue 2330, they mention [NavigationActions.SetParams](https://reactnavigation.org/docs/navigators/navigation-actions#SetParams). Maybe it can work for us.

lex111:
@Antoine38660 thanks, but unfortunately did not work `NavigationActions.SetParams` for me, the title update did not come through.

lex111:
@Antoine38660 hurraaay, I did it, I misunderstood how to use `setParams`, in general, you can shoot a smiley of disappointment, it's works! :smile:

Antoine38660:
Awesome dude! 🎉 🎉 🎉 

11/11 checks. I'll look at your code tonight 😃

src/auth/screens/events.screen.js
lex111: I solved this problem, you can see in [this commit](https://github.com/gitpoint/git-point/pull/193/commits/bc73e882087340041f4418ccfef1b1a74d405ae7). Now the configuration of the locale is made in `App.js`, as it probably should be there and localization works as I described it in [my comment](https://github.com/gitpoint/git-point/pull/193/commits/bc73e882087340041f4418ccfef1b1a74d405ae7).

In my opinion, it turned out pretty good, as usual, I'm waiting for your feedback :raised_hands:

housseindjirdeh:
Woops I'm sorry if I merged it too prematurely in case you wanted to take a look as well @Antoine38660. Please let us know if you see anything that should be changed and we can address it in a future PR

PR #195, opened 2017-08-03
Remove unnecessary permissions for Android: Phone and Storage

ferrannp:
React Native still adds two unnecessary permissions to all Android apps. Phone one is specially sensitive as is labelled as "dangerous": https://developer.android.com/guide/topics/permissions/requesting.html#normal-dangerous

If you have a dangerous permission, you are obliged to write a privacy policy for your app (referring why you need and how you use such permission).

You can read more about it in this react-native issue https://github.com/facebook/react-native/issues/5886.

This PR just removes those unnecessary permissions.


.all-contributorsrc
```diff
@@ -211,6 +211,15 @@
       "contributions": [
         "doc"
       ]
+    },
+    {
+      "login": "ferrannp",
+      "name": "Ferran Negre",
```

housseindjirdeh:
Already included and deployed on [our latest release](https://play.google.com/store/apps/details?id=com.gitpoint&hl=en).

Thanks again for this @ferrannp 🙏

PR #198, opened 2017-08-04
chore(dev): add reactotron for inspection

patw0929:
Please refer this project: [reactotron](https://github.com/infinitered/reactotron).

Reactotron is cross-platform react/react-native inspector.
It might be helpful when developing. We can see not only the redux actions, store structure & content, but also the API response (including response content/header & request header) in our app.

Screenshots:

- API Response:

(If we want to see the network response content, it need to turn off the debug mode. Seems their bug.)

- store:

### DIFF
--- a/package.json
+++ b/package.json
@@ -5,10 +5,13 @@
   "scripts": {
     "start": "react-native start",
     "start:logger": "LOGGER_ENABLED=true react-native start --reset-cache",
+    "start:tron": "TRON_ENABLED=true react-native start --reset-cache",
     "start:android": "concurrently -r 'react-native start' 'react-native run-android'",
     "start:android:logger": "LOGGER_ENABLED=true concurrently -r 'react-native start --reset-cache' 'react-native run-android'",
+    "start:android:tron": "TRON_ENABLED=true concurrently -r 'react-native start --reset-cache' 'react-native run-android'",
     "start:ios": "concurrently -r 'react-native start' 'react-native run-ios'",
     "start:ios:logger": "LOGGER_ENABLED=true concurrently -r 'react-native start --reset-cache' 'react-native run-ios'",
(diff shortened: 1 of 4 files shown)


lex111:
Wow, cool, I always wanted to try it, now there will be an opportunity!

Antoine38660:
Seems really cool! I never used it too, so good opportunity yes.

Is not it going to slow down the execution in dev mode? I know we had to remove the `redux-devtools` for this reason.

patw0929:
The `reactotron` is not like `remote-redux-devtools` which is running in browser (Chrome extensions), it is a standalone app.
And I think the speed is ok for me. Could anybody help to try it? 😄 Thanks!

package.json
```diff
@@ -67,6 +67,8 @@
     "react-navigation": "^1.0.0-beta.11",
     "react-redux": "^5.0.2",
     "react-syntax-highlighter": "^5.6.2",
+    "reactotron-react-native": "^1.12.2",
+    "reactotron-redux": "^1.12.2",
```
housseindjirdeh: Probably makes more sense to have both as part of `devDependencies`

housseindjirdeh:
Oooooooh I've stumbled on `reactotron` before and it looked awesome :O

Will pull this down and give it a shot tonight. If everyone's okay with having it included in our app as part of our development workflow - I'm okay with that as well :)

I would also like to possibly have it as part of a separate script command similar to the amazing work @lex111 did with `logger`. Something like `yarn:start:ios:tron` to run boot the application with reactotron based on a environment variable. What do you think?

Antoine38660:
I totally agree with you, a specific  command like `yarn:start:ios:tron` is a very good idea

patw0929:
Good idea. 👍  I have added the `yarn:start:ios:tron` & `yarn:start:android:tron` command in `packages.json`.

housseindjirdeh:
Awesome thanks a million matey :+1: 

Excited to try this out, merging this beauty in :)

PR #207, opened 2017-08-06
fix(comment): allow new lines in comments on Android

### LINKED ISSUE #206: Cannot add new lines when writing issue comments
Feels weird that pressing return immediately posts the issue comment and there seems to be no way to insert new lines 🤔

lex111:
Fixes #206.

**Need to check on iOS**. There I understand, there was no such problem?

### DIFF
--- a/src/components/comment-input.component.js
+++ b/src/components/comment-input.component.js
@@ -7,6 +7,7 @@ import {
   TouchableOpacity,
   Text,
   TextInput,
+  Platform,
 } from 'react-native';
 import { Icon } from 'react-native-elements';
 
@@ -59,7 +60,7 @@ export class CommentInput extends Component {
     userHasPushPermission: boolean,
     issueLocked: boolean,
     language: string,
-    onSubmitEditing: Function,
+    onSubmit: Function,
   };
 
   state: {
@@ -76,8 +77,14 @@ export class CommentInput extends Component {
     };
   }
 
-  handleSubmit = (body: string): void => {
-    this.props.onSubmitEditing(body);
+  handleSubmitEditing = (body: string): void => {
+    if (Platform.OS === 'android') {
+      this.setState({ text: `${body}\n` });
+    }
+  };
+
+  handleSubmit = (): void => {
+    this.props.onSubmit(this.state.text);
     this.setState({ text: '' });
   };
 
@@ -113,12 +120,12 @@ export class CommentInput extends Component {
                   : translate('issue.main.commentInput', language)
               }
               multiline
-              blurOnSubmit
+              blurOnSubmit={false}
               onChangeText={text => this.setState({ text })}
               onContentSizeChange={event =>
                 this.setState({ height: event.nativeEvent.contentSize.height })}
               onSubmitEditing={event =>
-                this.handleSubmit(event.nativeEvent.text)}
+                this.handleSubmitEditing(event.nativeEvent.text)}
               placeholderTextColor={colors.grey}
               style={[
                 styles.textInput,
--- a/src/issue/screens/issue.screen.js
+++ b/src/issue/screens/issue.screen.js
@@ -301,7 +301,7 @@ class Issue extends Component {
               }
               issueLocked={issue.locked}
               language={language}
-              onSubmitEditing={this.postComment}
+              onSubmit={this.postComment}
             />
           </KeyboardAvoidingView>}
       </ViewContainer>


housseindjirdeh:
Nice thank you! Will test this in a bit - will try and solve the issue happening on iOS myself as well 💃

lex111:
@housseindjirdeh what's new? Did you manage to test?

src/components/comment-input.component.js
```diff
@@ -76,8 +77,14 @@ export class CommentInput extends Component {
     };
   }
 
-  handleSubmit = (body: string): void => {
-    this.props.onSubmitEditing(body);
+  handleSubmitEditing = (body: string): void => {
+    if (Platform.OS === 'android') {
```
lex111: @housseindjirdeh If there is this problem on iOS, can this condition be removed? It is necessary to check on iOS with these changes, and then you can remove this condition, and check without it.

housseindjirdeh:
My apologies, my week has been swamped and I haven't had the chance to test it :( If not today then hopefully tomorrow and I'll let you know. The thing with iOS is that it's just the return key that's causing to post and I feel like there should be a prop or something similar to do newline instead. 🤔

lex111:
@housseindjirdeh so by pressing the return key, a new line should be inserted? I thought the matter in the `onSubmitEditing` handler, because it was sending a message, and now it happens in the `onSubmit` handler. Perhaps this will help for iOS?

src/issue/screens/issue.screen.js
```diff
@@ -301,7 +301,7 @@ class Issue extends Component {
               }
               issueLocked={issue.locked}
               language={language}
-              onSubmitEditing={this.postComment}
```
lex111: Now you can only set the `onSubmit` handler.

src/components/comment-input.component.js
```diff
@@ -113,12 +120,12 @@ export class CommentInput extends Component {
                   : translate('issue.main.commentInput', language)
               }
               multiline
-              blurOnSubmit
+              blurOnSubmit={false}
               onChangeText={text => this.setState({ text })}
               onContentSizeChange={event =>
                 this.setState({ height: event.nativeEvent.contentSize.height })}
               onSubmitEditing={event =>
```
Antoine38660: According to the [doc](https://facebook.github.io/react-native/releases/0.47/docs/textinput.html#onsubmitediting) the `onSubmitEditing` is useless

lex111: Yes, but even if it is removed (or `false`), the problem remains. As I wrote, this is due to non-standard keyboards (on Android at least). Therefore, need to add an new line. I do not know how things are on iOS? Maybe there is enough to remove `blurOnSubmit`.

Antoine38660: I just tried to post a comment with new line from the iOS simulator and nothing was posted... simulator screen shot 12 aout 2017 a 21 28 22

lex111: Rather, it's because of my other changes, I changed the handlers? It would be cool to solve this on iOS, because everything is good on Android.

Antoine38660: Ok, so let me try to fix this one for iOS :)

lex111: It will break afterwards, when this PR merge?

src/components/comment-input.component.js
Antoine38660: You can remove the condition, np with iOS

src/components/comment-input.component.js
Antoine38660: After some tests, in the logger, I have `{type: "POST_ISSUE_COMMENT_SUCCESS", payload: {message: "Not Found", documentation_url: "https://developer.github.com/v3"}}`.

src/components/comment-input.component.js
lex111: Then new lines will be added to iOS, did you check?

Antoine38660: Yes it tested without the condition, new lines are added but it still "a bug" when I post the comment (nothing is posted)

src/components/comment-input.component.js
lex111: It seems I figured out what I did wrong: TexInput does not have the `onSubmit` handler, and I just changed `onSubmitEditing` to `onSubmit`, but it's strange, then it should not work at all, in the sense that the sending should not happen at all, but Android works.

I'll check it again later.

Antoine38660: But the `onSubmitEditing` handler is no more required if we have `blurOnSubmit={false}`

lex111: What kind of handler is needed to send?

Antoine38660: Idk, can you post a comment from Android (with these changes)?

lex111: Yes, it's sent ... and wait .. After all, we have a comment sent only by [the "Post" button](https://github.com/lex111/git-point/blob/d1668b4e0cd1e003554991efaa0109cd30caa0c4/src/components/comment-input.component.js#L146), just there is a send handler.

Antoine38660: I think it's sufficient :)

lex111: I do not understand why iOS does not work? Do I need to remove the condition?

Antoine38660:
This is a test sent from iOS

With new lines.

lex111:
@Antoine38660 It is perfectly! It turns out we can merge?

PR #209, opened 2017-08-06
Add support generating the release app for Android

lex111:
Based on https://facebook.github.io/react-native/docs/signed-apk-android.html

P.S. It was cool to have a similar for iOS, but I do not have Mac, so it's done so far for Android.


package.json
```diff
@@ -9,9 +9,12 @@
     "start:android:logger": "LOGGER_ENABLED=true concurrently -r 'react-native start --reset-cache' 'react-native run-android'",
     "start:ios": "concurrently -r 'react-native start' 'react-native run-ios'",
     "start:ios:logger": "LOGGER_ENABLED=true concurrently -r 'react-native start --reset-cache' 'react-native run-ios'",
+    "build:android:release": "cd android && ./gradlew assembleRelease",
     "link": "react-native link",
     "ios": "react-native run-ios",
     "android": "react-native run-android",
+    "android:release": "react-native run-android --variant=release",
+    "android:signkey": "cd android/app/ && keytool -genkey -v -keystore my-release-key.keystore -alias my-key-alias -keyalg RSA -keysize 2048 -validity 10000 && cd -",
```
Antoine38660: Maybe you should use an env var for the `my-release-key.keystore` keystore file 😃

lex111: @Antoine38660 did not quite understand how? Can you please explain?

Antoine38660: I mean, change the `"android:signkey"` script by something like `"cd android/app/ && keytool -genkey -v -keystore RELEASE_KEYSTORE_PATH -alias ..."` where `RELEASE_KEYSTORE_PATH` is a local environment variable that points to the keystore file.

But I checked the `.gitignore` file and [every *.keystore are ignored](https://github.com/gitpoint/git-point/blob/master/.gitignore#L42) so @housseindjirdeh can put it here without problem. Actually I think it is already in the repo in the `android/keystore` directory :)

lex111: Well, I do not know, it's recommended to generate it in the `app/android` folder, I can do it, but I do not see it necessary, because all the same, you need to manually specify the key in the `android/gradle.properties` file, including the `MYAPP_RELEASE_STORE_FILE` path, i.e. you can specify it.

housseindjirdeh: Yep so I generated the keystore file and it's currently living on my machine. It is being ignored in `.gitignore` and that's a good thing since we shouldn't be committing it in

If I'm not mistaken , `android:signkey` is for _creating_ a new signing key correct? Not sure how or why we should pass in the keystore as a variable here

android/app/build.gradle
```diff
@@ -105,6 +105,16 @@ android {
             abiFilters "armeabi-v7a", "x86"
         }
     }
+    signingConfigs {
+        release {
+            if (project.hasProperty('MYAPP_RELEASE_STORE_FILE')) {
+                storeFile file(MYAPP_RELEASE_STORE_FILE)
+                storePassword MYAPP_RELEASE_STORE_PASSWORD
```
housseindjirdeh: I actually used [this](https://pilloxa.gitlab.io/posts/safer-passwords-in-gradle/) approach when setting up a password for release. Not sure how we can have that flow integrated into this PR or even if we should as it's  quite user specific :thinking:

package.json
lex111: I'll explain, this can be useful if you need to create a release version of the application. Yes, you need to create a signature key, this command does this.

According to [the manual](https://facebook.github.io/react-native/docs/signed-apk-android.html), the key data must be placed in the `android/gradle.properties`.

@housseindjirdeh I really do not know how you release the new version on Google Play, this change does not hurt?

android/app/build.gradle
lex111: @housseindjirdeh that's the point, or something I do not understand. This is necessary to create its own release version (because for example it is not always enough for Android simulator, you need to check on a real device).

package.json
housseindjirdeh: Nope I probably wouldn't need to run the command since I have a key but it simplifies the process. Was curious about what @Antoine38660 meant by adding an env variable but it should be good.

android/app/build.gradle
housseindjirdeh: Sorry you may have missed the link (click [here](https://pilloxa.gitlab.io/posts/safer-passwords-in-gradle/)). So I don't store my password inside the gradle file and use keychain access instead. _However_, there's no real way to include that logic into the placeholder I think so I'm okay with having this as a placeholder to be honest.

android/gradle.properties
```diff
@@ -1,20 +0,0 @@
...
-# The setting is particularly useful for tweaking memory settings.
-# Default value: -Xmx10248m -XX:MaxPermSize=256m
-# org.gradle.jvmargs=-Xmx2048m -XX:MaxPermSize=512m -XX:+HeapDumpOnOutOfMemoryError -Dfile.encoding=UTF-8
-
-# When configured, Gradle will run in incubating parallel mode.
-# This option should only be used with decoupled projects. More details, visit
-# http://www.gradle.org/docs/current/userguide/multi_project_builds.html#sec:decoupled_projects
-# org.gradle.parallel=true
-
-android.useDeprecatedNdk=true
```
housseindjirdeh: Any reason why we're removing this file? Isn't this file where we need to add the following details:

MYAPP_RELEASE_STORE_FILE=my-release-key.keystore
MYAPP_RELEASE_KEY_ALIAS=my-key-alias
MYAPP_RELEASE_STORE_PASSWORD=*****
MYAPP_RELEASE_KEY_PASSWORD=*****

lex111: Yes, this is it, so it must be ignored.

lex111:
@housseindjirdeh could we merge?

machour:
@housseindjirdeh could you give this a fresher look?

PR #210, opened 2017-08-06
Show message if no contributors found

lex111:
Preview:
image

### DIFF
--- a/src/components/members.component.js
+++ b/src/components/members.component.js
@@ -1,5 +1,4 @@
 import React from 'react';
-
 import {
   StyleSheet,
   View,
@@ -14,6 +13,7 @@ import { colors, fonts } from 'config';
 type Props = {
   title: string,
   members: Array,
+  noMembersMessage: string,
   containerStyle: Object,
   smallTitle: string,
   navigation: Object,
@@ -46,11 +46,16 @@ const styles = StyleSheet.create({
     ...fonts.fontPrimarySemiBold,
     marginBottom: 10,
   },
+  noMembersStyle: {
+    color: colors.black,
+    ...fonts.fontPrimary,
+  },
 });
 
 export const MembersList = ({
   title,
   members,
+  noMembersMessage,
   containerStyle,
   smallTitle,
   navigation,
@@ -59,6 +64,12 @@ export const MembersList = ({
     <Text style={smallTitle ? styles.sectionTitleSmall : styles.sectionTitle}>
       {title}
     </Text>
+
+    {noMembersMessage &&
+      !members.length &&
+      <Text style={styles.noMembersStyle}>
+        {noMembersMessage}
+      </Text>}
 
     <FlatList
       data={members}
--- a/src/components/members.component.js
+++ b/src/components/members.component.js
@@ -7,6 +7,7 @@ import {
   TouchableHighlight,
   Image,
 } from 'react-native';
+import { List, ListItem } from 'react-native-elements';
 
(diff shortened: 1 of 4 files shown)


andrewda:
Out of curiosity, does this happen only on empty repositories?

lex111:
@andrewda probably yes, eg here is [my test repository](https://github.com/lex111/test).
But all the same, it seems to me necessary to process such situation.

andrewda:
Yep! Looks good to me!

lex111:
@housseindjirdeh I did this initially, but it was not very pretty, I'll show you a screenshot later, so you can see it. This is because the list of contributors is not a list (SectionList).

lex111:
@housseindjirdeh here is how the message will look if it is pasted using the `ListItem`, it looks like this because of the padding of the container. But even if this is fixed, it's still not very suitable for this block, because this is not a list.

image

Instead, I would add the bottom line to the title, like so:

image

housseindjirdeh:
Ah that's a very fair point @lex111 - how about we we show the `ListItem` with the message within if there are no contributors but leave it the way it is (without the `ListItem`) if members are present?

lex111:
@housseindjirdeh correctly, did I understand that we are doing the same as on the first screenshot, despite the fact that it does not look so good?

housseindjirdeh:
Ah you're right it's a little more complicated then that - just dived in a bit. Something like this should work: 

```
import React from 'react';

import {
  StyleSheet,
  View,
  Text,
  FlatList,
  TouchableHighlight,
  Image,
} from 'react-native';

import { List, ListItem} from 'react-native-elements';

type Props = {
  title: string,
  members: Array,
  smallTitle: string,
  navigation: Object
};

const styles = StyleSheet.create({
  container: {
    marginTop: 30,
  },
  avatarContainer: {
    //..
  },
  avatar: {
    //.. 
  },
  list: {
    marginTop: 0,
  },
  sectionTitle: {
    color: colors.black,
    ...fonts.fontPrimaryBold,
    marginBottom: 10,
    paddingLeft: 15,
  },
  sectionTitleSmall: {
    color: colors.primaryDark,
    ...fonts.fontPrimarySemiBold,
    marginBottom: 10,
    paddingLeft: 15,
  },
  listTitle: {
    color: colors.black,
    ...fonts.fontPrimary,
  },
});

export const MembersList = ({
  title,
  members,
  smallTitle,
  navigation,
}: Props) =>
  <View style={[styles.container]}>
    <Text style={smallTitle ? styles.sectionTitleSmall : styles.sectionTitle}>
      {title}
    </Text>

    {!members.length &&
    <List containerStyle={styles.list}>
        <ListItem
          title={'No contributors found'}
          titleStyle={styles.listTitle}
          hideChevron
        />
      </List>
    }

    {members.length &&
      <View style={styles.listPadding}>
        <FlatList 
          // ...
        />
      </View>
    }
  </View>;
```

housseindjirdeh:
1. Don't think `containerStyle` prop is being used (I know it's outside the scope of this PR but please feel free to just remove it)
2. Needs a little tinkering with the styles but I think what I have above _roughly_ should make things work so it looks like this without any members:

image

lex111:
@housseindjirdeh thanks for the support! Added only `paddingLeft` for FlatList, now it seems to be good.

lex111:
@housseindjirdeh resolved the conflict via GitHub :heavy_check_mark:

housseindjirdeh:
Beautiful thank you matey 🎉

PR #211, opened 2017-08-06
Navigate to auth profile screen for the current user

lex111:
Now the screen auth profile can only be when navigating from the tabbar, although it is displayed for the current user from other places: if the user is on the organization list, if it is mentioned in the issue description or in its comments. Of all these places user should be navigated **not** to the normal profile screen, but to the auth profile.

From this follows such a problem that I can _follow myself_ (:rofl:), or again, organizations will not all be displayed.

### DIFF
--- a/src/components/members.component.js
+++ b/src/components/members.component.js
@@ -1,5 +1,5 @@
 import React from 'react';
-
+import { connect } from 'react-redux';
 import {
   StyleSheet,
   View,
@@ -11,12 +11,17 @@ import {
 
 import { colors, fonts } from 'config';
 
+const mapStateToProps = state => ({
+  authUser: state.auth.user,
+});
+
 type Props = {
   title: string,
   members: Array,
   containerStyle: Object,
   smallTitle: string,
   navigation: Object,
+  authUser: Object,
 };
 
 const styles = StyleSheet.create({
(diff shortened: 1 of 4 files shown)


src/components/members.component.js
```diff
@@ -65,10 +71,14 @@ export const MembersList = ({
       showsHorizontalScrollIndicator={false}
       renderItem={({ item }) =>
         <TouchableHighlight
-          onPress={() =>
-            navigation.navigate('Profile', {
-              user: item,
-            })}
+          onPress={() => {
+            navigation.navigate(
+              authUser.login === item.login ? 'AuthProfile' : 'Profile',
```
RolfKoenders: I noticed that you use the short if syntax on 1 line here and on line 255 for example of the 'comment-list-item.component.js' file you spread it over multiple lines. Would be nice if we have a consistency.

lex111: @RolfKoenders yep, it's good that you noticed this! I thought about this, I also did not want to duplicate the code, but where to put it? In `utils`? And how to call the function correctly, something like `getProfileScreen`?

housseindjirdeh: I'm okay with leaving the logic here, I don't think this is a helper we would essentially need in `utils`.

Regarding @RolfKoenders's mention, you're right but this is something `prettier` and our precommit hook should take care of actually. @lex111 if I'm not mistaken this is going through the hook and getting prettified correct?

housseindjirdeh:
Good stuff thank you @lex111 :)

src/components/members.component.js
lex111: @housseindjirdeh yep, my hooks are working, I'm watching this.

PR #216, opened 2017-08-06
Not show title "Info" if there is not any information to display

lex111:
Prevent render EntityInfo component if there is nothing to show (eg [github-beta](https://github.com/github-beta)).

Before | After
| - | - |
image | image

### DIFF
--- a/src/components/entity-info.component.js
+++ b/src/components/entity-info.component.js
@@ -68,6 +68,12 @@ const navigateToCompany = (company, orgs, navigation) => {
 };
 
 export const EntityInfo = ({ entity, orgs, language, navigation }: Props) => {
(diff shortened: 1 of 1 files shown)


RolfKoenders:
I'm wondering if we should do the same as you did in issue #210. I think its better to show the user that there is no info and show that as a message instead of nothing. They may think its a bug?

lex111:
@RolfKoenders No, I do not think so, just now, users can think it's a mistake.
Now the "Description" title is actually hidden, if there is no description for the repository. Therefore, by analogy with this, I propose to do with the title "Info".

housseindjirdeh:
Thanks for this @lex111 :)

You make a good point @RolfKoenders but I'm good either way to be honest. Happy to further discuss if we feel like we should be having a message here as well.

PR #227, opened 2017-08-11
feat(repo): add label to show watch status

### LINKED ISSUE #185: Show if a user is watching a repo or not
After merging #184 you should be able to tell if you're watching a repo or not, it will be good to add a visual indicator for it.

SammyIsra:
Resolves #185 
Shows a label to indicate that the logged in user is currently watching/subscribed to a repository. It looks like this: 

Watched label

I think I made a mistake when getting updated with master, which is why so many commits show up. However the changes are accurate to what I changed and is necessary for the issue.


lex111:
Can I add an icon instead of an text, for example [this](https://octicons.github.com/icon/eye/) next to the name of the repository? For greater clarity, you can display it in green, eg. Now it does not seem to me that it looks good.

P.S. Better to do `git rebase`, that was not such a large number of commits.

src/components/repository-profile.component.js
```diff
@@ -141,6 +143,10 @@ export const RepositoryProfile = ({
         {emojifyText(repository.description) || ' '}
       </Text>
 
+      <Text style={styles.languageInfoTitle}>
+        {subscribed ? '(watching)' : undefined}
```
lex111: It needs to be moved to the local file (`src/locale/languages/en.js`) and output via [translate](https://github.com/gitpoint/git-point/blob/master/src/utils/method-helpers.js#L33) helper.

And yet, can it be better to return `null` instead of `undefined`?

SammyIsra: I agree regarding the languages, I will update that.

Since the style did not need to change much, I decided to use a style that was already created. In any case, I can copy the contents into its own and call it `watchingLabel`.

How would `null` be better than `undefined` in this situation? Both end up rendering nothing to the DOM (And I think the same goes for `false`)

SammyIsra:
@lex111 The reason I went for the text instead of the icon is that I want to avoid a "what does this mean?" situation. I think there was a talk about this and the Followers number been green when you are following someone already. I've never had to do git rebase before, I will try to do that to avoid this situation again in the future. Latest commit did not work for me (the hooks were removed?) I saw on gitter that it was recommended to remove `./git/hooks` and then run `yarn`, so that's what I did. Apparently that did not work out well, will take a look tonight at that. I'll be more careful witht he commit naming int he future!

@andrewda The way to show if a user has starred a repo right now is making the text green, same as when youa re following a user. Personally, I think this is not clear and it leads to people trying to find out what the green means. If you disagree I can of course add the number of subscribers and make it green also when a user is watching the repo

andrewda:
@SammyIsra I agree that it's confusing, but keeping it the way it is maintains consistency. Adding this might make the green star count even _more_ confusing because we're using two different status indications for the repo.

I'd say just copy the star component, change it to watchers and put it between "Stars" and "Forks". Or you could also change the way the star status is displayed. Either way I think they need to be done together to avoid more confusion.

lex111:
> I'd say just copy the star component, change it to watchers and put it between "Stars" and "Forks". Or you could also change the way the star status is displayed. Either way I think they need to be done together to avoid more confusion.

In my opinion the best option :+1:

SammyIsra:
Perfect, I'll submit the changes to this branch tonight! Thank you

housseindjirdeh:
Thanks so much for this @SammyIsra 

Agreed with @lex111 and @andrewda's suggestion: I think that would be absolutely perfect. Appreciate it and let us know once those changes are in my friend :)

SammyIsra:
Do you all think this looks ok?

When a user is not watching:
Not watching

When a user is watching:
Watching

lex111:
@SammyIsra great! About localization: we have an open PR #193, can better make these fixes in it?

housseindjirdeh:
This looks awesome @SammyIsra :tada: 

Are you still adding some changes to this PR? WIP label is still attached so just wanted to make sure. It looks good to me, so if not could you fix the conflicting files when you get the chance and I'll merge this in :)

SammyIsra:
@housseindjirdeh As soon as I resolve the conflicting files the PR will be ready for merge!

src/repository/screens/repository.screen.js
```diff
@@ -252,6 +255,7 @@ class Repository extends Component {
                 }
                 loading={isPendingRepository}
                 navigation={navigation}
+                subscribed={isPendingSubscribe ? false : subscribed}
```
andrewda: Just a tiny thing: Could you move this up right below `starred` and follow the same format as `starred`?

i.e.

```
subscribed={
  isPendingRepository || isPendingSubscribe ? false : subscribed
}
```

SammyIsra: Addressed! Also did the same fix for another line of code.

Issue #29, opened 2017-07-12
Can't run the project


alejandronanez:
I follow the steps under CONTRIBUTING.md but couldn't run the project. What I'm seeing is this:
 ```
No bundle URL present.
Make sure you're running a package server or have included a .jsbundle file in your application bundle
```
Any idea what could be wrong?

Thanks!

**Comment by housseindjirdeh** [comment]
Aaaah that's most likely due to me removing a localhost exception in `Info.plist` since I had to in order to deploy the app. Included it back in (https://github.com/gitpoint/git-point/commit/f471f044ecf60209da62be1a84d96e9979bcf93a) - I'll remember to remove it locally before I deploy the next update.

**Comment by machour** [comment]
Facing the exact same error message today.

Reproduce case:
```
git clone git@github.com:gitpoint/git-point.git git-point-orig
cd git-point-orig
yarn && yarn run link && yarn start:ios
``` 
None of the provided troubleshooting options provided in `CONTRIBUTING.md` did the trick for me.

The localhost exception **is** present in `Info.plist`

Environment infos:
```
🍺  ~/git-point-orig (master)*$ node --version
v8.1.3
🍺  ~/git-point-orig (master)*$ npm --version
5.0.3
🍺  ~/git-point-orig (master)*$ yarn --version
0.27.5
🍺  ~/git-point-orig (master)*$ react-native --version
react-native-cli: 2.0.1
react-native: 0.46.2
The app is loaded in a simulator with the following specs: iPhone 6 - iOS 10.3
Running all this on a MacBook Pro (15-inch, 2016) with OsX 10.12.6
```

Let me know if I can provide more details

**Comment by housseindjirdeh** [comment]
@machour are you still having trouble with iOS and have only been able to test on Android? Please let me know and we'll reopen this. Apologies if we didn't get to the bottom of this in our gitter chat but I can't recall if we have :thinking:

**Comment by machour** [comment]
Just did a fresh checkout and hit the same issue again: "No bundle URL is present".
This is the case when running "yarn start:ios" or "yarn start -- --reset-cache" and then yarn run ios in a another terminal window

```
🍺  ~/git-point (master)*$ yarn --version
0.27.5
🍺  ~/git-point (master)*$ npm --version
5.3.0
🍺  ~/git-point (master)*$ node --version
v8.4.0
```

machour:
I did some further digging, starting with a fresh repo and XCode, here's what I've noticed:

[...]

PR #229, opened 2017-08-11
fix(share): stop pre-filling email subject

### LINKED ISSUE #228: Don't pre-fill the email subject
Currently, if I click on the email action in a user profile, my email application opens with a pre-filled subject "Hi !".

I don't think this is a good behavior:
- I might forget to edit the subject, and my mailer won't warn me
- "Hi !" is really not a good email subject, and we shouldn't encourage users to use it

I know, I'm picky 🤓

machour:
Tested on Android only as I didn't succeed in running it on iOS (See #29)

### DIFF
--- a/src/components/entity-info.component.js
+++ b/src/components/entity-info.component.js
@@ -124,7 +124,7 @@ export const EntityInfo = ({ entity, orgs, language, navigation }: Props) => {
           subtitle={entity.email}
           subtitleStyle={styles.listSubTitle}
           onPress={() =>
-            Communications.email([entity.email], null, null, 'Hi!', '')}
+            Communications.email([entity.email], null, null, null, '')}
           underlayColor={colors.greyLight}
         />}
 
--- a/src/components/entity-info.component.js
+++ b/src/components/entity-info.component.js
@@ -124,7 +124,7 @@ export const EntityInfo = ({ entity, orgs, language, navigation }: Props) => {
           subtitle={entity.email}
           subtitleStyle={styles.listSubTitle}
           onPress={() =>
-            Communications.email([entity.email], null, null, null, '')}
+            Communications.email([entity.email], null, null, null, null)}
           underlayColor={colors.greyLight}
         />}
 


src/components/entity-info.component.js
```diff
@@ -124,7 +124,7 @@ export const EntityInfo = ({ entity, orgs, language, navigation }: Props) => {
           subtitle={entity.email}
           subtitleStyle={styles.listSubTitle}
           onPress={() =>
-            Communications.email([entity.email], null, null, 'Hi!', '')}
+            Communications.email([entity.email], null, null, null, '')}
```
andrewda: While not technically under the scope of this PR, could you change the last value to `null`, too, instead of an empty string? They both work, but the react-native-communications documentation says `null` is the way to go.

machour: There you go: https://github.com/gitpoint/git-point/pull/229/commits/14d719fee6894eb98d3a2ec45e40cd046e1a477a
Thanks for the quick review :)

lex111:
@machour just wondering how this fix solves #29?

lex111:
@machour strange, it seems that not work git hooks for you?

machour:
@lex111 indeed, it does not: 

```
git commit -m "Pass null for body as well" src/components/entity-info.component.js
husky > npm run -s precommit (node v8.1.3)

 ✔ Running tasks for *.js
husky > npm run -s commitmsg (node v8.1.3)

⧗   input: Pass null for body as well
✖   message may not be empty [subject-empty]
✖   type may not be empty [type-empty]
✖   found 2 problems, 0 warnings
```

Provided that the change is simple and tested on an actual device, I allowed myself a `--no-verify`
I'm on gitter if you want to investigate this

andrewda:
@machour The hook that's failing there is for the commit message. It should be something like `fix(share): stop pre-filling email subject`

machour:
Duly noted for next commit messages. Thanks a lot !

housseindjirdeh:
Thanks a million for this @machour <3

Don't forget to add yourself as a contributor with [`yarn constributors:add`](https://github.com/gitpoint/git-point/blob/master/CONTRIBUTING.md#setup)!!!!

PR #237, opened 2017-08-12
fix(events): Properly handle the User made repo public event

### LINKED ISSUE #223: Handle "[user] made [repository] public"
Currently the event is wrongly shown as "[user] [missing {(payload)} value] issue [repository]"

machour:
Closes #223


lex111:
@machour please resolve the conflicts.

src/auth/screens/events.screen.js
```diff
@@ -149,7 +149,7 @@ class Events extends Component {
       case 'MemberEvent':
         return `${userEvent.payload.action}`;
       case 'PublicEvent':
-        return translate('auth.events.issuesEvent', language);
+        return translate('auth.events.publicEvent', language);
```
andrewda: I think this will turn out to look like `User made public {{repo}}`. If at all possible, I think it would be best to word it `User made {{repo}} public`.

screen shot 2017-08-12 at 9 51 05 am

machour: I tried doing this at first, but I don't think it's possible with the current translation implementation. 

From my (new to react-native world) understanding, the repository link is prepended to the sentence around here: https://github.com/gitpoint/git-point/blob/master/src/auth/screens/events.screen.js#L492 and the payload is empty in this action.

Splitting the sentence in two parts and using getConnector() to inject "public" seemed quiet hackish to me, so I settled on a poorer phrasing 😊 

I'd appreciate it if one of you guys could have a look and fix my PR the proper way so that I could learn from it.

lex111: It would be better to fix this in #193, because this refers to a localization error, although once there is a separate PR, can this be fixed here.

lex111: @machour I did this (`{{user}} made {{repo}} public`), but in PR #193, however you can copy the [changes](https://github.com/gitpoint/git-point/pull/193/commits/1c2aa767e6cd0cd2168620f780d55135367f3301) (at the same time, you can come up with a better solution than mine)  and move it to this PR.

machour: Ok, no need to keep my PR then. We will come up with a cleaner solution later.
Nice work !

PR #239, opened 2017-08-12
refactor(*): change repository actions order

### LINKED ISSUE #224: Change ordering of repository actions
At the moment, the order goes:
- `Star`
- `Fork`
- `Share`
- `Watch`

`Share` is the only one that's different from the others in that it doesn't take any action on the repository itself. It would be good to move `Share` to the bottom, beneath `Watch` to organize them correctly.

Probably best suited for another PR, but maybe take away the emojis from the Repository Actions menu. Until we get a way to use actual icons, these look weird IMO.

unnamed

wlisrausr:
Share menu is the only one that's different from the others
because it doesn't take any action on the repository itself. 
It would be better to move it to the bottom (beneath Watch menu) to organize them correctly.

Closes #224


wlisrausr:
You're welcome @Antoine38660. It is actually my really first contribution to the open source project. Feels great and would love to contribute some more to this git-point amazing open source 🎉

PR #243, opened 2017-08-12
fix(new-issue): various submit new issue improvement

### LINKED ISSUE #215: Issue is created twice when having slow internet
Hey, as you can see at #213 and #214 the issue has been created twice because I'm in Italy and have a slow internet connection.

### LINKED ISSUE #236: Don't allow back after submitting issue
Hey, it should not be possible to go back to the "new issue" page with the contents after I submitted it.

Antoine38660:
- [x] Fix #215 
- [x] Fix #236

Also give ability to add new line in the issue body

### DIFF
--- a/routes.js
+++ b/routes.js
@@ -134,11 +134,15 @@ const sharedRoutes = {
       const { issue, issueURL, isPR, language } = navigation.state.params;
       const number = issue ? issue.number : issueURL.match(issueNumberRegex)[1];
       const langKey = isPR ? 'pullRequest' : 'issue';
-      const langTitle = translate(`issue.main.screenTitles.${langKey}`, language);
+      const langTitle = translate(
+        `issue.main.screenTitles.${langKey}`,
+        language
+      );
 
       return {
         title: `${langTitle} #${number}`,
         headerLeft: navigation.state.params.headerLeft,
+        gesturesEnabled: !(navigation.state.params.gesturesEnabled === false),
       };
     },
   },
--- a/routes.js
+++ b/routes.js
@@ -138,6 +138,7 @@ const sharedRoutes = {
 
       return {
         title: `${langTitle} #${number}`,
+        headerLeft: navigation.state.params.headerLeft,
       };
     },
   },
--- a/src/issue/screens/new-issue.screen.js
+++ b/src/issue/screens/new-issue.screen.js
@@ -3,7 +3,7 @@ import { connect } from 'react-redux';
 import { ScrollView, StyleSheet, TextInput, View, Alert } from 'react-native';
 import { ListItem } from 'react-native-elements';
 
-import { ViewContainer, SectionList } from 'components';
+import { ViewContainer, SectionList, LoadingModal } from 'components';
 import { translate } from 'utils';
 import { colors, fonts, normalize } from 'config';
 import { submitNewIssue } from '../issue.action';
@@ -33,6 +33,7 @@ const styles = StyleSheet.create({
 const mapStateToProps = state => ({
   language: state.auth.language,
   repository: state.repository.repository,
+  isPendingSubmitting: state.issue.isPendingSubmitting,
 });
 
 const mapDispatchToProps = dispatch => ({
@@ -46,6 +47,7 @@ class NewIssue extends Component {
     language: string,
     repository: Object,
     navigation: Object,
+    isPendingSubmitting: boolean,
   };
 
   state: {
@@ -97,11 +99,12 @@ class NewIssue extends Component {
   };
 
   render() {
-    const { language, repository } = this.props;
+    const { language, repository, isPendingSubmitting } = this.props;
     const { issueTitle, issueComment } = this.state;
 
     return (
       <ViewContainer>
+        {isPendingSubmitting && <LoadingModal />}
         <ScrollView>
           {repository.full_name &&
             <ListItem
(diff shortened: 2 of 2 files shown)


lex111:
> Also give ability to add new line in the issue body

Look #207 please, I did there for Android.

routes.js
```diff
@@ -135,6 +135,7 @@ const sharedRoutes = {
 
       return {
         title: isPR ? `Pull Request #${number}` : `Issue #${number}`,
+        headerLeft: navigation.state.params.headerLeft,
```
housseindjirdeh: Although this would essentially solve hiding the header on the left, have you tried swiping the screen from the left side? That might still allow the user to navigate back which we wouldn't want.

Also see comment 2 about possibly simplifying this flow 🕺

src/issue/screens/new-issue.screen.js
```diff
@@ -90,17 +92,19 @@ class NewIssue extends Component {
...
   };
 
   render() {
-    const { language, repository } = this.props;
+    const { language, repository, isPendingSubmitting } = this.props;
     const { issueTitle, issueComment } = this.state;
 
     return (
       <ViewContainer>
+        {isPendingSubmitting && <LoadingModal />}
```
housseindjirdeh: <3

src/issue/screens/new-issue.screen.js
```diff
@@ -90,17 +92,19 @@ class NewIssue extends Component {
       ).then(issue => {
         navigation.navigate('Issue', {
           issue,
+          headerLeft: false,
```
housseindjirdeh: Comment #2 

So instead of passing a prop to hide the header, it might just be simpler to use `navigation.reset` to [`reset`](https://reactnavigation.org/docs/navigators/navigation-actions#Reset) the current navigation stack instead of pushing another route on top. That way there shouldn't be a navigate back functionality at all and we wouldn't have to workaround trying to pass a manual prop to hide it.

Let me know what you think @Antoine38660 :)

src/issue/screens/new-issue.screen.js
```diff
@@ -140,7 +144,6 @@ class NewIssue extends Component {
             <TextInput
               underlineColorAndroid={'transparent'}
               placeholder={translate('issue.newIssue.writeAComment', language)}
-              blurOnSubmit
```
housseindjirdeh: 🎉

routes.js
Antoine38660: You're right, I haven't tried to swipe left, and it still allow the user to navigate back... 

(-> Comment 2)

src/issue/screens/new-issue.screen.js
Antoine38660: I tried to use the [reset function](https://github.com/gitpoint/git-point/blob/master/src/utils/method-helpers.js#L20) (defined in `utils/method-helpers.js`) but I had an error, something like:
> 'Issue' key is not defined, you should use Login, Main ...

I'll try now to use directly `navigation.reset` 😃 

Thanks for reviewed this!

Antoine38660: With `navigation.reset`, we can't navigate back ✅  
**BUT** the bottom menu item selected when we create the issue can no longer be used 😞
It redirect back to the created issue instead of its normal behavior...

housseindjirdeh: Ah that error message is most likely because its only on the top Stack (see [here](https://github.com/gitpoint/git-point/blob/master/routes.js#L339-L366)) hence why it can't recognize the `Issue` screen :thinking: 

Glad to hear that just using `.reset` worked (somewhat) :P. The bottom menu item? Are you referring to the tab navigator?

Antoine38660: Yes I was referring the tab navigator.

I tested it again today and it works like a charm :
ezgif-4-9d4873f833

😃

Antoine38660: Yes I was referring the tab navigator.

Here's a gif of the behavior:
ezgif-4-915da04836

When I go back to the "User" tab, I spam the "User" tab item but it doesn't redirect me to my profile screen....

housseindjirdeh: AH snap, didn't think of that 🤔 

So double tapping the a single tab navigator will take you back to the `root` of that stack because of [this code here](https://github.com/gitpoint/git-point/blob/master/routes.js#L309-L336). However, using `navigation.reset` in this step will cause that to be the root of the stack which we don't really want.

This is a little tricker than I thought it would be 😂. There's an attribute that we should easily be able to extend to the issue screen to prevent swipe backs (https://github.com/react-community/react-navigation/issues/1063#issuecomment-312842038) but I don't want to always disable it for the screen. 

Some thoughts:
1. Maybe we can use this attribute but connect to navigation params so that it's only false based on the param (which we can set only on this navigate step to `Issue Screen`? That way with this + `headerLeft: null` would not allow the user to go back. Sounds easier said than done 🤔 
2. Maybe we can leverage a `modal` screen here instead? Haven't tried to load a modal with React Navigation yet but maybe it might be worthwhile to try here?

Either way, happy to merge this in if you would rather we deal with this as part of a separate bug ticket 🎉

Antoine38660: I can try to implement the 1) (with `gesturesEnabled: false` + `headerLeft: null`) 😄 

But I don't understand the number 2 with modal :(

Antoine38660:
It works like a charm mate 😃 

I disabled swipe back with `gesturesEnabled: false` and hide the header left btn with `headerLeft: null` on the Issue screen **only** if we just submitted the issue. 👍

PR #250, opened 2017-08-14
fix(repository): Swap watch and share actions

### LINKED ISSUE #249: Share and Watch actions are inverted
This is a bug introduced in #239 which took care of swapping the strings, but not the associated actions.

machour:
Fixes #249

### DIFF
--- a/src/repository/screens/repository.screen.js
+++ b/src/repository/screens/repository.screen.js
@@ -137,14 +137,14 @@ class Repository extends Component {
       forkRepoByDispatch(repository.owner.login, repository.name).then(json => {
         navigation.navigate('Repository', { repository: json });
       });
-    } else if ((index === 2 && showFork) || (index === 1 && !showFork)) {
-      this.shareRepository(repository);
-    } else if (index === 3) {
+    } else if (index === 2) {
       const subscribeMethod = !subscribed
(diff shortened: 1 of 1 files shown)


src/repository/screens/repository.screen.js
```diff
@@ -137,14 +137,14 @@ class Repository extends Component {
...
-    } else if ((index === 2 && showFork) || (index === 1 && !showFork)) {
-      this.shareRepository(repository);
-    } else if (index === 3) {
+    } else if (index === 2) {
       const subscribeMethod = !subscribed
         ? this.props.subscribeToRepo
         : this.props.unSubscribeToRepo;
 
       subscribeMethod(repository.owner.login, repository.name);
+    } else if ((index === 3 && showFork) || (index === 2 && !showFork)) {
```
Antoine38660: The second condition `(index === 2 && !showFork)` will never be called (Because of `if (index === 2)` before)

machour: Whoops 😓  I'll look into that after lunch

machour: Fixed !

andrewda:
If we start to add more to the menu, we should probably figure out a way to make this simpler. For now, though, this is good.

Issue #167, opened 2017-07-25
Use async/await


housseindjirdeh:
I think using async/await will be nice, should make things easier to digest in `api/index.js` 😊  

Shouldn't be too much trouble to refactor everywhere in the app that uses promises to leverage `async/await` instead but will definitely take a bit of time.

caudaganesh:
hello Mr. @housseindjirdeh , i am new to react native and want to try to contribute to this issue, so what i have to do is to convert all promises to async/await in api/index.js is that correct? thanks before

andrewda:
@caudaganesh Welcome! And yep, doing that should make the code a lot easier to read and process. Comment here if you have any difficulties!

housseindjirdeh:
Thank you @caudaganesh, welcome :)

Yep just like @andrewda mentioned it's just some nice sugar over promise syntax. This ticket is pretty much replacing all the promises in that file with async/await but there are also some parent promises in some component files/action files. Just searching for `.then` should show you all the instances in the app but there are just a few that aren't in the `api/index.js` file.

caudaganesh:
@housseindjirdeh @andrewda 
Thank you sirs, I will start working on this :) 
definitely will need some advice but i will try to work on it first. cheers

housseindjirdeh:
I think we can close this thanks to your awesome work @caudaganesh :)

PR #253, opened 2017-08-15
refactor(*): use async/await instead of promises

caudaganesh:
#167 
Breakdown 
- [x] Change all promises to async/await


andrewda:
Looks good. I'll test it out in a bit.

caudaganesh:
@lex111 @andrewda thanks sirs, pls let me know if there is something i need to fix

.all-contributorsrc
```diff
@@ -222,6 +222,15 @@
       ]
     },
     {
+      "login": "caudaganesh",
+      "name": "Ganesh Cauda Salim",
```
housseindjirdeh: :man_dancing:

Issue #159, opened 2017-07-24
Change green number status for following/starring repos


housseindjirdeh:
Currently if you follow a user _or_ star a repo, the followers/stars count becomes green as an indicator.

Probably not the best design decision :P - I think a simpler approach would be having a little text field underneath that shows (`Following` or `Starred`) as @Antoine38660 suggested. Definitely open to other ideas however so suggestions are welcome 🙌

sasongkoadi:
is this issue still open? @housseindjirdeh

housseindjirdeh:
@sasongkoadi yes it is! Please feel free to work on it if you would like :)

sasongkoadi:
oke i handle this :) thank you @housseindjirdeh oh ya if you have any idea about this issue, you can tell me.

sasongkoadi:
is it okay, if i'm using something like toast to solve this issue? we can use toast component from [native base](https://docs.nativebase.io/Components.html#Toast)

reyhansofian:
@housseindjirdeh I've created a PR for this issue. please take a look

sasongkoadi:
![screenshot from 2017-08-16 15-05-58](https://user-images.githubusercontent.com/23546252/29353666-31e63c64-8295-11e7-9fe5-68c17bd6c882.png)

How about this? @housseindjirdeh @lex111 @andrewda 
Can you give me some suggest? :smile: I'll try my best

chinesedfan:
Hi @housseindjirdeh, I know you may be very busy everyday. But can you close those solved issues?

machour:
Closed by #260

PR #260, opened 2017-08-16
feat(*): status for following users/starring repos

reyhansofian:
Issue #159 

- [x] Add `Starred` status below `Stars` on repo screen
- [x] Add `Following` status below `Following` on user screen

# Preview

image
____

image


lex111:
Can make a smaller font? And/or make a clarification like in Twitter, i.e. instead of "following" show "follows you", if he follows you? Similarly do in "Following", but there will be "following you" if you follow him?

sasongkoadi:
Can I give any suggest  for following @lex111 ??

reyhansofian:
@andrewda sure, I'll update my branch and add to watcher as well
@lex111

sasongkoadi:
Can I edit following, and starring style, for more looks good? @reyhansofian

reyhansofian:
oh wait, I'm still working on it @sasongkoadi 
I'll let you know when it's pushed 😄

reyhansofian:
## Updated User Status
screen shot 2017-08-16 at 7 43 55 pm

## Updated Repo Status
screen shot 2017-08-16 at 7 45 15 pm

reyhansofian:
@sasongkoadi please let me know if you're still want to add something to this PR

reyhansofian:
## Added 'Follows You' Status

screen shot 2017-08-16 at 9 47 48 pm

lex111:
@reyhansofian It looks cool indeed! :+1: Although I'm not sure if this fits into the design, I would add an top margin before the statuses. Because of the rounded corners, it looks like buttons, although they are not them, can they be removed? In any case, thank you, what other ideas, how should be these statuses look like?

reyhansofian:
@lex111 ah you're right! it need to some margin on the top. I'll fix that! Regarding the statuses style, what about make it looks like a label? Just like what Twitter did

screen shot 2017-08-16 at 11 13 25 pm

@andrewda will update soon!

andrewda:
@reyhansofian Yea that'd look good. I think you could also try reducing the padding within the label to make it look less like a button.

reyhansofian:
@andrewda OK, will do!

reyhansofian:
@lex111 @andrewda updated! please check 😄 let me know if the statuses border still doesn't make sense and we can remove it

screen shot 2017-08-17 at 12 31 10 am

screen shot 2017-08-17 at 12 31 26 am

lex111:
@reyhansofian not bad! One clarification: will the width of the status increase with the change in the text? Simply in English it looks good, but there will be other languages, and the length will differ and then how will the statuses be displayed, I mean the word breaks is there?

reyhansofian:
@lex111 yeah, it will break into newline. I'm not sure how to do that in RN tho 😞  

screen shot 2017-08-17 at 1 14 56 am

lex111:
Then everything is fine? I just thought that there would be no breaks. Although it was cool if the width of the block expanded, or increase more it, than now, do you think?
image

reyhansofian:
if you're okay with current approach (break into new line instead of expand the container width), then everything is fine 😄 I'll try to make it expanded by tomorrow.

lex111:
I do not know for sure, but what think @andrewda @housseindjirdeh?

andrewda:
Not sure. I think I kinda prefer expanding the width, but either way works.

src/locale/languages/en.js
```diff
@@ -157,6 +157,9 @@ export const en = {
     followingList: {
       title: 'Following',
     },
+    followYou: {
+      title: 'Follows you',
+    },
```
housseindjirdeh: As part of trying to streamline translation updates, could you add a comment to the top of `en.js` with the commit hash and link to the PR? That way we can sort of be able to tell if translations are not the same between multiple different languages.

CC @machour

src/user/user.action.js
```diff
@@ -85,13 +85,36 @@ export const checkFollowStatus = url => {
...
+};
+
 export const getUserInfo = user => {
   return dispatch => {
     return dispatch(getUser(user)).then(() => {
       dispatch(getOrgs(user));
       dispatch(
         checkFollowStatus(`https://api.github.com/user/following/${user}`)
       );
+      dispatch(getFollowers(user));
```
housseindjirdeh: I'm assuming we fire this to get the list of followers _as soon_ as a user navigates to the user profile screen so that we can determine whether we're following them correct? If that's the case, it's also currently being fired in the [`follower-list.screen`](https://github.com/gitpoint/git-point/blob/80495531c8307aabe3b486b53a2833bac9d68a96/src/user/screens/follower-list.screen.js#L29) so we'll need to remove it there to prevent it being called unnecessarily for the second time.

.all-contributorsrc
```diff
@@ -238,6 +238,15 @@
       "contributions": [
         "doc"
       ]
+    },
+    {
+      "login": "reyhansofian",
+      "name": "Reyhan Sofian",
```
housseindjirdeh: ❤️

src/config/colors.js
```diff
@@ -12,6 +12,7 @@ export const colors = {
   greyMidLight: '#f6f8fa',
   greyVeryLight: '#fbfbfb',
   lightGreen: '#00FF00',
+  paleGreen: '#359d68',
```
housseindjirdeh: Curious, do you think `paleGreen` will look better than `green` here? Thinking a lighter shade bolded might give it more pop 🎉 but maybe that's just me 😂

reyhansofian: naming is hard 😂 how about `lightBoldGreen`? 🙈

src/user/user.action.js
reyhansofian: exactly! I'll remove that action on `follower-list.screen`. Thanks for pointing out 👍

src/locale/languages/en.js
reyhansofian: will do

lex111: And what if you add new phrases for translation into all language files? Then it will be clear what needs to be translated in other languages. In fact, we need to do this, because we get the situation that in the English version we have `followYou`, and in that case there will be an error (missing value) in other languages.

reyhansofian: I just realised that we have France translation. I will add the `followYou` key there

andrewda: Can we make it fall back on the English translation if a phrase doesn't exist in a certain language? That would make everything a lot smoother IMO, and make it so we don't have to update every language in small PR's.

machour: @housseindjirdeh no need to do that for the `en.js` file, this is only for translators to keep some kind of internal record.

+1 on @andrewda request, we need a fallback to english and developers shouldn't have to touch translations files. It used to be the case with the previous system.

@reyhansofian if you do touch the french file, here's the translation: "Vous suit".

reyhansofian:
@machour @andrewda added French translation and activate fallback for translation. please check it out

src/locale/languages/fr.js
```diff
@@ -161,6 +161,9 @@ export const fr = {
     followingList: {
       title: 'Following',
     },
+    followYou: {
+      title: 'Vous suit',
```
Antoine38660: 👍

reyhansofian:
thanks guys! 👍

PR #262, opened 2017-08-16
fix(events): Fix wrong wording for commented on pull request

### LINKED ISSUE #261: Wrong wording for "commented on pull request"
Currently, it's displayed as "User created on pull request" : 

wrong wording

it should be "commented on"

machour:
Closes #261

### DIFF
--- a/src/auth/screens/events.screen.js
+++ b/src/auth/screens/events.screen.js
@@ -155,7 +155,7 @@ class Events extends Component {
             'auth.events.pullRequestReviewCommentEvent',
             language,
             {
-              action: translate(`auth.events.actions.${action}`, language),
+              action: translate('auth.events.actions.commented', language),
             }
           );
         } else if (action === 'edited') {


lex111:
Hmm, for sure, here is a similar situation as the [IssueCommentEvent](https://github.com/machour/git-point/blob/061d836bfe3621e84e417bd5c483c0a3e93be5dc/src/auth/screens/events.screen.js#L123)

Do you think it can then refactor the code of this event, as it is done in `IssueCommentEvent`, or is it superfluous?

machour:
I didn't see that wording yet, so I'm not sure. But I think it's superfluous for now, as I don't think the current translation system is okay in general.

I'll open a new ticket with all my thoughts on the subject and a much cleaner implementation proposal where `en.js` wouldn't be needed anymore, and translators wouldn't have to struggle with words as some translations are used in different places with different meanings.

lex111:
Specifically about these cases - this is an exception, if GitHub returned the correct action, would not have to do a substitution.

It would be interesting to know about the cleaner implementation, well, let's discuss this in the new issue.

src/auth/screens/events.screen.js
```diff
@@ -155,7 +155,7 @@ class Events extends Component {
             'auth.events.pullRequestReviewCommentEvent',
             language,
             {
-              action: translate(`auth.events.actions.${action}`, language),
+              action: translate('auth.events.actions.commented', language),
```
housseindjirdeh: I'm assuming it's incorrect because passing in `created` is why the wrong message shows right? Perfectly okay with having `.commented` specifically here in that case

housseindjirdeh:
Thanks for this homie 👍

PR #266, opened 2017-08-17
fix(profile): prepend user.updated_at to avatar url

### LINKED ISSUE #265: Current user avatar never updates
I changed my GitHub avatar some days ago, but GitPoint keeps on showing me the old one (bearded me).

I even tried to sign out & sign in again, but the old avatar stays.

machour:
Gives GitPoint an additional chance of refreshing an updated profile picture

Close #265

### DIFF
--- a/src/components/user-profile.component.js
+++ b/src/components/user-profile.component.js
@@ -86,8 +86,8 @@ export const UserProfile = ({
       <ImageZoom
         uri={{
           uri: initialUser.avatar_url
-            ? `${initialUser.avatar_url}&cache=${initialUser.updated_at}`
-            : `${user.avatar_url}&cache=${user.updated_at}`,
+            ? `${initialUser.avatar_url}&lastModified=${initialUser.updated_at}`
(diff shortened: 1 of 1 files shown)


src/components/user-profile.component.js
```diff
@@ -86,8 +86,8 @@ export const UserProfile = ({
       <ImageZoom
         uri={{
           uri: initialUser.avatar_url
-            ? initialUser.avatar_url
-            : user.avatar_url,
+            ? `${initialUser.avatar_url}&cache=${initialUser.updated_at}`
```
lex111: Although the param `cache` could be called for example `dateUpdate` (to make it more meaningful), but I would generally convert the date into a timestamp, so it was cleaner, but I think it is superfluous.

andrewda: I think `lastModified` might be the best query name to use so it's as close as possible to the `last-modified` header we would pass to the API.

machour: Changed to `lastModified`.
@lex111 keeping the value as it is, as it's easier to read for debugging purpose

PR #269, opened 2017-08-19
Add reselect and redux actions to organizations

alejandronanez:
Hey there!
~I have a WIP branch with some updates on our codebase, it is still not done but I want to get early feedback on what I have here.~

~Will update this description as I make more progress - thanks for your patience!~

## Update
I completed my work on this but I decided to not include `Immutable` (not in this PR at least), because reselect + redux actions can live without Immutable.

So, the structure of the code has changed slightly, it goes like:

1. We have a `organization.selectors.js` file. Here we're using `reselect` https://github.com/reactjs/reselect to "query" our store and get the data we need (reselect also has a bunch of benefits that you can read in their docs). This will save us some duplication when we want to get data from the store in different containers (we should reuse the selectors we have in the selectors file whenever we want to use an specific part of the state)
2. We have a `organization.constants.js` file, I took this approach because we can use the constants through our reducers and actions (and in our tests in the future) also, I find the `*.type.js` notation quite ambiguous
3. We no longer have that nasty `switch` statement in our reducer (I love this part)
4. We have individual actions that take care of specific parts of the state, for example we can call an action just to update the loading state of the organization repos or update the error state only on the  organization members - No more "batch" updates to the store (What I mean by that is having actions that takes care of 2+ things in our store) This will give us way more control over how our store is being updated.


src/organization/organization.selectors.js
```diff
@@ -0,0 +1,47 @@
+import { createSelector } from 'reselect';
+import { getOrganization as getOrganizationFromStore } from '../../root.reducer';
+
+export const getOrganization = createSelector(
```
alejandronanez: Benefits of using `reselect` instead of querying the store `by.doing.something.like.this`

```
Selectors can compute derived data, allowing Redux to store the minimal possible state.
Selectors are efficient. A selector is not recomputed unless one of its arguments change.
Selectors are composable. They can be used as input to other selectors.
```
More info https://github.com/reactjs/reselect

src/organization/screens/organization-profile.screen.js
```diff
@@ -12,18 +15,32 @@ import {
...
+  // Selectors
+  getOrganization,
+  getOrganizationRepositories,
+  getOrganizationMembers,
+  getOrganizationIsPendingOrg,
+  getOrganizationIsPendingRepos,
+  getOrganizationIsPendingMembers,
+} from '../index';
+
+const selectors = createStructuredSelector({
```
alejandronanez: This will be almost the same as `mapStateToProps` but it uses reselect.

src/organization/organization.type.js
```diff
@@ -1,5 +0,0 @@
-import { createActionSet } from 'utils';
-
-export const GET_ORG = createActionSet('GET_ORG');
```
alejandronanez: I got rid of this `*.type.js` file in favor of `*.constants.js` and `*.actions.js` files

src/organization/organization.reducer.js
```diff
@@ -7,63 +19,65 @@ const initialState = {
...
   isPendingRepos: false,
   isPendingMembers: false,
-  error: '',
+  organizationError: '',
+  organizationRepositoriesError: '',
+  organizationMembersError: '',
 };
 
-export const organizationReducer = (state = initialState, action = {}) => {
-  switch (action.type) {
```
alejandronanez: Got rid of this ugly `switch` statement.

src/organization/organization.reducer.js
```diff
@@ -7,63 +19,65 @@ const initialState = {
   isPendingOrg: false,
   isPendingRepos: false,
   isPendingMembers: false,
-  error: '',
+  organizationError: '',
```
alejandronanez: Created one error for organization, another for repositories and another for members.

src/organization/organization.record.js
```diff
@@ -0,0 +1,3 @@
+import { Record } from 'immutable';
+
+// WIP
```
alejandronanez: Immutable work should go here =)

src/organization/organization.action.js
```diff
@@ -1,68 +1,80 @@
...
+  GET_ORG_REPOS,
+  GET_ORG_REPOS_LOADING,
+  GET_ORG_REPOS_ERROR,
+  GET_ORG_MEMBERS,
+  GET_ORG_MEMBERS_LOADING,
+  GET_ORG_MEMBERS_ERROR,
+} from './organization.constants';
 
-    dispatch({ type: GET_ORG.PENDING });
+export const getOrg = createAction(GET_ORG);
```
alejandronanez: Creating actions by using [redux-actions ](https://github.com/reduxactions/redux-actions) I think this is more organized than have a manual dispatch with some boilerplate like `{type: ... , payload: ...}`

root.reducer.js
```diff
@@ -16,3 +16,7 @@ export const rootReducer = combineReducers({
   search: searchReducer,
   notifications: notificationsReducer,
 });
+
+// Used to get organization information in selectors
+export const getOrganization = state => state.organization;
```
alejandronanez: This is used in the selectors to get some data.

src/organization/organization.reducer.js
housseindjirdeh: OMG how have I not used `redux-actions` before, it looks so much cleaner :heart_eyes:

src/organization/organization.type.js
housseindjirdeh: Nice this is perfect, using `redux-actions` looks like we don't need a helper utility like `createActionSet` whatsoever

src/organization/organization.action.js
housseindjirdeh: Nice agreed,  makes things look so much more oganized :+1:

src/organization/organization.selectors.js
```diff
@@ -0,0 +1,48 @@
+import { createSelector } from 'reselect';
+
+const getOrganizationFromStore = state => state.organization;
+
+export const getOrganization = createSelector(
+  getOrganizationFromStore,
+  organization => organization.organization || {}
```
housseindjirdeh: I was having a disussion with colleagues from work this past week and a few were under the impression that when simply getting object parameters with dot notation doesn't necessarily _need_ `createSelector`. For example, this would be sufficient:

```
export const getOrganization = state => state.organization.organization;
```

Not necessarily sure if I agree however. Majority of all our selectors are simple getters and not involve manipulation of data in anyway so it seems if we went with this approach we wouldn't really be using `createSelector` for the majority of our state.

Just wanted to get your input on this mate. Do you usually use `createSelector` for all data or _only_ if you're retrieving data and performing manipulation to it?

alejandronanez: Hey dude! I usually use it to query the store and do some computations based on that data when needed. The big plus I see is that we can compose any selectors we want, for example combine one selector from organizations, one from repos and two from auth, then produce a new prop to pass to any component we want. I truly believe this is a good adition to the stack.

Here's a copy paste from reselect's site

```
Selectors can compute derived data, allowing Redux to store the minimal possible state.

Selectors are efficient. A selector is not recomputed unless one of its arguments change.

Selectors are composable. They can be used as input to other selectors.
```

alejandronanez:
@housseindjirdeh yes! I think that's the right approach! 🙌🏼

PR #281, opened 2017-09-01
fix(profile): use affiliation=own to get user repos

blakegong:
### Problem

The number of repositories on your own profile page is the same as shown on github.com, which is 
`(user.public_repos + user.total_private_repos)`. However, when the button is clicked, the repository list actually shows all the repositories where you have `admin` permission. The *delta* here as I observed, is the repositories in a private organization you are part of where you are also granted `admin` permission to.

image

Another minor issue is the current way of querying extra data then filter in client is inefficient compared to querying the exact data that the client actually wants:

https://github.com/gitpoint/git-point/blob/809a8092c949f3a3376e978883de9ccbfbe8b5d2/src/user/user.action.js#L132-L136

### Approach

This PR uses `affiliation=own` from https://developer.github.com/v3/repos/#parameters to get user repos, which provides the exact data needed here. The number of repositories returned in this PR matches the number on user's profile page, which is `(user.public_repos + user.total_private_repos)`.


.all-contributorsrc
```diff
@@ -275,6 +275,15 @@
       "contributions": [
         "code"
       ]
+    },
+    {
+      "login": "blakegong",
```

blakegong: Wow this is fast! Thanks 😄

PR #284, opened 2017-09-03
feat(translation): add brazilian portuguese translations

arthurdenner:
Hi, guys.

Sorry for the wait to make this PR, I was very busy during the week.

I want to point that my last commit was a fix to the changes made in the turkish file after committing (idk what happened tbh, sorry about any mess that this may cause).

While translating, I created a function to make "one line texts" (I don't like to scroll horizontally xD). Hope that's OK. Also, during testing, I found three texts without translation available through files, so I changed them. Hope that's OK also.

Last but not least, sorry about the unformatted commits, I totally forgot about git cz (in 12 minutes, because my first commit was using it, smh).


src/locale/languages/br.js
```diff
@@ -0,0 +1,330 @@
...
+      hh: '%dh',
+      d: '%dd',
+      dd: '%dd',
+      M: '%dmo',
+      MM: '%dmo',
+      y: '%dy',
+      yy: '%dy',
+    },
+  },
+};
```
andrewda: I'm not a language expert by any means. How different are Portuguese and Brazilian Portuguese? If they're very similar (UK English vs. US English), I think it probably makes more sense just to use "Portuguese".

src/auth/screens/language-settings.js
```diff
@@ -15,4 +15,8 @@ export default [
     code: 'tr',
     name: 'Türkçe',
   },
+  {
+    code: 'br',
```
lex111: Because this is not pure Brazilian, the code should be such `pt-br`, this also applies to the file name.

src/locale/languages/br.js
arthurdenner: In my opinion, Portugal and Brazilian Portuguese are not very similar like UK and US English, altough they share a lot of aspects, but vocabulary, phonetics and syntax are differ a bit. But yeah, maybe we could use the same translation.

src/auth/screens/language-settings.js
arthurdenner: I was going to use `pt-br`, but the [instructions](https://github.com/gitpoint/git-point/blob/master/CONTRIBUTING.md#translations) says that we should use a two-letter code from [here](https://en.wikipedia.org/wiki/ISO_3166-1_alpha-2#Officially_assigned_code_elements), so I used only `br`. Should I change?

src/locale/languages/br.js
andrewda: Alright, if they're not too similar then this is totally fine!

src/auth/screens/language-settings.js
andrewda: Yea, those instructions should probably be changed. In this case, `pt-br` makes perfect sense.

arthurdenner: Okay, I will change the code and the file name as soon as I can. Thank you both for the suggestions and corrections.

arthurdenner: Changes made.

src/components/repository-profile.component.js
```diff
@@ -206,7 +206,9 @@ export const RepositoryProfile = ({
             ? abbreviateNumber(repository.watchers_count)
             : ' '}
         </Text>
-        <Text style={styles.unitText}>Watchers</Text>
```
housseindjirdeh: Thanks so much for noticing some stray translations <3

src/utils/text-helper.js
```diff
@@ -9,3 +9,5 @@ export const emojifyText = text => {
 export const abbreviateNumber = count => {
   return count >= 1000 ? (count / 1000).toFixed(1) + thousandUnit : count;
 };
+
+export const oneLineText = text => text.replace(/\s\s+/g, ' ');
```
housseindjirdeh: :+1:

src/auth/screens/language-settings.js
```diff
@@ -15,4 +15,8 @@ export default [
     code: 'tr',
     name: 'Türkçe',
   },
+  {
+    code: 'ptBr',
```
lex111: The code must be written with a hyphen (`pr-br`); It will then be used in the library `moment`, and there it is necessary to write the language through a hyphen.

arthurdenner: Sorry, but I'm not sure how to proceed... Shouldn't this be the same name I use on the file and to export the translation (ptBr)? Because when I try to use `pt-br`, selecting the language doesn't work and it fallbacks to English.

arthurdenner: @lex111, sorry, but I'm not sure how to proceed... Shouldn't this be the same name I use on the file and to export the translation (ptBr)? Because when I try to use pt-br, selecting the language doesn't work and it fallbacks to English.

arthurdenner:
@housseindjirdeh I just committed updating the branch with the last master commit, changing what lex requested and, following your request, adding myself as contributor (to spare another PR). I guess everything is okay now? If not, let me know, please.

andrewda:
Looks all good except for the small merge conflicts. I can try to resolve.

PR #208, opened 2017-08-06
chore(update): update react-native to 0.47.1

lex111:
All is well, we are waiting for the update of the dependency [react-native-photo-view](https://github.com/alwx/react-native-photo-view) (PR [open](https://github.com/alwx/react-native-photo-view/pull/76)). I hope this will be solved in the near future :blush:

PR #303, opened 2017-09-06
refactor(onboarding-swiper): replace component

lex111:
Unfortunately after updating React Native to 0.47.1 in PR #208, stopped working `onboarding swiper` (#174). I noticed this just now, because I left from  the GitHub account. And since the application has not been updated in the App Store/Google Play for a month, then this problem has not been detected, so we need to update the application _more often_ to detect such problems as quickly as possible.

On Android does not applies the background color (screenshot below), issue was [created](https://github.com/leecade/react-native-swiper/issues/511), but so far it has not been fixed and it is not known when it will be fixed. 

Therefore, I propose to replace the `react-native-swiper` component with the `react-native-app-intro`. All looks also, except for the sizes of dots, they are a little more, but in fact it is not the big problem?

### DIFF
--- a/package.json
+++ b/package.json
@@ -64,7 +64,7 @@
     "react-native-photo-view": "^1.4.0",
     "react-native-safari-view": "^2.0.0",
     "react-native-search-bar": "^3.0.0",
-    "react-native-swiper": "^1.5.4",
+    "react-native-app-intro": "^1.1.5",
     "react-native-syntax-highlighter": "^1.2.1",
     "react-native-vector-icons": "^4.0.0",
     "react-navigation": "^1.0.0-beta.11",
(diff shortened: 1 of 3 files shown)


lex111:
> Was about to open an issue today because I noticed it just last night as well :O Do you happen to know why it stopped working all of a sudden?

The only thing I know is that this is due to RN 0.47.1. I tried to figure it out, but it will take a long time, and not the fact that it will work out something, so it's better to replace the component, especially since there is an alternative.

housseindjirdeh:
One thing I noticed was `react-native-app-into` uses `react-native-swiper` under the hood. Is it an extension of it with added features?

> The only thing I know is that this is due to RN 0.47.1. I tried to figure it out, but it will take a long time, and not the fact that it will work out something, so it's better to replace the component, especially since there is an alternative.

Agreed, no need to spend significant amounts of time troubleshooting if we can find a very suitable solution so quickly. Thank you mate :)

lex111:
@housseindjirdeh hmm, and true, but not the original version is used, but fork, but whatever it is, it works. I previously looked for similar alternatives, but found only this working component, so I decided to use it.

PR #311, opened 2017-09-07
fix(profile): fix the wrong `follows you` badge

### LINKED ISSUE #308: Error with the "follows you" state
Currently, no matter what user profile I visit in GitPoint, I see the "Follows you" badge, even for people not following me.

Example with @CameronSamuels :
 
API url: https://api.github.com/users/CameronSamuels/following

profile

Already spotted a typo in https://github.com/gitpoint/git-point/blob/master/src/user/screens/profile.screen.js#L124 where `length` is misspelled, but fixing it doesn't solve this issue.

chinesedfan:
Closed #308

### DIFF
--- a/src/user/screens/follower-list.screen.js
+++ b/src/user/screens/follower-list.screen.js
@@ -3,6 +3,7 @@ import { connect } from 'react-redux';
 import { FlatList, View } from 'react-native';
 
 import { ViewContainer, UserListItem, LoadingUserListItem } from 'components';
+import { getFollowers } from 'user';
 
 const mapStateToProps = state => ({
   user: state.user.user,
@@ -10,12 +11,23 @@ const mapStateToProps = state => ({
   isPendingFollowers: state.user.isPendingFollowers,
 });
 
+const mapDispatchToProps = dispatch => ({
+  getFollowers: (user, type) => dispatch(getFollowers(user, type)),
+});
+
 class FollowerList extends Component {
   props: {
+    getFollowers: Function,
     followers: Array,
     isPendingFollowers: boolean,
     navigation: Object,
   };
+
+  componentDidMount() {
+    const user = this.props.navigation.state.params.user;
+
+    this.props.getFollowers(user);
+  }
 
   keyExtractor = item => {
     return item.id;
@@ -51,4 +63,6 @@ class FollowerList extends Component {
   }
 }
 
-export const FollowerListScreen = connect(mapStateToProps)(FollowerList);
+export const FollowerListScreen = connect(mapStateToProps, mapDispatchToProps)(
+  FollowerList
+);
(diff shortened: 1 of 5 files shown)


machour:
This partially fixes the problem. Now, a user following me isn't shown as so :/

For example, the user `tunecino` is one of my followers, but its profile only shows that I'm following him.

machour:
Guys, after some analysis, here's what I've found:

- When on a user profile page, the `user` parameter passed to `getFollowers` is a string, not an object, so `user.login` is undefined in this context and `user` should be used instead: https://github.com/gitpoint/git-point/blob/master/src/user/user.action.js#L96
- In here, `auth` is an object, so we should test against `auth.login` instead: https://github.com/gitpoint/git-point/blob/master/src/user/screens/profile.screen.js#L124
- I think there's an original logic flaw, as isFollower is supposed to tells if a user X is following me, yet, the function takes the user X followers and compares them against my login, which makes **no sense at all**. 

We should either take the list of users X is following and compare them towards the authenticated user, or maybe use another lighter API call:

GH Api allows the following call: 
 https://api.github.com/users/tunecino/following/machour

If `tunecino` is following `machour`, the request will just respond with a 204 status code and no contents.
If `tunecino` is not following `machour`, the request will respond with a 404 status code

```bash
🍺  ~/gitpoint-clean (master)*$ wget https://api.github.com/users/tunecino/following/machour
--2017-09-08 14:50:05--  https://api.github.com/users/tunecino/following/machour
Resolving api.github.com... 192.30.253.117, 192.30.253.116
Connecting to api.github.com|192.30.253.117|:443... connected.
HTTP request sent, awaiting response... **204 No Content**
2017-09-08 14:50:07 (0.00 B/s) - ‘machour’ saved [0]

🍺  ~/gitpoint-clean (master)*$ wget https://api.github.com/users/tunecino/following/chinesedfan
--2017-09-08 14:50:13--  https://api.github.com/users/tunecino/following/chinesedfan
Resolving api.github.com... 192.30.253.117, 192.30.253.116
Connecting to api.github.com|192.30.253.117|:443... connected.
HTTP request sent, awaiting response... **404 Not Found**
2017-09-08 14:50:14 ERROR 404: Not Found.
```

Could you please review my assertions? My brain may have been scrambled by following/followers/isFollowing :) (maybe we do need to rename variables to something like `isFollowingMe`, .. )

chinesedfan:
@machour I agree with you.

The current method is not 100% right even if we fixed those bugs. Because obtained followers are just the first page, whose size is 100. I will try the specified API to check whether A is following B.

chinesedfan:
@machour Hope I really solved the problem this time. Renaming variables needs to modify too many codes, so I give up. But I add 2 comments in `src/user/user.reducer.js` to clarify differences of them: https://github.com/chinesedfan/git-point/blob/issue308/src/user/user.reducer.js#L18.

@lex111 @andrewda Please review again.

machour:
Nice work @chinesedfan, everything looks fine with my tests 👌

Issue #182, opened 2017-07-29
Strange behavior after logout


lex111:
After clicking on "Sign out", a browser opens with a suggestion to exit GitHub, but why? For example, do I just want to exit the application, not from the GitHub account?

![image](https://user-images.githubusercontent.com/4408379/28743281-022b36ea-744f-11e7-908d-42c73be394d6.png)

Just in that case, if I do not leave the account, ie. I just close the browser with a suggestion to sign out, and then again I log into the application, I will see the following:

| Events | Profile 
| - | - |
| ![image](https://user-images.githubusercontent.com/4408379/28743287-1b3f5832-744f-11e7-9b14-55b7bdd6c54d.png) | ![image](https://user-images.githubusercontent.com/4408379/28743288-1e3c94aa-744f-11e7-80c8-0841d5738391.png)

Although I should see the login screen? This is due to the fact that I did not leave the account on GitHub? It's strange, I think we need to consider this case, can this be fixed?

andrewda:
> After clicking on "Sign out", a browser opens with a suggestion to exit GitHub, but why?

If you don't sign out from the browser, when you sign in again GitHub will use your caches credentials. That means you wouldn't be able to sign in as a different user. One idea for fixing this is clearing the cache/cookies in the WebView, but I'm not sure how that can be done.

housseindjirdeh:
I know - it is strange and thanks for logging this as I was hoping to keep the discussion going.

So authentication in the app uses the [OAuth web flow](https://developer.github.com/apps/building-integrations/setting-up-and-registering-oauth-apps/about-authorization-options-for-oauth-apps/). User clicks sign in, is directed to GitHub on the browser where they need to type their user name and password and once that completes are redirected back to the app with an **access token**.

Now the first thing @RolfKoenders and I tried was when the user logs out, we just clear the access token from the persisted store and the user is taken back to the login screen which is perfect. However we noticed that when the user clicks sign in again - they're taken to the browser but because `Safari` (or `Chrome`/`FIrefox` if you're on Android) has the user's authentication information _cached_  so they don't have type their username and password again. This can be annoying if the user is trying to login with different credentials.

So as a workaround, this is what we did when the user clicks Log Out:

1. Clear the access token from the persisted Redux store
2. Open up the browser asking the user to sign out from there. When they do this, click `Done` and try logging in again - they'll have to reenter their username and password again which is what we want

housseindjirdeh:
[...]

Issue #221, opened 2017-08-08
Implement use of the in-app web view on Android


introwit:
**Current process:**
- Right now, when a user clicks on the `Sign In` button, the browser is fired up and the user needs to complete the GitHub authentication in the browser window.

**Problem:**
- In the practical sense using this flow, GitPoint is dependent on an another app to let users start using it.
- This creates problems for our app which are out of our own control.
- What if the browser doesn't fire up for some reason or there are any other issues?
- @mathieutu have faced one such [issue](https://github.com/gitpoint/git-point/issues/199) already.
- There might be others who haven't reported it and directly uninstalled our app.

**Solution:**
- Using the [Android System Webview](https://play.google.com/store/apps/details?id=com.google.android.webview&hl=en)
- There are alternatives for this as well, which I won't suggest because as @housseindjirdeh says things are not always black and white so would leave this upto the developer's preferences :smile:

andrewda:
Not sure if this is about what we want: https://github.com/lucasferreira/react-native-webview-android

housseindjirdeh:
I was looking for something _exactly_ like this. I agree - opening a separate application to allow the user to login is definitely not the smoothest flow :)

Opened a separate issue to address this earlier #109 but will close that one and leave this open. @introwit thank you, and not at all please don't hesitate to leave as many alternatives as you like!! The more we discuss, the more we discover.

Not sure if that exact library would suit our needs exactly but it looks very promising 🎉

housseindjirdeh:
If no one picks it up sooner, this will be the next thing I work on. Hopefully will have it implemented within the next few days 🙌

machour:
Fixed in #323

PR #323, opened 2017-09-11
feat(auth): use WebView for authentication

### LINKED ISSUE #109: Andriod Deep Linking
So when a user authenticates, we leverage `SafariView` (react-native-safari-view) [for iOS](https://github.com/gitpoint/git-point/blob/master/src/auth/screens/login.screen.js#L122-L132) so that a user isn't directed into a separate Safari tab and navigated back to the app.

For Android, we're still using `Linking` like we normally would and that takes the user to the default browser before navigating them back to the app itself (leaving both applications open). I'm not sure - but it would be extremely useful if we can find an Android version of [`react-native-safari-view`](https://github.com/naoufal/react-native-safari-view) or something along those lines.

Open to all kinds of suggestions!

machour:
Closes #109 and #221 

Rather than using another library to handle Android, I went ahead and used the WebView component from ReactNative. This seems to be working fine for me on iOS Simulator and my Android Device.

I would really love to have a thorough review of my code, as I feel I've kinda abused the view system here. (and it's the app entry point, so it's a *really* critical change 😆 )


machour:
Thanks! Although it still seems buggy on my device, I sometime find myself logged in with an empty events tab and nothing clickable. So still WIP.

src/auth/screens/login.screen.js
```diff
@@ -39,6 +47,32 @@ const styles = StyleSheet.create({
...
+  },
+  browserLoader: {
+    flex: 1,
+    position: 'absolute',
+    justifyContent: 'center',
+    alignItems: 'center',
+    height: '100%',
+    width: '100%',
+    textAlign: 'center',
+    backgroundColor: '#1f2327',
```
housseindjirdeh: Let's put `#1f2327` in the `colors` file. Can call it whatever you think sounds reasonable (`githubDark` or something if you like).

machour:
@housseindjirdeh I think I fixed the problem with iOS by using Linking only for Android.
I also disabled the cancel button when needed, revamped the Welcome screen to be consistant and cleaned up my UI mess.
Could you give it a new try?

src/auth/screens/login.screen.js
```diff
@@ -183,6 +184,22 @@ class Login extends Component {
     this.handleOpenURL({ url });
   };
 
+  onLoadEnd = e => {
+    const url = e.nativeEvent.url;
+
+    if (url === 'https://github.com/session') {
```
lex111: I'm just thinking out loud: can made a separate method, because this code line two times is used, then the meaning of this check will be more clear?

machour: Great advice, thanks! Refactored here: https://github.com/gitpoint/git-point/pull/323/commits/c2c4d565b591eae4b50b05a3ec2ba3b0041d0213

machour:
I went ahead and fixed #182 in this same PR by resetting the WebView's cookie as per @andrewda recommendation. The logout is now instantaneous (maybe even too quick 💭 )

I used https://github.com/joeferraro/react-native-cookies to clear the cookies, which needs some extra step in XCode and Android build, documented in their README. I didn't commit the `ios/` and `android/`directories cause I had to tweak a lot of weird things to have a working environment in the same place. I'd rather have one of you guys with a cleaner environment do it when merging this PR.

src/locale/languages/en.js
```diff
@@ -4,6 +4,9 @@
 export const en = {
   auth: {
     login: {
+      connectingToGitHub: 'Connecting to GitHub..',
+      preparingGitPoint: 'Preparing GitPoint..',
```
andrewda: Add an extra period to these: `Connecting to GitHub...` and `Preparing GitPoint...`.

lex111: @machour and more you need to add new keys to other languages, otherwise there will be errors.

machour: @andrewda thanks, fixed!
@lex111 currently if a string is missing, it's displayed with an error `Missing translation for tr.auth.login.cancel`. Isn't that good enough while working on master? Or should I introduce the keys in all language with the english sentence?

lex111: @machour yes, you need to add English sentences to all languages, just so that there are no such errors.

machour: @lex111 done buddy! I think I'll tackle down i18n right after auth & markdown 😄

src/locale/languages/fr.js
```diff
@@ -2,6 +2,9 @@
 export const fr = {
   auth: {
     login: {
+      connectingToGitHub: 'Connexion à GitHub..',
+      preparingGitPoint: 'Configuration de GitPoint..',
```
lex111: Missing dot :smile:

machour: not in french ;)

lex111: Wow, I did not know :smiley:

src/auth/screens/login.screen.js
```diff
@@ -236,6 +236,7 @@ class Login extends Component {
           <View style={styles.container}>
             <Modal
               animationType="slide"
+              onRequestClose={() => null}
```
lex111: :+1:

ios/GitPoint.xcodeproj/project.pbxproj
```diff
@@ -623,6 +642,7 @@
 			isa = PBXGroup;
 			children = (
 				273F19201F1E1AD5004B4410 /* libCodePush.a */,
+				48F9DE9B1F6DB8C60064194C /* libCodePush.a */,
```
machour: @housseindjirdeh not sure why Xcode added this line, you may want to have a closer look

ios/GitPoint.xcodeproj/project.pbxproj
```diff
@@ -1148,6 +1181,20 @@
 			remoteRef = 48F49F661F19026D0012FAD6 /* PBXContainerItemProxy */;
 			sourceTree = BUILT_PRODUCTS_DIR;
 		};
+		48F9DE9B1F6DB8C60064194C /* libCodePush.a */ = {
```
machour: same here @housseindjirdeh

machour:
@housseindjirdeh @andrewda @lex111 
I just cleaned my project and was able to include links in `android/` and `ios/` for `react-native-cookies`. I thing this PR is all set now.

I'd love to see this merged soon in order for it to be fully tested by everyone before making it to the next release 🙏

PR #324, opened 2017-09-11
feat(comment-input): submit comment on clicking on the icon instead text

lex111:
Now the send icon do nothing, although in many applications the message is sent by clicking on this icon, while the "Send" text is superfluous, so it was deleted. Now there is more free space for the text, which was the main point.

| Before | After
| - | -
|  |

### DIFF
--- a/src/components/comment-input.component.js
+++ b/src/components/comment-input.component.js
@@ -23,7 +23,6 @@ const styles = StyleSheet.create({
   },
   wrapper: {
     padding: 10,
-    marginLeft: 5,
     flexDirection: 'row',
     alignItems: 'center',
     justifyContent: 'center',
@@ -31,7 +30,6 @@ const styles = StyleSheet.create({
   textInput: {
     fontSize: normalize(12),
     flex: 1,
-    marginLeft: 15,
     marginRight: 5,
     color: colors.black,
     ...fonts.fontPrimaryLight,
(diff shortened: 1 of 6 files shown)


housseindjirdeh:
Nice thank you for thinking about how we can make the input field a little more user friendly 👍 

Definitely in favour of just using the icon as the send button since it feels more natural. However one thing that I liked was having `Add a comment...` inline with the text of previous comments. What do you think, do you think it's okay without this though? Maybe I'm overthinking padding here.

If you think it looks nicer the way it is, only thing I would suggest is adding a little more paddingLeft for the text. Looks a little squeezed on iOS:

image

lex111:
@housseindjirdeh I understood what you mean, but I removed the left padding, so that there was more space for the text, I would leave it as it is now. Let's find out the opinions of the other contributors :raised_hands:

housseindjirdeh:
Thanks @lex111, I'm okay with leaving the way it is and adding a little more left padding as well (to match the padding on the right hand side of the arrow icon) if everyone likes this as well.

Let us know what you think @andrewda @alejandronanez @CameronSamuels etc...

CameronSamuels:
@housseindjirdeh, I think removing the send icon on the left, and replacing the “post” text at the right with the send icon will be good to do (the same thing as the pr description).
I do think there should be more padding at the left (maybe 5px to 10px more).

lex111:
I will also explain that I made horizontal indents the same as for other elements, see the screenshot below:

image

Therefore, if we add more padding, although small, it will no longer be exactly on the grid

housseindjirdeh:
@lex111 fair enough, definitely okay with leaving things the way it is now. Will defer the decision to you if you feel like you'll rather have this however :)

lex111:
@housseindjirdeh can we also merge?

Issue #202, opened 2017-08-05
Code renders without newline in markdown


rkkautsar:
I'm writing this on GitPoint so I can't upload a screenshot, but I got no-newline code when I open keajs/kea readme, for example

rkkautsar:
[keajs/kea](https://github.com/keajs/kea)

![Screenshot](https://lh3.googleusercontent.com/XCD9K-e-IXqhmQvgyj3txlvfSsLgP2chzkLlim3wY5f96R3pzhU8RGzTNlAXqeLTE70905L15QmN_EnzO8FfKMTVy1tgNZGTuA8y3YxwRCWiaWeSjzd1qzkf3MD332pLw4SzPYdKYf0lbihSjDxkyFZkmfhCoecRrYrg2YUWgoTWrV9IFwT_6VH5SL4_Ax0EuiieVR1qqnejJ_AmE2G-Cp89gGSf2hhHrzNuCj1Xbzgm-9XVXccyMVp6wbAPY3NrysqIOiqrk0hJPIVFGmI-GD5SZK5umtE7diEtdkhwe-yM_w97hFwL8le3QsjXZiQ7_gFeAZJ4XJY7-B2xAwmkXjOX__AOO_9UyKFpdiRfID8Js5sBiMOkRST4HuRXmfWjJXXSpY3R823LQFjIatw_V4DwuxwcLouegtxLeBcr8qUu0HCaDNdc3Upk5oaBTyKlT0nyYWtuOalwEaJaTeterF7ZC2bF93NxQB7e6c8oGajF9Lf3u6iZa1dbT6Qmw3yWW7xSQYFeQUdGWlpJS8SNnNK1wCCq6cbe1o4qoTzswyAwRsvqDxWhCnHwli6mk6Ra8W9QIRYvWf1JuLVZzHFFzXFVC83GOtqu5J2lJykuz1rtcRgTbBxM-g=w540-h960-no)

wlisrausr:
Is this issue happened in android? Because in iOS, the codes in README looks good @rkkautsar

rkkautsar:
Yes it happened in android

machour:
Thanks a lot for your report.
This have been fixed in #236, it will be live in the next release.

Issue #106, opened 2017-07-18
Horizontal scroll on `README.md`


vpanjganj:
When you open a  markdown file  in the repo, while the content fit nicely on the screen, you get an annoying horizontal scrollbar.

RolfKoenders:
Im only experiencing this on some repo's not on all. When there is a horizontal scroll bar it is not visible (only when you scroll) and its because there is code or markdown tables which take up space. 
Not sure what a solution would be. Scrolling is fine here i guess?

machour:
Thanks a lot for your report.
This have been fixed in #236, it will be live in the next release.

Issue #285, opened 2017-09-04
Open in browser feature


randy3k:
It would to nice to have a button to open safari to perform anythings gitpoint is not (yet) able to perform, e.g. editing comment and look at commit history.

RolfKoenders:
Good point! Yesterday i had exactly the same thought. It would be nice to have a 'open in browser' button on the readme page for example.

CameronSamuels:
I have used another GitHub client, and I love that/this feature. This should be added.

housseindjirdeh:
I agree, this is something we should definitely have and shouldn't be too complicated. Here's where I'm thinking:

- Issues screen: an `Open in browser` ListItem in the settings screen.
- Repository Screen: an `Open in browser` action in the action list
- User Profile Screen: an `Open in browser` action in the action list
- Organization Screen: an `Open in browser` action in an action list. Will need to add action list icon just like we have it in the user profile screen.
- Auth User Profile Screen: an `Open in browser` action in the user settings screen.
- ReadMe screen: an `Open in browser` action in an action list. Will need to add an action list icon on the top right as well.

If there are any other screens that anyone think we can have this, please let us know :raised_hands: 

In terms of implementation, we should be able to just use `openURLInView` [in utils](https://github.com/gitpoint/git-point/blob/master/src/utils/method-helpers.js#L8-L18) and pass the URL we need. Hopefully, the URL we need is in state for each of the screens when we need it.

jouderianjr:
Hey everyone, I'm almost finishing this issue, probably today I will open a pull request, but I think this should go after the 1.3 version 💃

housseindjirdeh:
@jouderianjr :raised_hands: 

Happy to have this in the 1.4 release :tada:

PR #326, opened 2017-09-12
feat(markdown): Markdown support

machour:
I looked at #106 and #202 and implemented a better support for README.md rendering in the app, by introducing a GitHub flavoured MarkDown component:

  MarkDown

It takes care of decorating the markup html with a stylesheet, fix a weird rendering issue with Android, and renders it in a WebView, with basePath set in order to handle internal links correctly.

Having it as component mean that we can later re-use it to render `.md` files in the Code section of the app :)

I'm just facing two issues if someone can advise:

1. `github-markdown.css` is from this package: https://github.com/sindresorhus/github-markdown-css. It should be included as an asset in the app, but I couldn't do it as WebView is kind of "tricky" to stay polite.
2. on Android, it's impossible to horizontally scroll as it's intercepted by the app navigation and we move away from the screen. Seems to be correctly handled on the iOS simulator

And one design concern:

If you navigate inside the WebView using a link, and hit the back arrow on top of the screen, you'll get back to the repository screen. Do we want to intercept the click instead and navigate back in history? And is letting the user navigating freely in this WebView is a good thing? 🤔


src/components/markdown.component.js
```diff
@@ -0,0 +1,57 @@
+import React, { Component } from 'react';
+import { WebView, Platform } from 'react-native';
+
+export class MarkDown extends Component {
+  props: {
+    html: string,
+    baseUrl: string,
+  };
+
+  componentDidMount() {}
```
andrewda: Can be removed, I believe.

src/components/markdown.component.js
```diff
@@ -0,0 +1,57 @@
+import React, { Component } from 'react';
+import { WebView, Platform } from 'react-native';
+
+export class MarkDown extends Component {
```
andrewda: Should probably be named `Markdown`.

src/components/markdown.component.js
```diff
@@ -0,0 +1,57 @@
...
+
+      if (lastIndex < this.props.html.length) {
+        markdown += this.props.html.substring(lastIndex);
+      }
+    }
+
+    return `<html>
+    <head>
+      <meta name="viewport" content="width=device-width, initial-scale=1">
+      <link rel="stylesheet" href="https://machour.idk.tn/github-markdown.css" />
```
andrewda: I'm guessing you're working on this, but for the sake of documenting it, `https://machour.idk.tn/github-markdown.css` should probably be changed to a local file eventually.

src/components/markdown.component.js
machour: Renamed to MarkdownWebview in last commit

src/components/markdown.component.js
machour: oupsy, done :)

src/components/markdown.component.js
machour: That's one of the issue I was facing.
 I added the file to src/assets/css for linking.

machour:
@andrewda Thanks a lot for the positive feedback! I took care of your two comments and implemented markdown rendering for issues comments, with images and emojis handling 🕺 

I first was tempted to get rid of MarkdownWebview all together in favor of MarkdownHtmlview, but images in `.md` files can be hosted anywhere (https:// and non http://) and I don't think we want to allow arbitrary hosts in the app. Security would become a nightmare.

Now, thinking about properly rendering the `.md` files in the code section, I think we could use MarkdownWebview and the [Markdown API from GitHub](https://developer.github.com/v3/markdown/) but that might be kind of slow (one request to get the file contents, another one to transform it, and then WebView).

Maybe we could have gitpoint.co hosting a simple transformer and point our WebView to it whenever needed? 🤔

machour:
By the way, here are a few things to tweak in MarkdownHtmlview later on:

- transform `@user` to a link to a profile
- transform `#ID` to a link to an issue/pr
- follow internal links and go to Code section with them

src/components/markdown-htmlview.component.js
```diff
@@ -0,0 +1,311 @@
...
+    margin: 0,
+    ...textStyle,
+  },
+  strong: {
+    ...textStyle,
+    ...fonts.fontPrimaryBold,
+  },
+  em: {
+    ...textStyle,
+    ...fonts.fontPrimaryBold, // FIXME: we need an italic font
```
housseindjirdeh: Should italic font go here?

src/components/markdown-webview.component.js
```diff
@@ -0,0 +1,55 @@
...
+    html: string,
+    baseUrl: string,
+  };
+
+  decorateHtml = () => {
+    let markdown = this.props.html;
+
+    if (Platform.OS === 'android') {
+      // Android's webview won't correctly render new lines in <pre> leading
+      // to code being rendered on a single line.. we convert those to <br />s
```
housseindjirdeh: 👍

housseindjirdeh:
This is great @machour 👏  👏 Tomorrow I'll pull this down and test it 😍 

Still trying to understand why MarkdownWebview and the Markdown API from GitHub will be slow 🤔 And what do you mean by a transformer? Sorry for the silly questions 😭

src/components/markdown-htmlview.component.js
machour: yes, to render *this markdown* correctly. We need to bundle it like we do for other variants

machour:
@housseindjirdeh I'm just worried about an extra API call, that's all 😄 
Forget about the transformer, that was me thinking out loud and in the wrong direction 😉 

Let me know when you check this PR, and if you're able to use a bundled css file in the Webview rather than the external one I'm using now. Couldn't achieve that on my own.

src/components/comment-list-item.component.js
```diff
@@ -89,155 +89,21 @@ const styles = StyleSheet.create({
...
-
 const mapStateToProps = state => ({
   authUser: state.auth.user,
 });
 
 class CommentListItemComponent extends Component {
   props: {
     comment: Object,
-    onLinkPress: Function,
+    // onLinkPress: Function,
```
alejandronanez: Oops! I think you missed this one

src/components/markdown-htmlview.component.js
```diff
@@ -0,0 +1,356 @@
...
+};
+
+const regularFont = {
+  ...fonts.fontPrimary,
+};
+
+const textStyleLight = {
+  fontSize: Platform.OS === 'ios' ? normalize(11) : normalize(12),
+  color: colors.primaryDark,
+  ...lightFont,
```
alejandronanez: I know this is a really controlled environment but you should write this like:
```js
const textStyleLight = {
  ...lightFont,
  fontSize: Platform.OS === 'ios' ? normalize(11) : normalize(12),
  color: colors.primaryDark
}
```
By doing that you'll avoid potential overwrites in the new object.

src/components/markdown-htmlview.component.js
```diff
@@ -0,0 +1,356 @@
...
+      switch (node.type) {
+        case 'text':
+          if (node.data === '\n') {
+            return <Text style={rendererStyles.newLine} />;
+          }
+
+          /* eslint-disable no-param-reassign */
+          // Convert checkboxes
+          node.data = node.data.replace(
+            /^\s*\[\s?\]/,
```
alejandronanez: What do you think about moving this regexp into a constant in order to improve its meaning?

Something like `const checkBoxes = /^.....` after doing that you can remove the comment. =]

src/components/markdown-htmlview.component.js
```diff
@@ -0,0 +1,356 @@
...
+          node.data = node.data.replace(/:(\w+):/g, text => {
+            const emoji = text.replace(/:/g, '');
+
+            return emojis[emoji] ? emojis[emoji] : text;
+          });
+
+          break;
+
+        case 'tag':
+          switch (node.name) {
```
alejandronanez: Personally, I'm not a big fan of `switch` statements (also they add complexity to the codebase).
What do you think about moving this into some kind of function? Or maybe use the strategy pattern to solve this?

I'd love to see a solution like that :)

src/components/markdown-htmlview.component.js
machour: Great readability advice, thanks!

src/components/markdown-htmlview.component.js
machour: definitely (wasn't well aware of that, thanks)

src/components/markdown-htmlview.component.js
machour: Do you mean something like this? http://robdodson.me/javascript-design-patterns-strategy/ (seems indeed way more readable)

alejandronanez:
Yep, that!

machour:
@alejandronanez thanks a lot, I applied your remarks in last commit.

Could you tell me if this is how the strategy pattern should be implemented?
I had to use `/* eslint-disable no-unused-vars */`, I hope that's okay.

Also, if you can advise: I think I'm doing something really wrong, if you look at `navigateToIssue` and `navigateToProfile` :
https://github.com/gitpoint/git-point/pull/326/files#diff-30d90f2bbbe7ca7893404044e8662a80R136 

I'm passing navigation from MakdownHtmlview as I don't have access to `this.props` anymore. Is that okay ?

And for the missing `onLinkPress` implementation you spotted, it was originally defined here, and I'd like to integrate it into `MakdownHtmlview`, but it requires the `authUser`.. I'm unsure if I need to propagate it as a prop from IssueScreen => CommentListItem => MarkdownHtml => as a callback parameter, or if there's a nicer way of getting this global state.. 
https://github.com/gitpoint/git-point/blob/master/src/issue/screens/issue.screen.js#L106

housseindjirdeh:
Just pulled it down to test, all in all README's seem to look just great :+1: 

image

Issue comments are also getting there as well. However I am noticing that there seems to be some slight font differences and styles with different element nodes: 

image

as well as the size seems a little too big, we might need a little more padding for some breathing room and for some reason: there seems to be line breaks when it seems there shouldn't :thinking: 

image

So all in all some smoothing needs to be done but it's really looking solid and almost there :heart_eyes: 

And to help answer your questions mate:

> I'm passing navigation from MakdownHtmlview as I don't have access to this.props anymore. Is that okay ?

Yep there are instances of the app where `navigation` is being passed down. However is there a particular reason why you can't access `this.props` in here?

> And for the missing onLinkPress implementation you spotted, it was originally defined here, and I'd like to integrate it into MakdownHtmlview, but it requires the authUser.. I'm unsure if I need to propagate it as a prop from IssueScreen => CommentListItem => MarkdownHtml => as a callback parameter, or if there's a nicer way of getting this global state..

`CommentListItem` is already stateful and is already referencing `authUser` from the state (https://github.com/gitpoint/git-point/blob/master/src/components/comment-list-item.component.js#L123). You should be able to just pass it down a single level chain from here :)

machour:
@housseindjirdeh I can't access `this.props` cause I'm outside of the `MarkdownHtmlview` object: https://github.com/gitpoint/git-point/pull/326/files#diff-30d90f2bbbe7ca7893404044e8662a80R136 I'm sure there's a better way to handle that :/

machour:
@housseindjirdeh I fixed the font issues and greatly simplified my work by understanding better how HtmlView works.

I needed to use only `<Text>` everywhere as the `<View>'s` reserved boxes and that what was making the weird wrapping.

I'm now struggling with two issues UI wise:

- I can't seem to place a View under h1 or h2 to emulate the divided line. No matter what I tried, the dividers if floating to the right of the text
- Same for blockquote, I can't draw a line to its right. I think I'm misunderstanding something with the flex logic (tried column/row/grow/shrink, nothing worked)

I also started supporting tables 🕺

machour:
Here's a big markdown chunk to showcase all features : 

It's very easy to make some words **bold** and other words *italic* with Markdown. 

You can even [link to Google!](http://google.com)

| Tables        | Are           | Cool  |
| ------------- |:-------------:| -----:|
| col 3 is      | right-aligned | $1600 |
| col 2 is      | centered      |   $12 |
| zebra stripes | are neat      |    $1 |

Sometimes you want numbered lists:

1. One
2. Two
3. Three

Sometimes you want bullet points:

* Start a line with a star
* Profit!

Alternatively,

- Dashes work just as well
- And if you have sub points, put two spaces before the dash or star:
  - Like this
  - And this

If you want to embed images, this is how you do it:
Image of Yaktocat


# h1
## h2
### h3
#### h4
##### h5
###### h6


If you'd like to quote someone, use the > character before the line:

> Coffee. The finest organic suspension ever devised... I beat the Borg with it.
> - Captain Janeway


There are many different ways to style code with GitHub's markdown. If you have inline code blocks, wrap them in backticks: `var example = true`.  If you've got a longer block of code, you can indent with four spaces:

    if (isAwesome){
      return true
    }

GitHub also supports something called code fencing, which allows for multiple lines without indentation:

```
if (isAwesome){
  return true
}
```

And if you'd like to use syntax highlighting, include the language:

```javascript
if (isAwesome){
  return true
}
```
GitHub supports many extras in Markdown that help you reference and link to people. If you ever want to direct a comment at someone, you can prefix their name with an @ symbol: Hey @kneath — love your sweater!

But I have to admit, tasks lists are my favorite:

- [x] This is a complete item
- [ ] This is an incomplete item

When you include a task list in the first comment of an Issue, you will see a helpful progress bar in your list of issues. It works in Pull Requests, too!

And, of course emoji! :sparkles: :camel: :boom: :notanemoji:

src/components/markdown-htmlview.component.js
alejandronanez: Yes!

machour:
Ok, I think this is as far as I can go for now on Markdown support inside comments before madness kicks in 😆 

screen shot 2017-09-16 at 12 29 23 am
screen shot 2017-09-16 at 12 30 32 am
screen shot 2017-09-16 at 12 30 13 am
screen shot 2017-09-16 at 12 29 55 am

Putting this for review and adoption (knowing that I still need someone to properly bundle the CSS file for MarkdownWebview which is used to render the README.md file through a WebView).

Cheers !

src/components/markdown-htmlview.component.js
```diff
@@ -0,0 +1,335 @@
...
+  transformMarkdown = md => {
+    const issueReference = /(\s)#(\d+)/;
+    const profileReference = /(\s)@([_0-9A-Za-z]+)/;
+    const todoItem = /<li>\s*\[(x|\s)?\](.*)<\/li>/g;
+    const emojiMarkup = /:(\w+):/g;
+
+    let rendered = marked(md);
+
+    rendered = rendered
+      .replace(new RegExp(/<p>*>/g), '<span>')
```
jouderianjr: Hey, should be better use Just `/<p>*>/g`, I think `new RegExp` is not necessary :)

src/components/markdown-htmlview.component.js
```diff
@@ -0,0 +1,335 @@
...
+    onLinkPress: Function,
+  };
+
+  transformMarkdown = md => {
+    const issueReference = /(\s)#(\d+)/;
+    const profileReference = /(\s)@([_0-9A-Za-z]+)/;
+    const todoItem = /<li>\s*\[(x|\s)?\](.*)<\/li>/g;
+    const emojiMarkup = /:(\w+):/g;
+
+    let rendered = marked(md);
```
jouderianjr: I think this `let` is not necessary, you can just 
```javascript
return marked(md)
   .replace...
```
This way we can avoid using let :) What do you think about?

src/issue/screens/issue.screen.js
```diff
@@ -120,12 +120,21 @@ class Issue extends Component {
       node.attribs.class &&
       node.attribs.class.includes('issue-link')
     ) {
-      navigation.navigate('Issue', {
-        issueURL: node.attribs['data-url'].replace(
+      let issueURL;
```
jouderianjr: I think this logic could be transformed in a method in order to avoid let

jouderianjr:
Hey, I'm not a reviewer but I've sent some comments in the code :)

housseindjirdeh:
Welcome to the team @jouderianjr :) You should be able to submit reviews now :raised_hands:

src/components/markdown-htmlview.component.js
```diff
@@ -0,0 +1,335 @@
...
+  return (
+    <View key={index} style={style}>
+      <Text style={{ ...styles[node.name] }}>
+        {defaultRenderer(node.children, node)}
+      </Text>
+    </View>
+  );
+};
+
+export class MarkdownHtmlview extends Component {
```
housseindjirdeh: the most minor nitpick of all nitpicks, but do you think we should name this `MarkdownHtmlView`? :)

src/components/markdown-htmlview.component.js
```diff
@@ -0,0 +1,335 @@
...
+  h3: { ...fonts.fontPrimarySemiBold, fontSize: normalize(18) },
+  h4: { ...fonts.fontPrimarySemiBold, fontSize: normalize(16) },
+  h5: { ...fonts.fontPrimarySemiBold, fontSize: normalize(14) },
+  h6: { ...fonts.fontPrimarySemiBold, fontSize: normalize(12) },
+  hr: { ...hrStyle },
+  li: textStyle,
+  a: linkStyle,
+};
+
+const stylessheet = StyleSheet.create(styles);
```
housseindjirdeh: Another minor nitpick :eyes:, but could we keep the variable camelcase like so: `styleSheet`. Also think it might be nicer to name it `styleSheet` rather than `stylesSheet` since it matches the RN API

src/components/markdown-webview.component.js
```diff
@@ -0,0 +1,55 @@
+import React, { Component } from 'react';
+import { WebView, Platform } from 'react-native';
+
+export class MarkdownWebview extends Component {
```
housseindjirdeh: Similarly here, `MarkdownWebView` since RN API also camel-cases `WebView`

src/components/markdown-htmlview.component.js
machour: I'm all for it. Was keeping it in a variable to be able to console.log() it ;)

machour:
@jouderianjr your comments are really appreciated and have been taken care of. Thanks a lot! ✋ 

@housseindjirdeh All done. I also embedded `OpenSans-Italic.ttf` for android, and loaded the css file from `file:///android_assets/`. 

I'm still unable to figure out how to do load a bundled CSS for iOS thought, so I'm still pointing to an external url, could you guys have a look at it ?

And last but not least, there's still the problem in Android webview where you can't scroll horizontally (to read a code sample), because the swipe is intercepted by the NavigationTab.. @jouderianjr is also having this issue for his PR : https://github.com/gitpoint/git-point/pull/327#issuecomment-329973461

Can we haz disable? 😆

housseindjirdeh:
@machour Thank you sooo much matey. Unfortunately, that's something I don't think we should be disabling :disappointed: I'm quite used to navigating that way and I'm sure quite a few folks are the same.

Isn't it only the far left of the screen that this `swipeEnabled` kicks in? Does it completely disrupt how `WebView` scrolls in Android even if my finger is on the middle of the screen?

And are you considering replacing `AvenirNext` for iOS with `OpenSans` throughout the app?

machour:
@housseindjirdeh just tried on a real Android device, no matter where I start scrolling, the swipeEnabled kicks in. The only was to scroll is to long press the `<pre>` until you get the copy/paste native popup, and extend the selection to the right, which is far from ideal.

If we want to keep swipeEnabled, here's a few options we might consider:

- tweak the CSS to make the `<pre>` contents wrap
- ditch `MarkdownWebView` completely and use `MarkdownHtmlView` to render the README, and fine tune `<pre>` handling to embed it in a scroll view.

(later option is something I'm considering for another PR, to have a strong `MarkdownView` component, dealing with all `md` related stuff in the app, and why not, extract it to `git-point/react-native-markdown-view` in the future).

But let's not make the scroll issue a show-stopper for now, especially if we're implementing #285 soon ;) 

For the fonts, I wouldn't dare messig with AvenirNext :D 
I'm using `AvenirNext-Italic` for iOS which was already bundled but not defined in `fonts.js`. I bundled `OpenSans-Italic` for Android: https://github.com/gitpoint/git-point/pull/326/files#diff-a0fa44442811d58ef50c6de7283b8916

housseindjirdeh:
@machour Nice thank you, appreciate you including `OpenSans- Italic` and sorry for not including it myself. Yep `AvenirNext` is included as system font for iOS.

100% agreed. i'm more than happy to merge this in for now and we can open an issue to fix the scroll issue on Android in the near future :star:

Issue #187, opened 2017-07-30
PoC for Immutable - Reselect


alejandronanez:
I'm creating this issue to keep track of the PoC with Immutable + Reselect, will add those to a small part of the app to see what everybody thinks.

housseindjirdeh:
Looking forward to it matey, any help please don't hesitate to ask :raised_hands:

zcei:
Out of interest: what do you mean "Immutable"? [Immutable.js](https://facebook.github.io/immutable-js/)?

Wondering whether that's really necessary when already utilizing `redux`, which embraces immutable data structures, but with the plain vanilla js data types.

👍 for `reselect`

alejandronanez:
Yes I meant Immutable.js, with it we can have some nice _models_ + will be forced to manage immutability (no risk of mutating the state at all)
Redux needs immutability but there's nothing stopping you from mutating the state.

zcei:
Mhm I like the idea of clearly defined models, just never saw the purpose of putting more overhead on top of the plain types. But yeah, it puts more responsibility on the contributors.
(Otherwise one would likely put its own layers on top, like `joi` schema validation.)

Thank you for the clarification!

alejandronanez:
You're welcome @zcei, we'll see how it goes.
Does Joi enforces Immutability? I have no experience with it.

zcei:
The `joi` objects and methods themselves are immutable, but I don't think they enforce it on the objects that are checked. That was more for the "clearly defined model" part.

If you're really scared people would mutate the state, in development one could [freeze the state tree](https://github.com/buunguyen/redux-freeze) (which for performance reason should never be a production middleware 🙂 ).

I'm just afraid that Immutable.js actually makes it harder for people to contribute and feels like an "eternal lock in" as now each and every piece of the app (except for the presentational components) needs to be aware it is not receiving a plain object, but a special wrapper.

housseindjirdeh:
I also feel the same with introducing a library like `Immutable.js` hence why I was hesitant at first as well. However definitely open to giving it a shot if everyone feels like the benefits of including it outweigh any potential concerns.

alejandronanez:
I'm sure we at least use something of my work on this haha ;)

alejandronanez:
We have a PR for this in #269

G2Jose:
[...]

PR #327, opened 2017-09-13
feat(comments): allow editing and deleting issue comments

### LINKED ISSUE #24: Allow for editing and deleting issue comments
Would definitely be awesome to have the ability to edit and delete issue comments!

jouderianjr:
Closes #24 

Before

After

<p align="center>
    
    
    
</p>

Thanks in advance :)


src/components/comment-list-item.component.js
```diff
@@ -243,10 +269,20 @@ class CommentListItemComponent extends Component {
...
 
+    const commentActionSheetOptions = [
+      translate('issue.comment.editAction', language),
+      translate('issue.comment.deleteAction', language),
+    ];
+
+    const isActionButtonVisible =
+      comment.user &&
+      authUser.login === comment.user.login &&
+      !Object.prototype.hasOwnProperty.call(comment, 'repository_url');
```
lex111: For what this check?

src/components/comment-list-item.component.js
```diff
@@ -284,29 +321,43 @@ class CommentListItemComponent extends Component {
...
+          )}
 
           <View style={styles.dateContainer}>
             <Text style={styles.date}>
               {moment(comment.created_at).fromNow()}
             </Text>
           </View>
+
+          {isActionButtonVisible && (
+            <View style={styles.iconContainer}>
```
lex111: Let's for greater clarity, rename the class to something like `actionButtonIconContainer`?

src/issue/issue.reducer.js
```diff
@@ -17,6 +19,8 @@ const initialState = {
   isMerged: false,
   isPendingComments: false,
   isPostingComment: false,
+  isDeletingComment: false,
+  isEditingIssueComment: false,
```
lex111: Maybe better just `isEditingComment`?

src/issue/screens/issue.screen.js
```diff
@@ -202,6 +208,27 @@ class Issue extends Component {
...
   };
 
+  deleteComment = comment => {
+    const { repository } = this.props;
+    const { id } = comment;
+
+    const repoName = repository.name;
+    const owner = repository.owner.login;
+
+    this.props.deleteIssueCommentByDispatch(id, owner, repoName);
```
lex111: I do not know for sure, but it makes sense to define the variable `id`, if it will be used once, I mean what maybe just pass `comment.id`?

src/locale/languages/en.js
```diff
@@ -248,6 +248,12 @@ export const en = {
       areYouSurePrompt: 'Are you sure?',
       applyLabelTitle: 'Apply a label to this issue',
     },
+    comment: {
```
lex111: Please duplicate these all new keys into the remaining language files.

src/issue/screens/issue.screen.js
```diff
@@ -254,12 +283,19 @@ class Issue extends Component {
...
       navigation,
     } = this.props;
 
-    const isLoadingData = !!(isPendingComments || isPendingIssue);
+    const isLoadingData = !!(
+      isPendingComments ||
+      isPendingIssue ||
+      isDeletingComment
+    );
+    const showLoadingContainer = isPendingComments || isPendingIssue;
```
lex111: Since this variable contains a boolean value, it can be renamed to `isShowLoadingContainer`?

src/issue/screens/edit-issue-comment.screen.js
```diff
@@ -0,0 +1,144 @@
...
+
+  render() {
+    const { language, repository, isEditingIssueComment } = this.props;
+    const { issueComment } = this.state;
+
+    return (
+      <ViewContainer>
+        {isEditingIssueComment && <LoadingModal />}
+        <ScrollView>
+          {repository.full_name && (
```
lex111: What do you think, besides the name of the repository, can show the title issue?

src/issue/issue.action.js
```diff
@@ -117,6 +121,56 @@ export const postIssueComment = (body, owner, repoName, issueNum) => {
   };
 };
 
+export const deleteIssueComment = (issueCommentId, owner, repoName) => {
+  return (dispatch, getState) => {
+    const accessToken = getState().auth.accessToken;
+
+    dispatch({ type: DELETE_ISSUE_COMMENT.PENDING });
```
alejandronanez: I know this is not in the scope of this PR but what do you think about using `redux-actions` instead? We can reduce boilerplate significantly.

You can find and example in this PR #187 

If you're not feeling ok with it, that's fine too!

src/components/comment-list-item.component.js
jouderianjr: Because the description of the issue it is also a commentListItem, so I only show the button if is a comment and when is an issue description it has this field (repository_url), does that make sense?

I think I could extract this `!Object.prototype.hasOwnProperty.call(comment, 'repository_url');`for a function for readability, what do you think?

src/issue/screens/issue.screen.js
jouderianjr: yeah, you are completely right! 😄

src/locale/languages/en.js
jouderianjr: Should I try to translate? or I add the text in English?

src/issue/screens/edit-issue-comment.screen.js
jouderianjr: Sounds good to me

src/issue/issue.action.js
jouderianjr: Oh, that's pretty amazing, I really like this approach! But to be honest, I think this should be done in a separated PR, I think is better just refactor the whole issue.action.js instead only this particular part, just for keeping the cohesion of the file.

What do you think about? 😄

alejandronanez:
I completely agree! Thanks!!

src/locale/languages/en.js
lex111: Generally not, although if you know any language and can translate, it would be good, but if not, then it is enough to duplicate it in English, otherwise there will be errors (because new keys will not be in the remaining languages)

src/issue/issue.action.js
alejandronanez: Completely agree!

Thanks!!!

jouderianjr:
@andrewda yeah, that can be done. To be honest like much this approach because we are maybe "hiding" some features from the user, but I'm not against this approach 😄 

I would love to get more opinions about, any thoughts @lex111 @alejandronanez @housseindjirdeh?

jouderianjr:
@lex111 I did most of all improvements, could you take a look? 😄

lex111:
@jouderianjr now everything is ok :ok_hand: 

As still a variant with actions, you can look at this https://github.com/AidenMontgomery/react-native-swiping-row

jouderianjr:
Cool, So, now we have basically 3 options :) Which one you guys would prefer?

src/issue/screens/issue.screen.js
```diff
@@ -240,6 +266,8 @@ class Issue extends Component {
       <CommentListItem
         comment={item}
         onLinkPress={node => this.onLinkPress(node)}
+        onDeletePress={comment => this.deleteComment(comment)}
+        onEditPress={comment => this.editComment(comment)}
```
chinesedfan: I think we can write shorter as `onDeletePress={this.deleteComment}`. Maybe another pr is required if you guys(@lex111, @jouderianjr) agree with it.

Of course, there are so many places like this. And it doesn't matter.

jouderianjr: Yeah, I agree with you, I do prefer the shorter version, I just tried to follow the file pattern, but I also agree in do another PR for that, should be better change all the file.

andrewda:
@lex111 :gasp: Swiping is a really good idea!

jouderianjr:
Cool, 2 votes 😄 I'll implement with swipe action

jouderianjr:
@lex111 @andrewda hey guys, bad news :/ Doing the implementation I realized that we already have swipe on these screens because of tab navigation :( because of that I think that cannot be done.

So, should I implement the long press action?

jouderianjr:
@andrewda I've implemented with long press click. Could you take a look?

lex111:
@jouderianjr it's a pity, but there is no way to disable swipe on the screen issue?

machour:
I'm having the same issue with TabNavigation swipe that prevents scrolling horizontally in webviews on Android (Readme.md is displayed in webview in #326) 

I would love to have the TabNavigation swipe disabled all together in the app. I personally find this type of navigation confusing, especially when you know that you can have the Issue screen (for example) displayed in all 4 tabs at the same time..

jouderianjr:
I think disabling the swipe only in a few screens not intuitive. We could disable for the whole tab navigation, is just set `swipeEnabled: false` in TabNavigation options, but this is a big change to do, I think we need to get thoughts from the @housseindjirdeh as well :)

housseindjirdeh:
Man thank you all so so much for thinking this through.

So some thoughts:

1. I highly agree we _shouldn't_ remove swipe on this screen or any screen. I'm actually used to swiping on the left side of the screen to go back and it would feel to weird if some screens didn't allow this. That being said, we could remove it altogether but I would be against this as it feels like a natural way to navigate in my opinion.
2. That being said, @jouderianjr what do you think of adding the swipe feature and leaving the `swipeEnabled` for navigation. Are you noticing significant interference? Not sure how exactly that swipe library works, but wondering why it wouldn't be possible to have the swipe throughout the comment list item and _only_ when we swipe on the very very left side of the screen - the navigation swipe kicks in just like the way it is now.

sep-17-2017 01-36-00

If they both don't seem to coexist together, then we'll have to forego swipe I'm afraid. However, I think this is definitely something we should explore if we can.

3. If we can't add swipe or this feature, then I'm in favour of long touch _and_ possible including ellipsis as well. From a UI perspective, I think a lot of users might not realize long touch is enabled and I do think a secondary method to open the action sheet would be ideal. Styling-wise, I think it might look better if we slightly reduce the size of the ellipsis icon and have it on the bottom right of the comment list item rather than competing with the time at the top right.

Let me know what you folks think :speech_balloon:

jouderianjr:
@housseindjirdeh Cool! The swipe feature really improves the UX, I will do more tests in order to add that for the app 😄

vitalkanev:
@housseindjirdeh This screenshot is not from the actual gitpoint app :joy: It's from music app?!

housseindjirdeh:
Haha @vitalkanev I know :joy: :joy: It was the first app I can think of (Spotify) that had a really nice swipe feature to remove and add songs to your library. Just wanted to show how even though I can still navigate from swiping the left side of the screen, the swipe still works well in the context of each item.

Probably should have explained why I threw that `.gif` in :joy:

housseindjirdeh:
And thank you @jouderianjr, please keep us posted. If it turns out to not be the smoothest experience, more than happy to use the simpler approach :dancer:

jouderianjr:
Unfortunately, the swipe feature is not possible because of how the tabNavigation is implemented :/

giphy

I've updated the code with @housseindjirdeh suggestions, we have long press and the icon button in the bottom



What do you think guys? 😄

src/components/comment-list-item.component.js
```diff
@@ -146,102 +145,108 @@ class CommentListItemComponent extends Component {
...
-            >
-              <Image
-                style={styles.avatar}
-                source={{
-                  uri: comment.user.avatar_url,
-                }}
-              />
-            </TouchableOpacity>
-          )}
+      <TouchableWithoutFeedback onLongPress={this.showMenu}>
```
lex111: Maybe use `TouchableOpacity` better to show feedback?

jouderianjr: yeah, looks better!

jouderianjr:
Hey guys, don't forget my PR 🙉 @housseindjirdeh @andrewda

src/issue/screens/edit-issue-comment.screen.js
```diff
@@ -0,0 +1,146 @@
...
+    const { language, repository, isEditingComment } = this.props;
+    const { issueComment } = this.state;
+    const issueTitle = this.props.navigation.state.params.issue.title;
+
+    return (
+      <ViewContainer>
+        {isEditingComment && <LoadingModal />}
+        <ScrollView>
+          {repository.full_name && (
+            <ListItem
```
housseindjirdeh: Feels kinda out of place to have this `ListItem` here, I think we can remove this to be honest (the user can easily see the back navigation step with the issue number)

src/issue/screens/edit-issue-comment.screen.js
```diff
@@ -0,0 +1,146 @@
...
+                type: 'octicon',
+              }}
+              hideChevron
+            />
+          )}
+
+          <SectionList
+            title={translate('issue.newIssue.issueComment', language)}
+          >
+            <TextInput
```
housseindjirdeh: This is great stuff. The only think I'm noticing is that with super long comments, this `TextInput` is massive

screen shot 2017-09-22 at 2 22 54 pm

That blocks the `ScrollView` scroll capability and hides the the submit button. We'll need to have a fixed height for this input field (or a max height if that's possible - will be better if we can)

housseindjirdeh: One other thing is that it seems slightly squeezed vertically. 

screen shot 2017-09-22 at 2 20 09 pm

A little `paddingVertical` for the TextInput would be ideal :)

src/issue/screens/issue.screen.js
housseindjirdeh: Yeah I've used arrow syntax within render callbacks throughout the app. Leveraging `onDeletePress={this.deleteComment}` is better and I think I'll dedicate some time to do that throughout the app

src/components/comment-list-item.component.js
housseindjirdeh: Also I notice that my initial issue post doesn't seem to allow me to edit:

screen shot 2017-09-22 at 2 17 25 pm

Not sure if this is where you've set that to happen, but I think it would make sense to show the ActionSheet but _only_ allow for editing for the first issue comment and not deleting. Happy to defer this if you feel like taking care of this in a separate PR however :)

jouderianjr: I'm using `repository_url` for distinguishing between issue description and issue comment, only issue description has this field like I said before I think is better extract that for a method in order to improve the "understanding".

About editing the description, you are completely right, I didn't realize this possibility, I will take care of that in this PR. 😄

jouderianjr:
Hey @housseindjirdeh I set a fixed height to the TextInput, I don't know if that is the best approach :/ Any thoughts about that?

Also, about the "edit issue description" feature I realize that is better handled with that in another PR because I think should be another screen, because we can edit the issue title and the title description, so I think this could be done on the same screen.

src/api/index.js
```diff
@@ -220,15 +244,6 @@ export const fetchChangeIssueLockStatus = (
...
-export const fetchSearch = (type, query, accessToken, params = '') =>
-  v3.getJson(`/search/${type}?q=${query}${params}`, accessToken);
-
-export const fetchNotifications = (participating, all, accessToken) =>
-  v3.getJson(
-    `/notifications?participating=${participating}&all=${all}`,
-    accessToken
-  );
-
 export const fetchMarkNotificationAsRead = (notificationID, accessToken) =>
```
machour: @housseindjirdeh these were deleted by mistake ⚠️

PR #328, opened 2017-09-13
feat(repository): Show issues only when is enable

### LINKED ISSUE #296: Don’t show issues when issues are disabled for the repo
Some repos don’t have issues enabled but GitPoint shows an “new issue” button. This should be hidden when issues are disabled for the repo.

jouderianjr:
Closes #296 

| Before  | After |
| ------------- | ------------- |
|  |  |

### DIFF
--- a/src/repository/screens/repository.screen.js
+++ b/src/repository/screens/repository.screen.js
@@ -274,25 +274,27 @@ class Repository extends Component {
           navigateBack
         >
           {initalRepository &&
-            !initalRepository.owner &&
-            isPendingRepository &&
+          !initalRepository.owner &&
+          isPendingRepository && (
             <SectionList
               title={translate('repository.main.ownerTitle', language)}
             >
               <LoadingUserListItem />
-            </SectionList>}
+            </SectionList>
+          )}
 
           {!(initalRepository && initalRepository.owner) &&
-            (repository && repository.owner) &&
-            !isPendingRepository &&
+          (repository && repository.owner) &&
+          !isPendingRepository && (
             <SectionList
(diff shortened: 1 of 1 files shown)


src/repository/screens/repository.screen.js
```diff
@@ -353,20 +358,25 @@ class Repository extends Component {
           </SectionList>
 
           {!repository.fork &&
+          repository.has_issues && (
```
jouderianjr: To be honest, that is the only real change, all other changes were made by the pre-commit hook.

housseindjirdeh: Yep no worries, we're using prettier (https://prettier.io/) and attached it as a precommit step :)

Not sure this is the solution to the issue however. As a user, I still want to be able ot add a new issue when there are no issues present and removing the section entirely won't help that. I think the issue was referring to removing the `New Issue` button only if adding new issues aren't enabled rather than the entire SectionList (and I'm not sure the `has_issues` attribute is for this). Does that make sense mate?

jouderianjr: Cool, so I didn't destroy anything 😄 

So I've tested in one repository and on the github site the whole section is removed when we disable the issue option, So I thought that could be better just remove the section in the app as well.

Ps. This has_issue field is a flag to describe if the repository has issue feature or not.

https://developer.github.com/v3/repos/

screen shot 2017-09-12 at 23 24 18

housseindjirdeh: Oooooh my apologies you're right, and here I thought it meant if was a boolean that that returned based on `issues.length` :P

Thanks my man, this is awesome thank you for this 👏

PR #330, opened 2017-09-13
chore(dev): add redux-devtools-extension

lex111:
For better Redux debugging and for React Native Debugger application work https://github.com/jhen0409/react-native-debugger

### DIFF
--- a/package.json
+++ b/package.json
@@ -76,7 +76,8 @@
     "redux-persist": "^4.3.1",
     "redux-persist-transform-encrypt": "^1.0.2",
     "redux-thunk": "^2.2.0",
-    "reselect": "^3.0.1"
+    "reselect": "^3.0.1",
+    "redux-devtools-extension": "^2.13.2"
   },
   "devDependencies": {
     "@commitlint/cli": "^3.1.0",
--- a/root.store.js
+++ b/root.store.js
@@ -3,6 +3,7 @@ import { autoRehydrate } from 'redux-persist';
 import Reactotron from 'reactotron-react-native'; // eslint-disable-line import/no-extraneous-dependencies
 import createLogger from 'redux-logger';
 import reduxThunk from 'redux-thunk';
+import { composeWithDevTools } from 'redux-devtools-extension';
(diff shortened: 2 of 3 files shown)


machour:
> I was blind, but now I see.

Really nice! 👍 👍 👍

housseindjirdeh:
Definitely a fan of the dev-tools, have used it quite a bit in the past few weeks for React web.

Just a few concerns. We still have logger _and_ reactotron in the app. Are we including too many logging middleware and can potentially limit it to 1? Also will this boot up the dev tools every single time we run the simulator?

lex111:
@housseindjirdeh actually, no, I think. Reactotron is different, it differs from React Native Debugger, and both of these applications do not work together, so `composeWithDevTools` is not added when Reactotron is enabled.
I can, like the logger and Reactotron, add an environment variable, such as `devtools`, i.e. `start: android: devtools`, then `redux-devtools-extension` will be run, so do?

lex111:
@housseindjirdeh what do we do, can we merge already?

package.json
```diff
@@ -76,7 +76,8 @@
     "redux-persist": "^4.3.1",
     "redux-persist-transform-encrypt": "^1.0.2",
     "redux-thunk": "^2.2.0",
-    "reselect": "^3.0.1"
+    "reselect": "^3.0.1",
+    "redux-devtools-extension": "^2.13.2"
```
housseindjirdeh: Can we have this as a `devDependency`?

lex111: Hmm, actually I looked at other repositories, and there the dependency `redux-devtools-extension` is in just `dependencies`. Given that this only works when the debugging mode is enabled, then I understand that there is no point in adding a script command, I may be mistaken, but as I understand it, this is so.

housseindjirdeh: Perfect no worries, if you don't think it makes sense to add a script then that's perfectly fine :)

Yeah and I've noticed the same with `redux-logger` and I'm still no sure why those two usually live as dependencies :thinking: Let's circle back to that after some time.

Issue #318, opened 2017-09-09
Incorrect contributors when trying to @ in issues


chinesedfan:
Because `issue.screen.js` uses `state.repository`, if visit issues from home or notifications tabs, incorrect contributors are shown.

housseindjirdeh:
Hmm doesn't a call fire to update `state.repository` when navigating through home or notifications though? :thinking:

chinesedfan:
Closed by #332.

PR #332, opened 2017-09-15
fix(issue): Make sure the right repository is fetched

### LINKED ISSUE #315: Wrong issue/pr permissions when visit from notifications tab
I can always reappear this bug on my iPhone, with GitPoint 1.2 installed. But it is hard to debug, because there is no notifications on my simulator.

The operation path is:

   1. visit any repository that you own admin/push permissions from home tab
   2. switch to notifications tab
   3. click any issue/pr that you DO NOT own admin/push
   4. the issue detail page is shown with 'Merge Pull Request' button and settings icon on the top right corner

Even though dragging down the page, it can not be refreshed.

chinesedfan:
Fixed #315 and #318.

In fact, #318 is fixed partially by #317, as `getContributorsByDispatch` is not called correctly. But the main problem is that the right repository is not always fetched.

If visit `issue.screen` from:
- notifications, then `issueURLParam` has value, while `issueParam` is undefined (#315)
- home events, then `issueParam` has value, while `issueURLParam` is undefined
   - in addition, for special home events, i.e. close pr, `issueParam.repository_url` is undefined (#318)

So I use `props.issue.repository_url` to check, which belongs to the fetched issue.


src/issue/screens/issue.screen.js
```diff
@@ -160,10 +159,11 @@ class Issue extends Component {
         issueURLParam ? issueCommentsURL : issueParam.comments_url
       ),
     ]).then(() => {
+      const issue = this.props.issue;
```
lex111: @chinesedfan can as earlier receive variable `issue` (144 line)? Then you just do not need this line.

chinesedfan: @lex111 No, we should read after those 2 promises, `getIssueFromUrlByDispatch` and `getIssueCommentsByDispatch`, are fulfilled. Or `this.props.issue` may refer to the previous wrong issue.

That's what I mean "the fetched issue". Sorry for my poor English pr description.

housseindjirdeh: Nope that shouldn't be the case :)

In the top of the file, we're mapping parts of our state object into our props. So `issue` isn't a fixed object parameter here, but will always reflect what's happening with state. If `getIssueFromUrlByDispatch` and/or `getIssueCommentsByDispatch` update `state.issue` where state is our Redux container - our issue props will always reflect this. So we should be good with leaving it as part of the destructuring in line 144 😃 

Please let me know to clarify this if it's still not making any sense to you mate

src/issue/screens/issue.screen.js
```diff
@@ -160,10 +159,11 @@ class Issue extends Component {
...
+
       if (
-        issueParam &&
         repository.full_name !==
-          issueParam.repository_url.replace(`${apiRoot}/repos/`, '')
```
housseindjirdeh: So this was the culprit. It looks like `issueParam`, unlike `props.issue`, is actually obtained from the navigation.params so it's _not_ being updated when those calls within `Promise.all` are fired

src/issue/screens/issue.screen.js
chinesedfan: Right. `this.props.issue` will always be updated by redux, but in line 144, `issue` is just a temporary local variable destructed from `this.props`, which will never be updated.

src/issue/screens/issue.screen.js
chinesedfan: Suppose we have visited repository A, and try to visit another issue/pr of repository B. `props.repository` will refer to A at first. `issueParam.repository_url` will refer to B.

The basic logic is correct here. But `issueParam.repository_url` does not always have value, which may be caused by `issueParam` itself or `issueParam.repository_url` as my PR description says.

chinesedfan: @housseindjirdeh Any feedbacks? :)

src/issue/screens/issue.screen.js
housseindjirdeh: Oh I'm sorry, you're right here. It gets modified in the promise call and our current local variable isn't really going to reflect this.

Cheers thank you mate, and I'm sorry for causing any confusion 😞

PR #333, opened 2017-09-15
feat(screen): add support for landscape orientation

lex111:
I do not know, maybe it was so necessary, but I added support for landscape orientation. In this regard, we had to slightly change the parallax component , but everything seems to be fine.

screenshot_1505503493


lex111:
@housseindjirdeh oh, oops, this is my experiment. I thought, to output the number of unread notifications in such a way that you think about it? Do the icons for notifications do this.

housseindjirdeh:
I think that's an amazing idea :) One thing that would also be nice is potentially show the unread notifications in a slightly different way (maybe with a gray background for example) when the user navigates to that tab. But I think that's great :D 

However will definitely like to get other people's opinions on this as well. Will open an issue later to discuss :tada:

PR #334, opened 2017-09-15
chore(npm): add clean scripts

lex111:


package.json
```diff
@@ -12,6 +12,8 @@
     "start:ios": "concurrently -r 'react-native start' 'react-native run-ios'",
     "start:ios:logger": "LOGGER_ENABLED=true concurrently -r 'react-native start --reset-cache' 'react-native run-ios'",
     "start:ios:tron": "TRON_ENABLED=true concurrently -r 'react-native start --reset-cache' 'react-native run-ios'",
+    "clean": "rm -rf node_modules",
+    "clean:android": "cd android && ./gradlew clean && cd -",
```
housseindjirdeh: Do you think `clean:android` should remove node modules as well as this or do you prefer having them separate like this? Okay with either way myself.

lex111: I prefer to split them, I sometimes have an error when building the application and helps to clean up via `gradlew`.

housseindjirdeh: Sounds good :)

PR #336, opened 2017-09-17
fix(ux): Fix user stars and repository watching status

chinesedfan:
Fixes 2 small bugs while loading:
- The label "stars" in user profile is not kept in the same line with other labels.
- The "watching" badge in repository profile is not reset if the previous repository is watching.

### DIFF
--- a/src/components/user-profile.component.js
+++ b/src/components/user-profile.component.js
@@ -146,7 +146,7 @@ export const UserProfile = ({
       {type !== 'org' &&
         <TouchableOpacity style={styles.unit}>
           <Text style={styles.unitNumber}>
-            {starCount}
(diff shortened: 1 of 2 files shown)


jouderianjr:
Hey @chinesedfan, I was looking the app code and I think this error is caused because in `user.reducer.js` the `starCount` field is not in initialState, so maybe it is easier just add the starCount in initialState object.

What do you think about this approach?

chinesedfan:
@jouderianjr Thanks for your advice! It is really another way to fix.

I do not like `parseInt`, either. But the nearby `public_repos/followers/following` uses this kind of method. Maybe someone can send another pr to refactor them later.

PR #338, opened 2017-09-19
feat(navigator): render tabs as needed

lex111:
There is a [possibility](https://github.com/react-community/react-navigation/blob/master/docs/api/navigators/TabNavigator.md#tabnavigatorconfig) to render tabs as necessary, this would be useful to us, because now the initial load takes a long time because there are many requests (notifications, profile).

However, I noticed one unpleasant thing for me: when lazy loading of tab, you can see a gray background for a few seconds, I can not understand where it comes from and whether it can be replaced with a loading indicator. Would be glad if someone else looked this up!

### DIFF
--- a/routes.js
+++ b/routes.js
@@ -370,5 +370,8 @@ export const GitPoint = StackNavigator(
   {
     headerMode: 'screen',
     URIPrefix: 'gitpoint://',
-  }
+    cardStyle: {
+      backgroundColor: 'transparent',
+    },
+  },
 );
--- a/routes.js
+++ b/routes.js
@@ -302,6 +302,7 @@ const MainTabNavigator = TabNavigator(
     },
   },
   {
+    lazy: true,
     tabBarPosition: 'bottom',
     tabBarOptions: {
       showLabel: false,


machour:
Hey @lex111, did some digging in react-navigation, here's where this color is set:
  https://github.com/react-community/react-navigation/blob/master/src/views/CardStack/Card.js#L41

(tested it by hardcoding another color in node_modules/). Now you just need to figure out how to pass the `style` prop ;)

lex111:
@machour Thank you, it's works! I set a transparent (white) background, it looks better now, but the very fact of such a delay is unpleasant, and apparently nothing can be done. But on the other hand, the application is faster loaded, but what is your opinion, do need a lazy load of tabs?

jouderianjr:
Hey, we could try a new approach for reduce this initial loading. We could start to fetch the data just when the user goes to the tab for the first time.

machour:
@lex111 tested on my phone, it seems quiet buggy.. maybe because I'm in dev mode.

I second @jouderianjr idea (if I got it right): creating our own lazy loading may be better: on start, only fire the requests used for the first screen. Once displayed to the user, fire the remaining requests for the other tabs.

jouderianjr:
@machour almost that. The difference is we could fire the requests only if the user goes to the specific tab. For example: after start the user goes to profile tab, then we will fire the necessary requests for this tab.

This way basically we could do request when is necessary. 

Has react-navigation event for tab change?

machour:
@jouderianjr as far as I understand, you're describing what @lex111 implemented using `lazy: true`.
Using this setting, the tab is only processed/rendered when the user navigate to it.

The problem he's facing is that when the user clicks, he's presented with an empty screen while the requests are being executed, without the possibility of displaying a spinner or whatever during that wait.

lex111:
Yes, in fact it remains to implement its own lazy loading, but I did not find the events for tabs, and even if it is, all the same without turning on the `lazy` param, the current problem will remain.

@machour it is possible more in detail, what does not work?

jouderianjr:
@machour is the same of lazy: true but in "our side", this way we could manipulate the UI. Tonight I will do some research about it.

jouderianjr:
@lex111 and @machour I did a POC for showing how is possible do the lazy requests on our side. Could you take a look on that guys? 😄 

https://github.com/jouderianjr/git-point/tree/poc_lazy_requests

Ps.The code is a bit ugly, I didn't polish much because I would like your opinions about that.

lex111:
@jouderianjr thanks a lot, this is a working solution, although now many actions need to be done manually, I would like to avoid hardcode.

And an important note in [App.js]( https://github.com/gitpoint/git-point/blob/master/App.js#L62) is to add key `blacklist: ['lazyTabLoading']` so that the `lazyTabLoading` state will not be saved, otherwise on reload application `firstTimeOnNotificationScreen` and `firstTimeOnProfileScreen` will be `true`.

If to rework, then I think we can use our own lazy loading of tabs, this will well affect the performance of the application. @housseindjirdeh @andrewda what do you think?

housseindjirdeh:
This is awesome folks, thank you all so much. I had no idea tabs could be lazy loaded so easily with React Navigation.

I'm not seeing any problem with a gray background or anything, it looks like it just fires the actions when the user opens a tab:

lazyload

@lex111 Are you seeing issues with this on Android?

@machour yeah things will be _slow_ booting up when running on dev mode (and especially if you have logger or any other middleware running :disappointed:)

andrewda:
@housseindjirdeh (sorry for spamming your notification feed 😞 )

Anyways, this looks super cool so far! Thanks @lex111!

lex111:
@housseindjirdeh hmm, strange, I have a gray background displayed. And can you look at the Android-simulator?

lex111:
Looked at the release version on real device, the delay in switching to screen smaller, so I suggest to merge the changes.

PR #339, opened 2017-09-19
feat(user-repositories): load more repositories when scrolling

lex111:
Now only 50 repositories are loaded, although the user may have more. I want to load new repositories as I scroll, now it works, but there is one big problem, the store keys `page`, `hasMoreRepositories` and `lastPage`  are not cleared when repositories list are reopened, and moreover, when the application is reloaded, in the store, these keys will be populated with values, those (cache values, I mean). Need to somehow clean them, but I do not know how to do it right.

Perhaps I incorrectly initially made the implementation of loading new data when scrolling. Therefore, I want someone to help me with this problem, because then in plans everywhere to make loading of the new data at scrolling, i.e. not just repositories.


lex111:
It seems I figured it out a little: I added the `user` key to redux-persist black list , i.e. when the application is loaded, `user` key will no longer be filled with the data, as it was before. That is in general, but it still has a problem with the fact that you need to clear the `page`, `lastPage` keys after all the data has been loaded, otherwise it will not be possible to re-loaded them.
In general, now added three new keys to the `user` store key, which I do not really like, maybe I need to do something differently, given that such a load (pagination) we will do in the rest of the screens? @housseindjirdeh  help is needed :hand:

src/user/screens/repository-list.screen.js
```diff
@@ -159,6 +170,11 @@ class RepositoryList extends Component {
                     repository={item}
                     navigation={navigation}
                   />}
+                onEndReached={
+                  !hasMoreRepositories
+                    ? () => null
+                    : () => getMoreRepositoriesByDispatch(user, page + 1)
```
alejandronanez: Do you think it would be possible to remove arrow function inside this prop?

machour:
This will be soon implemented in #751. Closing this stale PR.

Issue #275, opened 2017-08-23
Add support for Android O's Adaptive Icons


JacobDB:
I'd love it if you could add Adaptive Icon support to this app; it's very easy to implement, [here's the documentation](https://developer.android.com/guide/practices/ui_guidelines/icon_design_adaptive.html). I could try submitting a pull request with this change, but I'm a web developer, and I've never worked with Android, so I can't guarantee quality.

housseindjirdeh:
Didn't even know what adaptive icons were until you submitted this issue, so thank you :)

I think this will be awesome to have, it really doesn't look too complicated at all so definitely something I think we can have!!

https://developer.android.com/guide/practices/ui_guidelines/icon_design_adaptive.html#creating_adaptive_icons_in_xml

machour:
Closed in #340, thanks a lot for the suggestion!

PR #340, opened 2017-09-19
feat(android): Implement round and adaptive icon

### LINKED ISSUE #297: Round the corners of the android app icon
I use GitPoint on an Android device (but mainly iOS). I don’t like how the icon is squared (meaning not transparent anywhere). Most people prefer non-squared icons as well. So rounding the icon will make it look nicer.

machour:
In this PR:

- Updated the original squared icon to a squircle one
- Added `roundIcon` for devices supporting it since 7.1
- Added an adaptive icon in `mipmap-anydpi-v26`
- Updated compile/buildTools/target SDK versions

Since we're using `roundIcon` now, I was forced to duplicate the Android-O icon XML file,
 or else Android O kept using the PNGs provided in roundIcon.

Fixes #297 and #275


android/app/build.gradle
```diff
@@ -92,13 +92,13 @@ def enableSeparateBuildPerCPUArchitecture = false
 def enableProguardInReleaseBuilds = false
 
 android {
-    compileSdkVersion 23
-    buildToolsVersion "23.0.1"
+    compileSdkVersion 26
+    buildToolsVersion '25.0.3'
```
lex111: Can the same up to 26 versions raise? I do not know for sure, just wondering.

machour: oopsy, updated

machour:
Here's what the progressive icon looks like among other icons in Android O:

screen shot 2017-09-19 at 7 57 13 pm

@housseindjirdeh do we give it more margin to make it look like the rounded icon on https://gitpoint.co/? or do we do the opposite? 🤔

android/app/build.gradle
```diff
@@ -92,13 +92,13 @@ def enableSeparateBuildPerCPUArchitecture = false
 def enableProguardInReleaseBuilds = false
 
 android {
-    compileSdkVersion 23
-    buildToolsVersion "23.0.1"
+    compileSdkVersion 26
+    buildToolsVersion '26.0.1'
```
lex111: I'll be pedantic, but let's change the quotes to double, as it used to be? :smile:

machour: ** curses Android Studio **

lex111: Heh, thank you :+1:

PR #341, opened 2017-09-19
feat(user): add starred repositories list

nersoh:
Show *starred repositories* list when clicking on start count number on user profile.

I think some improvements should be done like showing a loading indicator when fetching more repos, so any feedback would be appreciated.


src/components/user-profile.component.js
```diff
@@ -144,7 +144,15 @@ export const UserProfile = ({
...
 
       {type !== 'org' &&
-        <TouchableOpacity style={styles.unit}>
+        <TouchableOpacity
+          style={styles.unit}
+          onPress={() =>
+            navigation.navigate('StarredRepositoryList', {
+              title: 'Starred Repositories',
+              user,
+              repoCount: starCount > 15 ? 15 : starCount,
```
machour: 15 should be moved to a const

src/locale/languages/ru.js
```diff
@@ -4,18 +4,22 @@ export const ru = {
       welcomeTitle: 'Добро пожаловать в GitPoint',
       welcomeMessage: 'Самый многофункциональный бесплатный GitHub клиент',
       notificationsTitle: 'Управление уведомлениями',
-      notificationsMessage: 'Просматривайте и управляйте всеми вашими непрочитанными и активными уведомлениями',
```
machour: Was there any changes in the `ru.js` file besides the formatting?
If not, it shouldn't have been part of this PR

src/repository/screens/repository.screen.js
```diff
@@ -275,44 +275,40 @@ class Repository extends Component {
           navigateBack
         >
           {initalRepository &&
-          !initalRepository.owner &&
-          isPendingRepository && (
+            !initalRepository.owner &&
```
machour: There is a lot of whitespaces changes in your PR that makes it really hard to read :(

(advice: I personally always perform a `git diff` on my changes before the commit and suppress any unwanted change in whitespaces. Helps me make sure my diff is minimalist and right to the PR point)

src/repository/screens/repository.screen.js
```diff
@@ -429,15 +420,15 @@ class Repository extends Component {
...
                   key={item.id}
                   type="pull"
                   issue={item}
                   navigation={navigation}
                   language={language}
                 />
-              ))}
+              )}
           </SectionList>
         </ParallaxScroll>
```
machour: it seems that this file also shouldn't be part of the PR, only formatting changes

lex111:
The implementation of loading new data when scrolling should be slightly different. I tried to do this today for a list of repositories, you can see in #339 However, there is a problem with the redux store. Perhaps you can give some improvements, any help is welcome! Although I now see that I could save the last page in the component state, not the redux state...

It is necessary to receive and save the last page and to define on the basis of it it is necessary to load still the data or not. Now you do not have one.

nersoh:
@lex111 I just saw your PR and realized we wrote a pretty similar code. I thought about your approach of taking page numbers from response headers, but I ended up not doing it even thought I should.

I'll work on that _last page_ state_ trying to follow your code and feedback. 😄 

Thanks for your feedback and thoughts.

src/locale/languages/ru.js
lex111: There simply made changes to the formatting, @nersoh you need to remove this change from PR.

machour:
@nersoh @lex111 thinking out loud: would it be possible to create a <RepositoriesList /> component that given a URL handle the pagination by itself? 

We would code all the logic of repositories pagination in one place and we could use it in the following places:
* Repositories search result
* User / Organisation repositories
* Starred repositories
* Forked repositories

Is that feasible?

src/repository/screens/repository.screen.js
lex111: This is probably because the code passes through the linter, and before this the code was not formalized according to the rules, so now we see so many changes. In general, nothing to do with this, except that do not create a new PR where all the code will be formatted according to the standards.

lex111:
@machour BTW, yes, a good idea. After all, the same actions are taken, we need to think about how this can be done with minimal costs. But the main thing is to determine the correct algorithm for obtaining data. By the way, could not you take a look at my PR #339?

nersoh:
@machour That's definitely a good point. I think that if we want to implement pagination on all those repositories pages, that would be a way to do. 👌

src/repository/screens/repository.screen.js
machour: if the file isn't changed, nor included in the commit, you won't get blocked

machour:
@lex111 I need to understand Redux before being able to help. Started a tutorial now 😄

src/components/user-profile.component.js
nersoh: @machour I noticed that there are a few not const numbers in this file such as _15_, _10_ etc. Maybe I could set all of them as const like **maxPublicRepos**, **maxStarCount**, **maxFollowers** etc. What do you think?

nersoh:
@machour After testing and debugging things, I analyzed that problem about going back and forth a profile page and a request being done to a 'wrong' user. That issue is related to those problems we've going through with redux shared data, when we open a profile screen both on Home and Profile tabs.

I don't see how I can fix this for now, but if you have any thoughts on that would be great. 😄

machour:
@nersoh unfortunately this can't be fixed without re-organizing the store :(

housseindjirdeh:
Yeah it's a serious issue @nersoh and we're going to have to rearchitect things in our store to ensure this doesn't happen :O

housseindjirdeh:
Yeah it's a serious issue @nersoh and we're going to have to rearchitect things in our store to ensure this doesn't happen :O

housseindjirdeh:
Friendly ping @nersoh, let us know if you need help with this and happy to take over and apply the final requested touched myself :)

nersoh:
Here's some change I've made:

- Remove `contraints.js` file and move its code to the `starred-repositories-list.screen` itself.
- Add the same behaviour as `repository-list.screen` (loading, styles etc).
- Remove loading of more starred repositories. As a result, starred repositories screen loads only 50 repos, the same amount as user repos screen. I decided not to add more complexity and keep things simple. That's because we've been having issues with our redux store and its state shared through all screens, so a scenario of **user A** loading more repositories of **user B** would often happen as @machour pointed.
- Do not add **SearchBar** to starred repositories since I did find any github endpoint to do it. If any of you guys know some about it, please let me know.
- @lex111 has also started to do [#339](https://github.com/gitpoint/git-point/pull/339) pretty much what I was doing about getting more repositories. So I thought we could later take his code to add components like _infinite-scroll-repositories-list_ as @housseindjirdeh mentioned.

What do you guys think?

dump.rdb
```diff
@@ -0,0 +1 @@
+REDIS0006�ܳC�Z��V
```
chinesedfan: What is this file? Maybe added by mistake.

routes.js
```diff
@@ -373,5 +380,5 @@ export const GitPoint = StackNavigator(
     cardStyle: {
       backgroundColor: 'transparent',
     },
-  },
+  }
```
chinesedfan: We always keep the tail comma.

src/components/user-profile.component.js
```diff
@@ -130,7 +137,10 @@ export const UserProfile = ({
           navigation.navigate('RepositoryList', {
             title: translate('user.repositoryList.title', language),
             user,
-            repoCount: user.public_repos > 15 ? 15 : user.public_repos,
+            repoCount:
+              user.public_repos > maxLoadingConstraints.maxPublicRepos
+                ? maxLoadingConstraints.maxPublicRepos
+                : user.public_repos,
```
chinesedfan: How about `Math.min`? There are 4 similar places in this file.

src/user/screens/starred-repository-list.screen.js
```diff
@@ -0,0 +1,66 @@
...
+const mapStateToProps = state => ({
+  user: state.user.user,
+  starredRepositories: state.user.starredRepositories,
+  isPendingStarredRepositories: state.user.isPendingStarredRepositories,
+});
+
+const mapDispatchToProps = dispatch => ({
+  getStarredRepositories: (user, type) =>
+    dispatch(getStarredRepositories(user, type)),
+});
```
chinesedfan: Now we prefer to use `bindActionCreators`.

src/user/index.js
```diff
@@ -1,5 +1,4 @@
 export * from './user.action';
 export * from './user.reducer';
 export * from './user.type';
-
```
chinesedfan: If it does not really need, maybe we can revert this line.

src/components/user-profile.component.js
nersoh: Sweet!

src/components/user-profile.component.js
```diff
@@ -144,7 +154,15 @@ export const UserProfile = ({
       </TouchableOpacity>
 
       {type !== 'org' &&
-        <TouchableOpacity style={styles.unit}>
+        <TouchableOpacity
+          style={styles.unit}
+          onPress={() =>
+            navigation.navigate('StarredRepositoryList', {
+              title: translate('common.stars', language),
```
lex111: @nersoh please do by analogy as [here](https://github.com/gitpoint/git-point/blob/master/src/locale/languages/en.js#L163), i.e. will be like this:

```js
stars: {
  title: 'Stars', // used on 161 line
  text: 'Stars' // used on 170 line
}
```
This separation is necessary for languages where declination is used. In English there will always be "Stars" in both the `title` and `text`.

src/user/user.action.js
```diff
@@ -201,6 +202,30 @@ export const getRepositories = user => {
   };
 };
 
+export const getStarredRepositories = user => {
+  return (dispatch, getState) => {
+    const url = `/users/${user.login}/starred?&sort=updated&per_page=50`;
```
lex111: Let's remove the sorting by `updated` (remove `sort=updated`), then the default sorting will work according to the date of creation (`created`), it is this kind of sorting that is used on GitHub right now.

src/user/index.js
lex111: :point_up:

routes.js
lex111: :point_up:

nersoh:
@lex111 I added some changes. Please, let me know if I missed anything.

src/user/screens/starred-repository-list.screen.js
```diff
@@ -0,0 +1,65 @@
...
+      starredRepositories,
+      navigation,
+      isPendingStarredRepositories,
+    } = this.props;
+
+    return (
+      <ViewContainer>
+        {isPendingStarredRepositories &&
+          [...Array(starredRepositories.length)].map(
+            (item, index) => <LoadingRepositoryListItem key={index} /> // eslint-disable-line react/no-array-index-key
```
housseindjirdeh: Can we use the repository ID here? Will save us from having to use `index` as well as having the lint check disabled

chinesedfan: @housseindjirdeh No repository IDs here. Because they are just loading placeholders, which should be similar with [repository-list.screen.js](https://github.com/gitpoint/git-point/blob/master/src/user/screens/repository-list.screen.js#L153). And the loop target is wrong, it should be `[...Array(repoCount)]`, instead of `[...Array(starredRepositories.length)]`.

chinesedfan:
@nersoh I did some changes directly on your branch, in order to merge it.

nersoh:
No problem @housseindjirdeh and @chinesedfan. That's awesome. 👏

PR #347, opened 2017-09-20
refactor(orgs): get organizations via GraphQL and display name instead login

### LINKED ISSUE #272: Use organisation Display name instead of organization name
Hey guys,

it would look a bit better if we Display the Displayname of an Organization insted of their "username"

f.e. at Adobe:
`adobe`
but:
`Adobe Systems Incorporated`
Is their Display Name.

Looks better doenst it?

I would take care of this if you guys want this to be improved too.

lex111:
There are currently two requests when receiving organizations: one requests public organizations, another request also requests organizations with private membership. Using GraphQL, we can get all organizations with one query. However, doing so required updating the scope of the token. We need to somehow come up with a way to update the token, how to do it better, who has any ideas?

Also, within the framework of this PR done that, the organization name is output instead of the login, this closes #272


src/api/index.js
```diff
@@ -153,13 +166,39 @@ export async function fetchUser(user, accessToken) {
...
+    accessTokenParametersGraphQL(accessToken, getOrganizationsQuery, {
+      login: user,
+    })
   );
 
-  return response.json();
+  return response.json().then(data => {
+    const orgs = [];
+
+    data.data.user.organizations.edges.reduce((key, item) =>
```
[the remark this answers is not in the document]
lex111: @jouderianjr I completely agree, thank you, I will fix it.

lex111:
There remains such a problem: we need to open a link to reauthorize, in order for the user to update the token. How is this best done?

src/api/index.js
```diff
@@ -153,13 +166,37 @@ export async function fetchUser(user, accessToken) {
 }
 
 export async function fetchUserOrgs(user, accessToken) {
-  const ORGS_ENDPOINT = `${root}/users/${user}/orgs`;
+  const getOrganizationsQuery = `
```
machour: I'm a bit scared of seeing `api/index.js` grow too much. Would it be possible to put queries in entities subfolders ? This one for example would go in `user/user.queries.js`

src/components/user-list-item.component.js
```diff
@@ -131,7 +133,7 @@ class UserListItemComponent extends Component {
               <Image
                 style={styles.avatar}
                 source={{
-                  uri: user.avatar_url,
+                  uri: imageUrl || user.avatar_url,
```
machour: I think it's best to make the `imageUrl` prop mandatory and pass it as { user.avatar_url } where needed

lex111: I do not think that it will turn out, i.e. when querying GraphQL, another notation is used (lowerCase instead of snake_case). We have already done this, [here](https://github.com/gitpoint/git-point/blob/fe77ed05a87f6221c468977351fa0336ad780c43/src/components/user-list-item.component.js#L143)

src/api/index.js
lex111: I agree :+1:

src/components/user-list-item.component.js
machour: I still think that having an optional prop, magically provided by another prop in some cases, is a bit confusing. I'm in favor of "real dumb" presentation components.

Also in the spirit of #350 (did you check it?), I think GraphQL responses should be stored using the same schemas as REST responses in the entities array. That would enable us to access the data in the same way, regardless of where it came from (rest or graphql). Am I making sense? :/

src/auth/screens/login.screen.js
```diff
@@ -153,7 +153,7 @@ class Login extends Component {
 
   signIn = () =>
     openURLInView(
-      `https://github.com/login/oauth/authorize?response_type=token&client_id=${CLIENT_ID}&redirect_uri=gitpoint://welcome&scope=user%20repo&state=${stateRandom}`
+      `https://github.com/login/oauth/authorize?response_type=token&client_id=${CLIENT_ID}&redirect_uri=gitpoint://welcome&scope=user%20repo%20read:org&state=${stateRandom}`
```
housseindjirdeh: Looks like this adds read access to org data. Without this, does your queries not work?

src/api/index.js
housseindjirdeh: _Highly agree_, I made the mistake of laying things all down in `api/index.js` but it really makes sense to leverage feature api files (like we do with other parts of the app).

src/api/index.js
```diff
@@ -153,13 +166,37 @@ export async function fetchUser(user, accessToken) {
...
+    query($login: String!) {
+      user(login: $login) {
+          organizations(first: 100) {
+            edges {
+              node {
+                id
+                login
+                name
+                avatarUrl
```
housseindjirdeh: Seconding @machour's comment. I still don't know how exactly GraphQL works, but can we not stick to the same mention of `avatar_url` or does this version of the API just return it as `avatarUrl` instead of `avatar_url`?

src/auth/screens/login.screen.js
lex111: Yes, without this, not getting an organization, there will be an error. Therefore, I write that there is a problem: we need to update the token when changing the scope, as for example in this case. And the question: how to do this?

src/api/index.js
lex111: That's just the point, that the GraphQL query returns the keys in another notation

src/components/user-list-item.component.js
lex111: I have not watched, but probably this will solve the problem, it seems to me...

src/api/index.js
alejandronanez: What do you think about moving this query into something like

`/src/queries/foo.query.js` or something like that?

lex111: Maybe it's better this way `/src/api/queries/foo.query.js`?

machour: What about /src/user/user.queries.js ? (just like we'll have user.api.js) ?

machour:
@lex111 For your changed scope and the need to re-authorize, I think this call should help: 
   https://developer.github.com/v3/oauth_authorizations/#check-an-authorization
You can see with this call if your needed permission is granted, or if you should ask the user to grant it

src/api/index.js
alejandronanez: FWIW I think you can make an `alias` to what you get from GQL. I'd prefer camelCase instead of snake_case though :P

src/auth/screens/login.screen.js
```diff
@@ -249,7 +249,7 @@ class Login extends Component {
                 <View style={styles.browserSection}>
                   <WebView
                     source={{
-                      uri: `https://github.com/login/oauth/authorize?response_type=token&client_id=${CLIENT_ID}&redirect_uri=gitpoint://welcome&scope=user%20repo&state=${stateRandom}`,
+                      uri: `https://github.com/login/oauth/authorize?response_type=token&client_id=${CLIENT_ID}&redirect_uri=gitpoint://welcome&scope=user%20repo%20read:org&state=${stateRandom}`,
```
alejandronanez: Maybe for later but I think we can move this `url` into a _constants_ files.

machour:
Closing this PR as it's superseded by the ongoing store refactoring. #272 remains open and will be dealt with.

PR #350, opened 2017-09-23
Experimenting with normalizr

machour:
This is not a real PR, but more a first experiment conducted using Redux with Normalizr, and some of my findings. Maybe this could lead to good ideas for the big refactoring due on our side to better handle the state.

As a reminder, our current Store problem is that we only have a user / repo in the state at runtime, and if you open two different profiles on two different screens, last data fetched is shown on both.

I came across https://github.com/reactjs/redux/tree/master/examples/real-world a few days ago, and decided to try to implement something similar in GitPoint. 

The main idea is to have a Redux middleware taking care of most of the APIs calls, and normalizing the stored data into entities and paginations.

Here's what is working and worth looking at in this PR:

- Normalization of User/Schema/Event
- Pagination of the list of followers and repositories for a user
- Minimizing the store footprint by excluding some predictable fields from the entity

I'm going to comment some files changes in the diff to give some more insight.

### DIFF
--- a/src/repository/repository.schema.js
+++ b/src/repository/repository.schema.js
@@ -6,19 +6,7 @@ import { userSchema } from '../user/user.schema';
 /*
 export const userSchema = new schema.Entity('users', {}, {
     idAttribute: user => user.login.toLowerCase(),
-    processStrategy: entity => omit(entity, [
-        'url',
-        'html_url',
-        'followers_url',
-        'following_url',
-        'gists_url',
-        'starred_url',
-        'subscriptions_url',
-        'organizations_url',
-        'repos_url',
-        'events_url',
-        'received_events_url',
-    ]),
+
 });*/
 
 export const repoSchema = new schema.Entity(
@@ -35,5 +23,50 @@ export const repoSchema = new schema.Entity(
       // In Events
       return repo.name.toLowerCase();
     },
+    processStrategy: entity => omit(entity, [
+      'url',
+      'html_url',
+      'url',
+      'forks_url',
+      'keys_url',
+      'collaborators_url',
+      'teams_url',
+      'hooks_url',
+      'issue_events_url',
+      'events_url',
+      'assignees_url',
+      'git_url',
+      'ssh_url',
+      'clone_url',
+      'svn_url',
+      'branches_url',
+      'tags_url',
+      'blobs_url',
+      'git_tags_url',
+      'git_refs_url',
+      'trees_url',
+      'statuses_url',
+      'languages_url',
+      'stargazers_url',
+      'contributors_url',
+      'subscribers_url',
+      'subscription_url',
+      'commits_url',
+      'git_commits_url',
+      'comments_url',
+      'issue_comment_url',
+      'contents_url',
+      'compare_url',
+      'merges_url',
+      'archive_url',
+      'downloads_url',
+      'issues_url',
+      'pulls_url',
+      'milestones_url',
+      'notifications_url',
+      'labels_url',
+      'releases_url',
+      'deployments_url',
+    ]),
   }
 );


src/api/api.client.js
```diff
@@ -0,0 +1,482 @@
+import { normalize, schema } from 'normalizr';
```
machour: This is the old `api/index.js` file, with some additions in the end : The client api.middleware uses to query the REST api and paginate the results

src/api/api.schema.js
```diff
@@ -0,0 +1,13 @@
+import { userSchema } from '../user/user.schema';
+import { repoSchema } from '../repository/repository.schema';
+import { eventSchema } from '../event/event.schema';
+
+// Schemas for Github API responses.
+export const Schemas = {
```
machour: Schemas are used by Normalizr to understand the JSON structure

src/api/api.type.js
```diff
@@ -0,0 +1,10 @@
+export * from '../event/event.type';
```
machour: Making all types available in this file.

src/auth/screens/events.screen.js
```diff
@@ -380,7 +395,7 @@ class Events extends Component {
             style={styles.linkDescription}
             onPress={() => this.navigateToRepository(userEvent, true)}
           >
-            {userEvent.payload.forkee.full_name}
+            {userEvent.payload.forkee}
```
machour: After normalization, userEvent.payload.forkee is simply the repo name (which can be later found in entities.repo)

src/components/repository-list-item.component.js
```diff
@@ -108,5 +108,6 @@ export const RepositoryListItem = ({ repository, navigation }: Props) =>
       type: 'octicon',
     }}
     underlayColor={colors.greyLight}
-    onPress={() => navigation.navigate('Repository', { repository })}
+    onPress={() =>
+      navigation.navigate('Repository', { name: repository.full_name })}
```
machour: Most of our screens can be passed either a "string" or an "object" for the principal entity they're dealing with.
For example, user-profile can either take a `login` name, or a `user` object. This is really confusing and generate a lot of checks.

By switching to normalizer, we can now pass only the login, and let the middleware fetch the data from store or api :)

src/components/user-profile.component.js
```diff
@@ -6,7 +6,6 @@ import { ImageZoom } from 'components';
 
 type Props = {
   type: string,
-  initialUser: Object,
```
machour: Those changes are soo wrong, disregard this file

src/config/reactotron.js
```diff
@@ -3,5 +3,10 @@ import Reactotron from 'reactotron-react-native';
 import { reactotronRedux } from 'reactotron-redux';
 
 if (__DEV__ && process.env.TRON_ENABLED) {
-  Reactotron.configure().useReactNative().use(reactotronRedux()).connect();
+  Reactotron.configure({
```
machour: oupsy too.

src/event/event.action.js
```diff
@@ -0,0 +1,28 @@
+import has from 'lodash/has';
```
machour: This is the file handling all events action.
I find it handy to have all functions here, instead of cluttering api/api.client.js.

`_fetchEvents` could be inlined in `loadEvents` too, but if you notice later, a lot of function looks just like `loadEvents`, maybe we could make a unique function out of it.

src/event/event.schema.js
```diff
@@ -0,0 +1,14 @@
...
+import { repoSchema } from '../repository/repository.schema';
+
+export const eventSchema = new schema.Entity('events', {
+  actor: userSchema,
+  repo: repoSchema,
+  payload: {
+    forkee: repoSchema,
+    member: userSchema,
+  },
+});
```
machour: This is one of my favorite files :) 
We describe the event structure and link nested objects to their own schema

src/user/screens/follower-list.screen.js
```diff
@@ -56,13 +75,15 @@ class FollowerList extends Component {
                   navigation={navigation}
                   showFullName
                 />}
+              onEndReachedThreshold={0.5}
+              onEndReached={() => getFollowers(login, true)}
```
machour: this is how pagination is done :) Passing true as the second parameter for the data fetcher

src/user/screens/profile.screen.js
```diff
@@ -144,31 +173,33 @@ class Profile extends Component {
...
     const isPending = isPendingUser || isPendingOrgs;
     const userActions = [
       isFollowing
         ? translate('user.profile.unfollow', language)
         : translate('user.profile.follow', language),
     ];
 
+    const orgs = [];
+
+    if (!user) {
```
machour: Instead of doing this, we could draw the screen, display the information we already have (user avatar and name) and draw loading elements for the rest until the data is retrieved from the API.

src/user/screens/repository-list.screen.js
```diff
@@ -9,15 +9,27 @@ import {
...
+const loadData = ({ login, getRepos }) => {
+  getRepos(login);
+};
+
+const mapStateToProps = (state, ownProps) => {
+  const login = ownProps.navigation.state.params.user.login.toLowerCase();
+
+  const { entities: { repos }, pagination: { reposByUser } } = state;
+  const reposPagination = reposByUser[login] || { ids: [] };
+  const repositories = reposPagination.ids.map(id => repos[id]);
```
machour: Here's how we reconstruct the repositories array, we fetch the ids from the pagination object and map them back to entities.repos

src/user/user.schema.js
```diff
@@ -0,0 +1,24 @@
+import { schema } from 'normalizr';
+import omit from 'lodash/omit';
+
+export const userSchema = new schema.Entity(
+  'users',
+  {},
+  {
+    idAttribute: user => user.login.toLowerCase(),
+    processStrategy: entity =>
+      omit(entity, [
```
machour: All this fields can be guessed by having just the username. We remove them from the object to minimize our footprint

machour:
Finally, here's how to test this PR : 

- Load the App
- Click on a user name in the events screen
- Click on either his repos or followers
- From there, click either on a repo or a username

This is the only testable path so far.

Here's how the redux store looks like during that  : 

- Events loaded, here's what we have on our users: 
screen shot 2017-09-23 at 5 19 17 pm

- User clicked, profile page loaded:
screen shot 2017-09-23 at 5 19 32 pm

- Number of repositories clicked : 

screen shot 2017-09-23 at 5 20 15 pm

- Clicking on the `samdark/laradock` repo: 

screen shot 2017-09-23 at 5 27 52 pm

src/repository/repository.schema.js
```diff
@@ -0,0 +1,72 @@
...
+      'downloads_url',
+      'issues_url',
+      'pulls_url',
+      'milestones_url',
+      'notifications_url',
+      'labels_url',
+      'releases_url',
+      'deployments_url',
+    ]),
+  }
```
machour: Now this 
screen shot 2017-09-23 at 5 27 35 pm
Looks definitely better than this:
screen shot 2017-09-23 at 5 23 14 pm

machour:
Deprecated by #457

PR #356, opened 2017-09-25
refactor(parallax-scroll): add TouchableOpacity to buttons, remove unused styles

lex111:


housseindjirdeh:
Side note: Do you usually create new branches off the main repo? Any chance you can work off a fork of the repo instead if that's okay? (Not a big deal really, but noticing a lot of stale branches in the main:

image

lex111:
@housseindjirdeh I create my own branches in the main repository so that other contributors (if needed) can easily fetch it. I delete the branches after unification, that's how it is now.

housseindjirdeh:
Ah that's a fair point. I'm still a fan of creating branches under my local fork since it also should be fetchable since it's all public, but I'm okay with that approach as well 👍

PR #357, opened 2017-09-25
chore: Display 6 contributors per line, credit translations

machour:
Changed numbers of contributors from 7 to 6 per lines, this will avoid layout issues due to wrapping in our README.md : 

Before:
screen shot 2017-09-25 at 4 42 26 pm

After:
screen shot 2017-09-25 at 4 42 07 pm

I also upgraded to last all-contributors-cli to latest version, and credited translators using the "translation" key. This will make it easier to spot them when a translation is needed.


andrewda:
The contributors section is getting very large. Do we want to consider moving it to a CONTRIBUTORS.md file?

housseindjirdeh:
@andrewda Hmmm I don't know, I quite like it being visible right in the README

What do you think @machour?

machour:
I had the exact same thought as @andrewda but refrained from doing it cause I was unsure of your opinion on that :D

housseindjirdeh:
No worries, happy to have it in there if most people think it should belong there.

#359 open to address this :D

PR #360, opened 2017-09-25
feat(tabbar): show notifications count

lex111:
So now the lazy loading of tabs works, then probably it makes sense to display the number of notifications in the tabbar.

This is the early version, you still need to deduct the notifications that were read by the **repository** (you need to determine how many they need to be  deducted, and do it by analogy [here](https://github.com/gitpoint/git-point/pull/360/files#diff-fae302351fa6f98e5f28a5995b1b14adR97)).

Quick demo:

If there are more than 99 notifications, 99+ will be displayed.

Also I had an idea with auto-updates of the number of notifications? Is it necessary?

We can also think about how to make this component differently, more precisely how to connect `redux` and `react-navigation`? I had an idea to make a component of the `TabBarIcon` type with the `badgeText` property and pass the count there, but it's not clear how to connect `redux` then to update the badge text.

Feel free to suggest changes, if you want.


lex111:
I made update the badge after mark the repo notifications, and also changed a bit of the icon design:

| > 99 | < 99
| - | - |
|  |

src/config/colors.js
```diff
@@ -17,6 +17,7 @@ export const colors = {
   darkGreen: '#27ae60',
   red: '#ee0701',
   darkRed: '#e74c3c',
+  redDarkest: '#da0000',
```
machour: darkerRed

src/api/index.js
```diff
@@ -300,6 +300,40 @@ export async function fetchMarkRepoNotificationAsRead(
...
+
+  let linkHeader = response.headers.get('Link');
+  let count = 1;
+
+  if (linkHeader) {
+    linkHeader = linkHeader.match(/page=(\d)+/g).pop();
+    count = linkHeader.split('=').pop();
+  }
+
+  return +count;
```
machour: what do `+count` do ? Oo

src/api/index.js
```diff
@@ -300,6 +300,40 @@ export async function fetchMarkRepoNotificationAsRead(
...
+export async function fetchRepoNotificationsCount(
+  owner,
+  repoName,
+  accessToken
+) {
+  const ENDPOINT = `${root}/repos/${owner}/${repoName}/notifications?per_page=1`;
+  const response = await fetch(ENDPOINT, accessTokenParameters(accessToken));
+
+  let linkHeader = response.headers.get('Link');
+  let count = 1;
```
machour: if the Link header is not present, count should be the count of json objects returned in the response, not 1

machour: See how this is handled in v3.count() I just added: https://github.com/gitpoint/git-point/pull/361/commits/6ce848985be790541f1cb2029e7a061dca47226b#diff-284817f1b6aad47c0ac6d61bd72f474cR58

lex111: > if the Link header is not present, count should be the count of json objects returned in the response, not 1
  
The fact is that when the notification in the repository is only one, the Link header will not be, and `response.json().length` will return 0, which is not correct. Therefore, the default is 1.

src/api/index.js
lex111: This convert the value to a number, similarly does the call to `parseInt()`

machour: Nice, thanks

src/api/index.js
lex111: > See how this is handled in v3.count() I just added: 6ce8489#diff-284817f1b6aad47c0ac6d61bd72f474cR58
  
Can you add a third parameter, such as `defaultCount`?

machour: @lex111 as I understand, here's how the API respond : 
- if you have 0 notifications, response.status is 404 with a json body that must be ignored
- if you have notifications : 
  - if only 1, no Link header and response.json() is an array of 1 element (so json().length = 1)
  - if more, you parse the Link header to get the number

am I missing something?

lex111: Yes, that's right, I checked again,  and found the error that [you have now](https://github.com/gitpoint/git-point/pull/361/commits/6ce848985be790541f1cb2029e7a061dca47226b#diff-284817f1b6aad47c0ac6d61bd72f474cR67), and because of which I incorrectly put 1

`response.json()` is a promise, so it should be like this:

```js
number = response.json().then(data => {
  return data.length;
});
```

machour: OMG, thanks! didn't know response.json() was a promise, I thought only response was.
I'll update count() to handle the 404, fix json().
Do you still think defaultValue is needed? I fail to see how

lex111: > Do you still think defaultValue is needed? I fail to see how

No, at least for now.

machour: just fixed the `count()` method and realized that json().length isn't needed at all as we request one element per page -_-

So, if 404 => 0 elements
If no Link header => 1 element
Else, count fetched from Link

I wonder if the 404 status code is sent for all empty request responses ..

machour: (let's chat over gitter if you want)

lex111:
I made changes according with the feedback, and now I thought, can show the total number notifications, instead 99+? I did it because I wanted the badge to be always round, and if there is a four-digit number, then it will be oval.

machour:
I personally think +99 is perfect

lex111:
Here is an example of how to make a badge:

image

src/components/notification-icon.component.js
```diff
@@ -0,0 +1,68 @@
...
+      <View>
+        <Icon
+          containerStyle={{ justifyContent: 'center', alignItems: 'center' }}
+          color={iconColor}
+          name="notifications"
+          size={33}
+        />
+
+        {notificationsCount
+          ? <View style={styles.badge}>
```
machour: Hey @lex111, maybe we could use <Badge> from react-native-elements which is already bundled : 
  https://react-native-training.github.io/react-native-elements/API/badge/

src/auth/screens/events.screen.js
```diff
@@ -86,8 +89,11 @@ class Events extends Component {
     }
   }
 
-  getUserEvents = (user = this.props.user) => {
+  getUserEvents = () => {
```
andrewda: We're passing a user into this function on lines 82 and 88, and this would be therefore ignoring it. Maybe something like this will work?

```javascript
getUserEvents = ({ user, accessToken } = this.props) => {
    // ...
}
```

While changing lines 82 and 88 to pass in `this.props` and `nextProps` respectively.

lex111: I tried to do it, it did not work.

src/components/notification-icon.component.js
lex111: It's not round, I like that currenly it's just round in shape, well okay

src/auth/screens/events.screen.js
housseindjirdeh: Agree with @andrewda. Line 82 shouldn't matter since it's already defaulting to `this.props.user` (would prefer if you can remove the argument call in there (line 82) if you can too :)

But if you're removing the arguments entirely from the function declaration in line 92, line 88 doesn't look it would work at all. We're passing in `nextProps.user` but in here it's not being used at all..

Curious why 

```
getUserEvents = ({ user, accessToken } = this.props) => {
    // ...
}
```

doesn't work? Destructuring and applying a default should work if I'm not mistaken

src/components/notification-icon.component.js
housseindjirdeh: Fair enough, if you would rather have a circular badge and think it looks neater I don't mind :)

I think we can do one of two things however:

1. Could we have a local badge component that contains the style for the badge and expose a background color and text color prop? Feels nice to have that exposed in it's own component. Would prefer if you call it `circular-badge` or something along those lines to not name-conflict with the react native elements badge component
2. OR we _could_ use React Native Elements badge here and just apply `containerStyle` to make it circular.

I'm okay with either option :raised_hands:

src/notifications/screens/notifications.screen.js
```diff
@@ -163,6 +167,8 @@ class Notifications extends Component {
     } = this.props;
     const { type } = this.state;
 
+    this.props.getNotificationsCount();
```
housseindjirdeh: Could we have this as part of our destructured object up above:

```
     const {
       getUnreadNotifications,
       getParticipatingNotifications,
       getAllNotifications,
       getNotificationsCount
      } = this.props;
```

So we can just call `getNotificationsCount()` here

src/notifications/screens/notifications.screen.js
```diff
@@ -145,6 +148,7 @@ class Notifications extends Component {
     this.props.getUnreadNotifications();
```
housseindjirdeh: Would prefer is we destructure here as well (I know this is slightly out of your scope however so don't worry if you would rather not):

```
     const {
       getUnreadNotifications,
       getParticipatingNotifications,
       getAllNotifications,
       getNotificationsCount
      } = this.props;
```

And then `Promise.all` all the functions instead of calling them one by one

src/components/notification-icon.component.js
lex111: Strangely, I can not import this component:

```js
import { Icon, Badge } from 'react-native-elements';

console.log(Badge); // undefined
```

I do not understand what's up, okay, I'll make a local component then.

lex111:
Ready for review.

src/notifications/screens/notifications.screen.js
lex111: Done except promises, how can they help?

housseindjirdeh: No worries, ideally `Promise.all` is used to make sure a _collection_ of promises are complete before submitting an event after them but since we're not really doing anything after each of these specifically - I think it's okay if we don't include it here :)

src/auth/screens/events.screen.js
```diff
@@ -74,20 +77,21 @@ const styles = StyleSheet.create({
 });
 
 class Events extends Component {
-  componentDidMount() {
-    if (this.props.user.login) {
-      this.getUserEvents(this.props.user);
+  componentDidMount({ user } = this.props) {
```
housseindjirdeh: I don't think we should be including `props` as our an argument to `componentDidMount`. We never call this method as a function anywhere in the component so there's no need to set an argument and define it's default (seems redundant).

Ideally, we can just destructure within the method itself:

```
componentDidMount() {
    const { user } = this.props;
   // ....
}
```

lex111: Oops, that's right, I fixed it.

PR #361, opened 2017-09-26
refactor(api): Api clean up

machour:
While looking at redux, I came accros a few facts with api/index.js : 

- `fetch()` is being called in a lot of functions, inside and outside `api/index.js`. Some actions files also borrowed the accessTokenParameters* helpers.
- weird function naming: `fetchUrl` for GET returning json, and `fetchNormal` for GET returning a `Response`
- duplicate functions: `fetchUrlCommentsHTML` was in fact performing the exact same thing as `fetchUrl`
- code was piling up and the file was getting unmaintainable.

This PR cleans up all this, and makes the API crystal clear, by introducing the `v3` namespace:

v3 is a simple object representing our GitHub Rest API v3 client :
- `v3.call()` is the only function that calls GitHub. It receives the `url` and `parameters` arguments for the `fetch()` call and returns a Response
- `v3.root` is the root endpoint
- `v3.parameters()` is the sole responsible for preparing the `fetch()` call parameters. It takes the accessToken **(mandatory)**, the HTTP method *(optional)*, the Accept header  *(optional)* and a body for the request *(optional)*, does its magic, and returns the correct object.
- The rest of the object consist of helpers method using a clear nomenclature : 
       **httpVerb** **AcceptedResult** ()

For example:
  - `v3.get()` performs a `GET` query and returns the `Response`
  - `v3.getJson()` performs a `GET` query and returns the .json() response
  - `v3.head()` performs a `HEAD` query and returns the `Response`
  - `v3.postRaw()` performs a `POST` request with the given object and returns the .text() response

I implemented all the needed helpers, and if a new combination arises (`postHtml()` for example), it should be added in v3 following this nomenclature.

I also adopted two enums, `METHOD` and `ACCEPT` to attempt to standardize things a bit more.

This clean up made obvious that `fetchStarCount()` was missing the accessToken, and that some weird calls were made by the watch/unwatch actions. I fixed the calls.

I did my best to test all the apps, but I might have missed a few spots, please let me know if it's the case!

Next step (once this is merged), is to re-dispatch all remaining functions (aside v3 of course) from `api/index.js` to the actions files where they belong.

### DIFF
--- a/src/api/index.js
+++ b/src/api/index.js
@@ -58,22 +58,18 @@ export const v3 = {
 
     return params;
   },
-
-  getJson: async (url, accessToken) => {
+  delete: async (url, accessToken) => {
     const response = await v3.call(
       url,
-      v3.parameters(accessToken, METHOD.GET, ACCEPT.JSON)
+      v3.parameters(accessToken, METHOD.DELETE)
     );
 
-    return response.json();
+    return response;
   },
-  getHtml: async (url, accessToken) => {
-    const response = await v3.call(
-      url,
-      v3.parameters(accessToken, METHOD.GET, ACCEPT.HTML)
-    );
+  get: async (url, accessToken) => {
+    const response = await v3.call(url, v3.parameters(accessToken));
 
-    return response.text();
+    return response;
   },
   getDiff: async (url, accessToken) => {
     const response = await v3.call(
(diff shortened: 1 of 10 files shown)


lex111:
@machour this is an amazing job! Could you please add [method for executing a GraphQL query](https://github.com/gitpoint/git-point/pull/347/files#diff-284817f1b6aad47c0ac6d61bd72f474cR57) from PR #347?

machour:
@lex111 GraphQL should be in a `v4` namespace, it will be added in another PR.
I'm trying to keep this PR as small and focused as possible for now to avoid side effects.

For the plurals, are you talking about changing `METHOD` to `METHODS` and `ACCEPT` to `ACCEPTS`, or do you mean something else ? 🤔

lex111:
Yes, that's right, GraphQL is already v4.

> For the plurals, are you talking about changing METHOD to METHODS and ACCEPT to ACCEPTS, or do you mean something else

Correct, do you think?

machour:
@lex111 it seems more natural to me to use the singular form. The idea is that you are selecting **one** method of the possible choices in the enum. I always see that in java projects for example.

PR #366, opened 2017-09-26
refactor(profile): show user data when it is fully loaded

lex111:
This avoids the problem shown below: i.e. the number of stars has not yet loaded, and the number of repositories has already appeared, but you need to wait until all the data is loaded and only then show them all.

Before | After
| - | -
|  |

### DIFF
--- a/App.js
+++ b/App.js
@@ -59,7 +59,11 @@ class App extends Component {
 
     persistStore(
       configureStore,
-      { storage: AsyncStorage, transforms: [encryptor] },
+      {
+        storage: AsyncStorage,
+        transforms: [encryptor],
+        blacklist: ['user'],
+      },
       () => {
         this.setState({ rehydrated: true });
       }
--- a/src/user/screens/profile.screen.js
+++ b/src/user/screens/profile.screen.js
@@ -102,13 +102,7 @@ class Profile extends Component {
   }
 
   componentDidMount() {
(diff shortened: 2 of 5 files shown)


src/user/screens/profile.screen.js
```diff
@@ -102,13 +102,7 @@ class Profile extends Component {
...
 
   componentDidMount() {
-    const user = this.props.navigation.state.params.user;
-    const auth = this.props.auth;
-
-    this.props.getUserInfo(user.login);
-    this.props.getStarCount(user.login);
-    this.props.getIsFollowing(user.login, auth.login);
-    this.props.getIsFollower(user.login, auth.login);
+    this.getUserInfo();
```
housseindjirdeh: :tada:

App.js
```diff
@@ -59,7 +59,11 @@ class App extends Component {
 
     persistStore(
       configureStore,
-      { storage: AsyncStorage, transforms: [encryptor] },
+      {
+        storage: AsyncStorage,
+        transforms: [encryptor],
+        blacklist: ['user'],
```
housseindjirdeh: We should probably `blacklist` more than just _user_ right? Something we need to discuss once we rearchitect our logic and state

lex111: Yes, it is probably possible to supplement it. Need to discuss.

Issue #348, opened 2017-09-22
Maintain and use a node-github fork?


machour:
I was playing around with redux & normalizr and came across this project:
   https://github.com/mikedeboer/node-github

It's a very complete & highly maintained client for GitHub Rest API v3
Here's a link to its API documentation: https://mikedeboer.github.io/node-github/

Maybe we could benefit from using it in the app instead of our own crafted client (node-github already handles everything we will need + a standardized pagination). We would remove any API v3 hassle :)

Unfortunately it's not React Native compatible, but the good news is that someone already tweaked a fork: https://github.com/brunolemos/react-native-github-api.

As the fork didn't seem really maintained, I went ahead and forked latest node-github master and made it RN compatible: https://github.com/machour/node-github

What are yout thoughts about maintaining our own fork (gitpoint/react-native-github-api) and simply keep it in sync with node-github? Other people would be able to benefit from it.

machour:
Abandoned in #457 

*walks out whistling*

PR #367, opened 2017-09-26
refactor(api): split and host api/index.js functions in *.api.js

machour:
Following up on #361


src/auth/auth.action.js
```diff
@@ -5,13 +5,11 @@ import { delay, resetNavigationTo, configureLocale } from 'utils';
...
 
 import {
-  fetchAccessToken,
-  fetchAuthUser,
-  fetchAuthUserOrgs,
   fetchUserOrgs,
   fetchUserEvents,
   fetchStarCount,
-} from 'api';
+} from '../user/user.api';
```
lex111: We have [aliases](https://github.com/gitpoint/git-point/blob/master/.babelrc), so that you can replace everywhere relative paths like `../user/user.api` with `user/user.api`, should work as before.

machour: I learned something new, thanks @lex111 !

machour:
I'm having second thoughts about this refactoring, as these functions doesn't seem to belong there:

`src/issue`, `src/user/`, .. in GitPoint doesn't represent the v3 API endpoints, but represent instead the sections in the app. 

It would make more sense to host the API related functions under `src/api/v3/`, and split them by endpoints. (`api/v3/users`, `api/v3/repos`, ..)

I also think we should consider renaming those functions in order to have more conventional names. `fetchSearch` for example doesn't feel right at all. `v3.search.users()` or `v3.search.repos()` seems a bit more standard as per GitHub naming.

I had a look at several GitHub clients apis over the last days, and I really love the naming  [node-github](http://mikedeboer.github.io/node-github/) adopted. I've opened a proposal in #348 to switch to this library for v3, but I'm not sure how it would play with our redux/normalizr plans..
We can however adopt the same naming for our home made client. 

Furthermore, we could even make use of [this file](https://github.com/mikedeboer/node-github/blob/master/lib/routes.json) in a little script to generate our API functions (only the one needed in the app). 🤔 

Something to consider right after release Monday!

src/auth/auth.action.js
housseindjirdeh: Yep, all thanks to [babel-plugin-module-resolver](https://github.com/tleunen/babel-plugin-module-resolver)

housseindjirdeh:
Thank you so much mate for taking the time to split these up in their separate domains.

I think this is _tons_ better than having a single `index` file with all of our API logic.

That being said I kinda agree. Having everything named `fetch` isn't the nicest way of doing things _and_ a convention would be nice. I also don't think it's a bad idea to have all the logic within `api/v3` and not keep their concerns feature-based.

Will love to hear other's opinions on this as well. Also happy to merge this in now if you would like to deep-dive and refactor and fix things up next week or close this since we'll be doing that anyway :)

machour:
Closing this for now, we'll handle that more properly next week 🙌

PR #370, opened 2017-09-26
fix(markdown): fix various markdown issues

### LINKED ISSUE #363: Markdown rendering issues
Spotted by @Antoine38660 <3

When viewing #360 in the app, it crashes with the following error : Views nested within a <Text> must have a width and height.
This is due to the use of images in a table.

We also discovered a weird markdown in this case caused by https://github.com/gitpoint/git-point/blob/master/src/components/markdown-htmlview.component.js#L142

Also, the table headers content should be HTML decoded : 
screen shot 2017-09-26 at 12 42 54 pm

There's also a, less critical, rendering issue for #361 : 
capture-d_e_cran-2017-09-26-a_-13 46 44

machour:
Fix #363 

The problems were : 

- Text inside `<th>` was HTML escaped: it needed to be displayed using `defaultRenderer`
- An `` inside a `<td>` caused a crash (View in Text): We created our own Cell component that avoids using `<Text>`
- marked() was rendering a simple `piece of code` as a `<pre>` if inside a `<li>`: Setting it to be pedantic helped

I also made images inside cells take 30% of the screen width while enlarging a bit images outside cells.

### DIFF
--- a/src/components/markdown-htmlview.component.js
+++ b/src/components/markdown-htmlview.component.js
@@ -81,11 +81,7 @@ marked.setOptions({
   renderer: new marked.Renderer(),
   gfm: true,
   tables: true,
-  breaks: false,
-  pedantic: false,
-  sanitize: false,
-  smartLists: true,
-  smartypants: false,
+  pedantic: true,
 });
 
 const { height, width } = Dimensions.get('window');
--- a/src/components/markdown-htmlview.component.js
+++ b/src/components/markdown-htmlview.component.js
@@ -98,6 +98,47 @@ const headingRenderer = (node, index, siblings, parent, defaultRenderer) => {
   );
 };
 
+/* eslint-disable no-shadow */
+class CellWithImage extends Cell {
+  render() {
+    const { data, width, height, flex, style } = this.props;
+    let borderWidth;
+    let borderColor;
+
+    if (this.props.borderStyle && this.props.borderStyle.borderWidth) {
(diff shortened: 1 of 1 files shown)


src/components/markdown-htmlview.component.js
```diff
@@ -102,6 +98,47 @@ const headingRenderer = (node, index, siblings, parent, defaultRenderer) => {
   );
 };
 
+/* eslint-disable no-shadow */
+class CellWithImage extends Cell {
+  render() {
+    const { data, width, height, flex, style } = this.props;
+    let borderWidth;
```
lex111: Do you think, if you set the variables (`borderWidth`, `borderColor`) to default values, and thus remove the `else` branch?

src/components/markdown-htmlview.component.js
```diff
@@ -266,19 +306,33 @@ export class MarkdownHtmlView extends Component {
             styleText.textAlign = cellAlign[attribs.style];
           }
 
+          const Component =
+            node.children.filter(
+              elem => elem.type === 'tag' && elem.name === 'img'
```
lex111: Do you think it would be possible to move this arrow function in a separate function, if it is used more than once?

machour: done!

src/components/markdown-htmlview.component.js
machour: done

PR #371, opened 2017-09-27
feat(notifications): add button to mark all notifications as read

lex111:
image


src/notifications/screens/notifications.screen.js
```diff
@@ -147,6 +165,16 @@ class Notifications extends Component {
     this.props.getAllNotifications();
   }
 
+  componentWillReceiveProps(nextProps) {
+    if (
+      !nextProps.isPendingMarkAllNotificationsAsRead &&
+      this.props.isPendingMarkAllNotificationsAsRead &&
+      !this.isLoading()
+    ) {
+      this.getNotifications()();
```
machour: ()() ?

src/config/colors.js
```diff
@@ -29,6 +29,8 @@ export const colors = {
   purple: '#8e44ad',
   orange: '#e67e22',
   githubDark: '#1f2327',
+  mercury: '#e3e3e3',
+  shark: '#24292e',
 };
```
machour: I know those are official names, but they're not explicite at all.
Also, do we really need to add news colors?

src/locale/languages/en.js
```diff
@@ -135,6 +135,7 @@ export const en = {
       allButton: 'All',
       retrievingMessage: 'Retrieving notifications',
       noneMessage: "You don't have any notifications of this type",
+      markAllAsRead: 'Mark all notifications as read',
```
machour: This sentence is too long for a button, + translations may be bigger.
Can we use "Mark all read" instead?

src/notifications/screens/notifications.screen.js
```diff
@@ -110,6 +116,16 @@ const styles = StyleSheet.create({
...
+    marginTop: 3,
+    paddingTop: 3,
+    paddingBottom: 3,
+    marginLeft: 11,
+    marginRight: 11,
+    borderColor: colors.mercury,
+    borderWidth: 1,
+    borderRadius: 3,
+  },
 });
```
machour: I'm not a fan of this button design, and it looks too different from the "Merge pull request" button we have : 
screen shot 2017-09-27 at 6 52 55 pm

andrewda: Hmm this is interesting... I agree that we should keep to a similar style in general, but I don't think a bulky button like this taking up a bunch of space is the best UI. I like how this style is more discrete at the top of the screen.

machour: Absolutely! Right after commenting on this, I started implementing a standard <Button /> component to be used everywhere (small/normal/big, info/danger/..) :)

lex111: Maybe yes, we can not make this button fixed, then it will be before the first notification, how do you do?

In style, I think this is the best option, I mean that the button is stretched to full width.

src/notifications/screens/notifications.screen.js
lex111: Because this function returns a function, it still needs to be called, so one more parentheses.

src/config/colors.js
lex111: As I understand it, we add all the colors to the configuration, or am I mistaken?

machour: yes, new colors should be added in configuration.
I was asking if we could nor reuse existing colors instead

src/locale/languages/en.js
lex111: The size of the button allows, so I made a detailed description, well I'll remove the word "notifications".

src/config/colors.js
lex111: Okay, I remove those colors

lex111:
Updated version
image

machour:
@lex111 what about putting the button to the bottom of the screen? Makes more sense to me: I go to the tab, scroll down while watching my notifications, then decide to mark them all read

lex111:
It's rather controversial, I'm comfortable when it's on top, and that's why I made it position is fixed initially, so that it was always visible .. let's wait for the opinions of others.

src/notifications/screens/notifications.screen.js
chinesedfan: How about rename `getNotifications` as `getFuncForGettingNotifications`, or something else?

lex111: It is unlikely that the first name is fine (`getFuncForGettingNotifications`), because this function does two things: not only returns another function, but also updates the notification counter.

chinesedfan: @lex111 Sorry, I didn't quite get your words. Can you tell me where "updates the notification counter"?

lex111: It's you who excuse me :smile: , I mixed up with my other PR #360  - I'm doing a notification counter there and there's [this](https://github.com/gitpoint/git-point/blob/f522c6bc7781864083418a858174d5026ed4f6c2/src/notifications/screens/notifications.screen.js#L171) function.
But we need to think more, we can be renamed.

andrewda:
I like the top, personally. It's more like what GitHub has.

src/notifications/screens/notifications.screen.js
housseindjirdeh: I think something like `getNotificationsForCurrentTab` or something along those lines will be okay

src/notifications/screens/notifications.screen.js
```diff
@@ -233,115 +309,104 @@ class Notifications extends Component {
...
-              color={colors.greyDark}
-              size={28}
-              name="check"
-              type="octicon"
+            title={translate('notifications.main.markAllAsRead')}
+            buttonStyle={styles.markAllAsReadButton}
+            color={colors.black}
+            backgroundColor={colors.white}
+            textStyle={styles.buttonGroupText}
+            onPress={() => markAllNotificationsAsRead()}
```
housseindjirdeh: Does this button show when `unread`, `pending` _and_ `all` tabs are selected? Not sure if that's the best way of doing things. Do you think it will make more sense to have it only show only when `all` is selected (and then maybe having buttons that show `Mark all unread notifications as read` and `Mark all participating notifications as read` for the other 2 tabs?

housseindjirdeh: Even if you think we don't need the other two buttons in this PR, I think having the Mark ALL button in each of the tabs can be kind of misleading

lex111: Currently, on notifications [page](https://github.com/notifications) this button is displayed everywhere, and it always marks notifications as read _all_ notifications, i.e. there is no way to mark as read only _participating_ notifications.

Let's display this button only on the first tab (_unread_)?

src/notifications/screens/notifications.screen.js
```diff
@@ -196,15 +234,53 @@ class Notifications extends Component {
...
+
+    switch (type) {
+      case 0:
+        return unread;
+      case 1:
+        return participating;
+      case 2:
+        return all;
+      default:
+        return null;
```
jouderianjr: Hey, looks like `#notifications` method should always return an array, it is safer in `default` return also an array( probably [])

I know that default cause should be not called, but I believe it's always nice to try keep the `type return` consistency 💃

src/notifications/screens/notifications.screen.js
```diff
@@ -196,15 +234,53 @@ class Notifications extends Component {
...
+
+    switch (type) {
+      case 0:
+        return unread && isPendingUnread;
+      case 1:
+        return participating && isPendingParticipating;
+      case 2:
+        return all && isPendingAll;
+      default:
+        return null;
```
jouderianjr: Same of `#notifications` method, but this time I should return a boolean.

src/notifications/screens/notifications.screen.js
```diff
@@ -196,15 +234,53 @@ class Notifications extends Component {
...
-    markAsRead(notification.id);
-    navigation.navigate('Issue', {
-      issueURL: notification.subject.url.replace('pulls', 'issues'),
-      isPR: notification.subject.type === 'PullRequest',
-      language: this.props.language,
-    });
+    this.setState({ contentBlockHeight: height - sizes.tabBar });
+  };
+
+  keyExtractor = (item, index) => {
```
jouderianjr: Hey, this is a pretty small piece of code, I think that could be added inline in the <FlatList> like that: 

`keyExtractor={(item, index) => index}`

src/notifications/screens/notifications.screen.js
chinesedfan: @lex111 I prefer to keep `getNotifications` as simple as now, instead of mixing `getNotificationsCount` in it.

Because there is [another place](https://github.com/gitpoint/git-point/blob/f522c6bc7781864083418a858174d5026ed4f6c2/src/notifications/screens/notifications.screen.js#L397) calls this function, which just expects a function. After #360 is merged, I guess `getNotificationsCount` will be called frequently by `render`.

src/notifications/screens/notifications.screen.js
housseindjirdeh: Ah that's a very good point 🤔 

Hmmmmmm, I think I'm okay with having it display only on the first tab in the case. If I'm not mistaken, all _participating_ notifications are _unread_ notifications anyway so this should be okay 👍

lex111:
Ready for review.

PR #380, opened 2017-09-28
fix(issues): handle PR struct correctly

machour:
We were having an `Uncaught error in promise` due to the `issueParam` navigation state sometime being a `PullRequest` instead of an `Issue` .. This caused the Merge permission check & diff to fail.

Here's a hotfix for now, but this definitely call for a big clean up post release, and why not a separate `src/pulls/` implementation.

### DIFF
--- a/src/issue/screens/issue.screen.js
+++ b/src/issue/screens/issue.screen.js
@@ -180,19 +180,21 @@ class Issue extends Component {
         issueURLParam ? issueCommentsURL : issueParam.comments_url
       ),
     ]).then(() => {
-      if (
-        issueParam &&
-        repository.full_name !==
-          issueParam.repository_url.replace(`${v3.root}/repos/`, '')
-      ) {
-        Promise.all([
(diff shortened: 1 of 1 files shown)


machour:
Prettier messed the diff a bit, here is it without whitespaces changes : 
  https://github.com/gitpoint/git-point/pull/380/files?w=1

I did only modify the changes near the if(), prettier did the rest.

chinesedfan:
@machour @housseindjirdeh How about my #332 ? I think it solved the same problem.

machour:
@chinesedfan oh man, so sorry, didn't know this was already addressed by your PR!
@housseindjirdeh please check #332 before this one, the fix is cleaner!

housseindjirdeh:
Just merged in #332, apologies for not merging that one in sooner and having this duplication happen @machour 😞

machour:
No problems, and thanks again @chinesedfan <3

PR #384, opened 2017-09-29
fix(comments): fix long press click error

jouderianjr:
Before this patch, the long press event in comment list item worked
in comments from other users, this action should work the logger
user comments.

Sorry for that mates! Was my bad 🙊

### DIFF
--- a/src/components/comment-list-item.component.js
+++ b/src/components/comment-list-item.component.js
@@ -150,7 +150,8 @@ class CommentListItemComponent extends Component {
 
     return (
       <TouchableWithoutFeedback
-        onLongPress={() => isActionMenuEnabled && this.showMenu()}
+        disabled={!isActionMenuEnabled}
+        onLongPress={this.showMenu}
       >
         <View style={styles.container}>
           <View style={styles.header}>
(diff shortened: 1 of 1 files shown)


src/components/comment-list-item.component.js
```diff
@@ -143,13 +143,15 @@ class CommentListItemComponent extends Component {
...
+    const isActionMenuEnabled =
       comment.user &&
       authUser.login === comment.user.login &&
       !this.isIssueDescription();
 
     return (
-      <TouchableWithoutFeedback onLongPress={this.showMenu}>
+      <TouchableWithoutFeedback
+        onLongPress={() => isActionMenuEnabled && this.showMenu()}
+      >
```
chinesedfan: How about the `disabled` attribute? It can avoid creating an anonymous function in `render`.

jouderianjr:
@chinesedfan Thanks, this approach is way better!

housseindjirdeh:
Seems like we'll be removing this entirely @jouderianjr? But merging this in if we're still discussing :)

jouderianjr:
@housseindjirdeh yes, should be removed, this PR was created before our discussion :D I already opened a PR removing it 💃

Issue #381, opened 2017-09-28
v1.3.0 Prep


housseindjirdeh:
## Next release: v1.3.0 - Monday October 2nd

_Note: iPhone is currently in v1.2.0 and Android is currently in v.1.1.0. Going to bump version number to 1.3.0 so that they both match_

- Today (September 28) - Plan is to close and merge all open PRs that are safe to merge

- The next few days, we should focus on fixing any bugs and applying UI fixes to anywhere that needs it. Please tag this issue on those PRs. We'll put a hold to merging large features/updates to the app until after Monday.

## Some things that need updating:

- [x] ImageZoom component shows a black screen when image is tapped (iPhone + Android)

   EDIT - Just seems to be the native module not linking properly. Thanks @machour for pointing this out

![image](https://user-images.githubusercontent.com/12476932/30973594-82bc7942-a43b-11e7-9d20-0dd7ec968b0a.png)

- [X]  Font on events screen seems _too_ bold on iPhone (only seeing this issue myself, @machour can't reproduce it which is interesting <span class="emoji-outer emoji-sizer"><span class="emoji-inner" style="background: url(chrome-extension://immhpnclomdloikkpcefncmfgjbkojmh/emoji-data/sheet_apple_64.png);background-position:56.25000000000001% 70.83333333333334%;background-size:4900%" data-codepoints="1f914"></span></span>)

![image](https://user-images.githubusercontent.com/12476932/30973677-b156341e-a43b-11e7-8e84-c8a7968d195f.png)

- [x] Something that I've been thinking, update fonts in app to [Nunito](https://fonts.google.com/specimen/Nunito)

![image](https://user-images.githubusercontent.com/12476932/30974049-a4517f3e-a43c-11e7-971e-dd62d42cc454.png)

Have been working on a React Native app this past week that uses this font and it really looks _crisp_. Will be also be perfect to have both platforms use the same font. Will give this a shot :)

- [X] Change notifications count badge, make number slightly larger/prominent

Instead of splitting these concerns into separate tickets, I'm okay with having PRs for each but just using this ticket ID. Anything that you can see that we most definitely need to include, don't hesitate to comment it here and we'll add it in the list.

# EXCITED TO GET A NEW RELEASE OUT!

![excited](https://media.giphy.com/media/pUs87dXYIIrTy/source.gif)

housseindjirdeh:
After our release, we'll spend some significant amount of time improving a number of things in the app:

[...]

PR #386, opened 2017-09-29
feat(markdown): Use markdown rendered on GitHub side

machour:
Silenced all missing keys warnings.
Fixed an issue where a mention or a issue link wasn't recognized at the start of a line.

#381 <- should work
@housseindjirdeh <- should work too


machour:
Noticed yet a few more problems
screen shot 2017-09-29 at 2 03 04 pm

machour:
Ok, so this is turning out to be more than small fixes for markdown.
I discovered that when we pass the HTML accept header to github API, all markdown gets rendered server side, and a clean HTML transformation is provided in `body_html`.

So we're getting rid of MarkdownHtmlView in favor of GithubHtmlView.

Still work in progress to fix some small glitches

machour:
bummer: syntax highlighting doesn't work anymore.. fixing it now.

src/components/comment-list-item.component.js
```diff
@@ -201,14 +199,14 @@ class CommentListItemComponent extends Component {
             </View>
           </View>
 
-          {!!commentPresent && (
+          {!!commentPresent &&
             <View style={styles.commentContainer}>
-              <MarkdownHtmlView
-                source={comment.body}
+              <GithubHtmlView
+                source={comment.body_html}
```
housseindjirdeh: I think you might have meant to use `comment.body`. Noticing this when I open an issue:

image

From the looks of it, you may have accidentally swapped it by accident :)

image

machour: @housseindjirdeh We should be using only comment.body_html from now on instead. I tried the branch again and everything works fine for me. Could you give it a new try?
Fair enough, your comment have made me realize that Adding a comment was broken, I fixed it and completely got rid of comment.body :D

housseindjirdeh: Yeah was thinking the same thing, appreciate you diving in and removing `body` entirely since we don't really use it anymore

PR #387, opened 2017-09-29
feat(issue): Show issues author and comments count

chinesedfan:
- Before:

- After:

The subtitle seems to be changed a little, because its original styles are provided by `react-native-elements`. Now in order to use custom subtitle, I set them based on our constants.

I am not very good at UI design. Waiting for your suggestions. :)

### DIFF
--- a/src/components/issue-list-item.component.js
+++ b/src/components/issue-list-item.component.js
@@ -32,7 +32,7 @@ const styles = StyleSheet.create({
   },
   title: {
     color: colors.primaryDark,
-    ...fonts.fontPrimary,
+    ...fonts.fontPrimarySemiBold,
   },
   subtitle: {
     marginTop: 2,
--- a/src/components/issue-list-item.component.js
+++ b/src/components/issue-list-item.component.js
@@ -1,6 +1,6 @@
 import React from 'react';
 import { StyleSheet, TouchableHighlight, View, Text } from 'react-native';
-import { ListItem } from 'react-native-elements';
+import { ListItem, Icon } from 'react-native-elements';
 import moment from 'moment/min/moment-with-locales.min';
 
 import { StateBadge } from 'components';
@@ -35,21 +35,19 @@ const styles = StyleSheet.create({
     color: colors.primaryDark,
     ...fonts.fontPrimary,
   },
-  subtitleView: {
+  commentsContainer: {
     flexDirection: 'row',
     marginTop: 2,
   },
-  desc: {
-    flex: 1,
+  comments: {
+    paddingLeft: 5,
     fontSize: normalize(12),
     color: colors.greyBlue,
   },
-  comments: {
-    paddingHorizontal: 5,
+  description: {
+    marginTop: 2,
     fontSize: normalize(12),
-    backgroundColor: colors.greyLight,
-    borderRadius: 5,
-    overflow: 'hidden',
+    color: colors.greyBlue,
   },
   badge: {
     alignItems: 'flex-end',
@@ -57,7 +55,7 @@ const styles = StyleSheet.create({
   },
 });
 
-export const IssueListItem = ({ type, issue, navigation, language }: Props) =>
+export const IssueListItem = ({ type, issue, navigation, language }: Props) => (
   <TouchableHighlight
     style={issue.state === 'closed' && styles.closedIssue}
     onPress={() =>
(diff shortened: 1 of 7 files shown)


machour:
@chinesedfan Nice addition ! We are currently in a feature-freeze mode preparing 1.3 (see #381) but maybe @housseindjirdeh will let this one ships in.

You can join us on https://gitter.im/git-point/Lobby to help us prepare 1.3 this week end ;)

lex111:
@chinesedfan how about moving the comment counter to the end of the issue title and adding a [comment icon from Octicons](https://octicons.github.com/icon/comment/) in front of it using the Icon component from `react-native-elements`? The background can then be removed, only can add a small padding on the left.

image

lex111:
And it's unclear what the date really is - the date of the start or update, I would add clarification like `created 3d ago` or `opened 3d ago`.

Or let's then make it as on GitHub? 
`#501 opened 3 days ago by mhf-ir`

housseindjirdeh:
@machour Thanks for mentioning that. I think we can work having this shipped in our next release :)

@chinesedfan thanks a million for this. @lex111 said _exactly_ what I was thinking :D. If you would rather not work on uI specifics however let us know! But ideally having:

1. Comment octicon with number of comments. This can either be at the end of the issue title OR I think it might be better as a separate line part of the subtitle (above the `#501 created 3 days ago by housseindjirdeh`)
2. Having the subtitle line to be `#501 created 3 days ago by housseindjirdeh` would be perfect

chinesedfan:
@machour Thanks for your invitation! I will join you if I have time this weekend.

@housseindjirdeh @lex111 Sorry, I don't very understand about "uI specifics". The date is not introduced in this PR. I agree that Github's message is clear and without any confusion, but we may have no enough space, especially for devices like iPhone5 or long user names like @housseindjirdeh. Can you accept the following UI or omit some words when it is too long?

lex111:
@chinesedfan I think it's necessary to change the layout: put the header and label in one block
And the information part after them to place the entire width. I showed it in the screenshot:
image

And let's probably move the number of comments to the end of the issue name:
image

machour:
@lex111 I suggested removing the "Open" badge and colorizing the icon on the left on Gitter, @chinesedfan is giving it a try

chinesedfan:
> And let's probably move the number of comments to the end of the issue name:

@lex111 I have a problem that we need add a padding between the title and the comments icon. But if the title is just one line, then the comments icon will start another line with a wired space.

housseindjirdeh:
@chinesedfan Here's my stab at it. Hope it helps :)

image
 
^ https://gist.github.com/housseindjirdeh/6953891ef108f2f284c2cecd2e372f05

chinesedfan:
@housseindjirdeh Thanks for your gist. In fact, we have tried this but find there will be a little wired if no comments. Because the title toggles between full width and partial width.

housseindjirdeh:
@chinesedfan as we chatted, we can just show the comments icon all the time even if there are 0 comments :)

chinesedfan:
@housseindjirdeh Updated. Now it looks the same with your screenshots.

src/components/issue-list-item.component.js
```diff
@@ -56,16 +66,23 @@ export const IssueListItem = ({ type, issue, navigation, language }: Props) =>
       <ListItem
         containerStyle={styles.listItemContainer}
         title={issue.title}
-        subtitle={`#${issue.number} - ${moment(issue.created_at).fromNow()}`}
+        subtitle={`#${issue.number} opened ${moment(
```
lex111: @chinesedfan All phrases must be moved to `locale` files.

And let's rename `opened` to `created`, because our icon shows an open or closed issue.

housseindjirdeh: As soon as you get to this @chinesedfan, we'll merge this in :)

chinesedfan:
@lex111 @housseindjirdeh I just made some changes:
- Different icon and message format for closed issues
- Make the subtitle font weight smaller, as #381 required(by the way, corresponding changes can be reverted in #409)

chinesedfan:
Ah, the title boldness is fixed, too.

src/components/issue-list-item.component.js
```diff
@@ -56,16 +75,29 @@ export const IssueListItem = ({ type, issue, navigation, language }: Props) =>
       <ListItem
         containerStyle={styles.listItemContainer}
         title={issue.title}
-        subtitle={`#${issue.number} - ${moment(issue.created_at).fromNow()}`}
+        subtitle={
+          issue.state === 'open'
+            ? `#${issue.number} opened ${moment(
```
housseindjirdeh: Could you please throw in two translation strings into our locale files and use them here as @lex111 mentioned :)

src/components/issue-list-item.component.js
chinesedfan: @lex111 @housseindjirdeh Done. I read so quickly that missed the first line about locale files. 🤣

PR #390, opened 2017-09-29
fix(events): get user events after login

lex111:
I have just noticed that after sign out the profile and when trying to log in, the events does not load, just a white screen, this PR fixes it.


machour:
@lex111 I noticed this behavior a few other times today, without signing in or out from the app.
Definitely something to keep an eye on before release

lex111:
@machour just after accepting this PR there will not be this error, because after login we do not have a user (auth user), we need to get it, and then it will come to the `componentWillReceiveProps` and it will load the events.

PR #398, opened 2017-10-01
style(notification-badge): Extend badge props + Refactor style

housseindjirdeh:
Update notifications icon badge:

1. Made it blue and text more prominent
2. Moved the `absolute` style logic to outside the badge component. If we plan on using this component elsewhere it shouldn't have absolute styling within
3. Extended props to include color and background color

CC @lex111 

screen shot 2017-09-30 at 11 29 28 pm

### DIFF
--- a/src/components/notification-icon.component.js
+++ b/src/components/notification-icon.component.js
@@ -1,10 +1,24 @@
 // @flow
 import React, { Component } from 'react';
 import { connect } from 'react-redux';
-import { View } from 'react-native';
+import { View, StyleSheet } from 'react-native';
 import { Icon } from 'react-native-elements';
 
 import { Badge } from 'components';
+
+import { colors } from 'config';
+
+const styles = StyleSheet.create({
+  icon: {
+    justifyContent: 'center',
+    alignItems: 'center',
+  },
+  badgeContainer: {
+    position: 'absolute',
+    right: 0,
+    top: 0,
+  },
+});
 
 const mapStateToProps = state => ({
   notificationsCount: state.notifications.notificationsCount,
@@ -13,7 +27,7 @@ const mapStateToProps = state => ({
 class NotificationIconComponent extends Component {
   props: {
     iconColor: string,
-    notificationsCount: string,
+    notificationsCount: number,
   };
 
   render() {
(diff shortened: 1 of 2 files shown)


lex111:
@housseindjirdeh could set `-1` for `top` and `right`, and also return `borderColor` and `borderWidth`? So it looks better to the badge in the tabbar.

src/components/notification-icon.component.js
```diff
@@ -1,19 +1,33 @@
...
 import { Icon } from 'react-native-elements';
 
 import { Badge } from 'components';
 
+import { colors } from 'config';
+
+const styles = StyleSheet.create({
+  icon: {
+    justifyContent: 'center',
+    alignItems: 'center',
```
lex111: Need to add `flex: 1`.

chinesedfan:
@housseindjirdeh I have never seen any blue notifications. I prefer to keep it as red. :)

src/components/notification-icon.component.js
```diff
@@ -1,19 +1,33 @@
...
-import { View } from 'react-native';
+import { View, StyleSheet } from 'react-native';
 import { Icon } from 'react-native-elements';
 
 import { Badge } from 'components';
 
+import { colors } from 'config';
+
+const styles = StyleSheet.create({
+  icon: {
```
lex111: This is still my fault, but we will not have alignment, without `flex: 1`, so it can move the styles to the `badgeContainer`? Though all the same the vertical alignment not absolutely on the center seems ..

housseindjirdeh:
Cheers thank you @chinesedfan @lex111. Thought blue would better match GitHub's current badge color but probably a bit head of myself :joy: 

image

Reverted back to dark red and added the border back.

image

Also used `-1` positioning like you mentioned @lex111 :dancer:

src/components/notification-icon.component.js
housseindjirdeh: You're right, adding `flex: 1` messes up positioning entirely so removed this style completely and it looks fine without :+1:

lex111:
@housseindjirdeh thanks! Strange border :confused: 
image

housseindjirdeh:
Left it to the same border settings you had previously @lex111 so not sure why if it looks weird to you  :(

Zoomed in side by side comparison of this PR update and what's currently on master:

image

lex111:
On Android, simply the color of the border is uniform:

image

housseindjirdeh:
Oh does it look more uniform on Android? Not sure why :thinking: 

Again we're zooming in for this, I don't think it looks weird normally on view:

image

lex111:
@housseindjirdeh I meant that iOS can see a red background around the border, compare the screenshots, I marked with an arrow. I do not understand why this is so, or is it in dev mode only?

housseindjirdeh:
Really not sure, I'm hoping this shows on simulator and not on the actual device. Will test it on my phone once we have it merged in. It is weird though I agree :thinking:

Issue #395, opened 2017-09-30
Find alternative to react-native-app-intro


andrewda:
It's very outdated and I'd be very surprised if it continues to work with future versions of React-Native, considering they're still using `PropTypes` from React.

andrewda:
It's one of the few things that's blocking tests at the moment.

housseindjirdeh:
@Ernoff was kind enough to create his forked package for this :dancer:

PR #399, opened 2017-10-01
fix(dependencies): make react-native-app-intro work with React 16

Ernoff:
Instead of finding alternative and changing the respective code, I just went along to help fix react-native-app-intro by including prop-types as needed thereby making it compatible with React 16.^

Kindly review and send feedback where needed.

#395


package.json
```diff
@@ -57,7 +57,7 @@
     "react": "16.0.0-alpha.12",
     "react-native": "0.48.1",
     "react-native-actionsheet": "^2.2.0",
-    "react-native-app-intro": "^1.1.5",
+    "react-native-app-intro": "https://github.com/Ernoff/react-native-app-intro.git#updated-packageJson",
```
andrewda: :clap: Though it would be sweet if you could publish on NPM if at all possible so we don't have too many GitHub repos in the `package.json`.

Ernoff: Okay Great!! 
 I will get down to it

housseindjirdeh:
Thanks a million @Ernoff, looks like you've addressed publishing to a package #400

PR #254, opened 2017-08-15
fix(repo): Show repo prefix for search repo result

reyhansofian:
Closes #252 

- [x] show repo prefix

# Preview

image

PR #409, opened 2017-10-02
fix(*): Style updates, add edit issue body functionality

housseindjirdeh:
Various fixes/updates for #381 

- Changes color of mark all notifications button

image

- Adds padding for no notifications message in notifications screen

image

- Fixes this problem:

image

- Allows editing of issue description

editdescription

- Makes title/subtitle of `IssueDescriptionListItem` better in terms of boldness

image

- Removes the user title from repository list item (@lex111 - with this it's unlikely we'll see the `private` box being squeezed anymore as most repo titles aren't nearly as long)

image

### DIFF
--- a/src/components/button.component.js
+++ b/src/components/button.component.js
@@ -9,6 +9,9 @@ const types = {
       borderColor: '#adadad',
       backgroundColor: '#e6e6e6',
     },
+    iconStyle: {
+      color: '#333333',
+    },
     textStyle: {
       color: '#333333',
     },
@@ -188,7 +191,12 @@ export class Button extends Component {
         }}
         disabledStyle={disabledStyle}
         icon={
-          icon && { ...defaultIconStyle, ...sizes[size].iconStyle, ...icon }
+          icon && {
+            ...defaultIconStyle,
+            ...types[type].iconStyle,
+            ...sizes[size].iconStyle,
+            ...icon,
+          }
         }
       />
     );
--- a/src/components/repository-list-item.component.js
+++ b/src/components/repository-list-item.component.js
@@ -61,7 +61,7 @@ const renderTitle = (repository, showFullName) =>
         {repository.private &&
           <View style={styles.privateIconContainer}>
             <Icon
-              size={20}
+              size={16}
               name="lock"
               type="octicon"
               color={colors.greyDarkest}
--- a/src/components/repository-list-item.component.js
+++ b/src/components/repository-list-item.component.js
@@ -7,6 +7,7 @@ import { colors, languageColors, fonts, normalize } from 'config';
 
 type Props = {
   repository: Object,
+  showFullName: boolean,
   navigation: Object,
 };
 
@@ -18,22 +19,14 @@ const styles = StyleSheet.create({
   },
   titleWrapper: {
     flexDirection: 'row',
+    alignItems: 'center',
   },
   title: {
     color: colors.primaryDark,
     ...fonts.fontPrimarySemiBold,
   },
-  private: {
-    borderColor: 'rgba(27, 31, 35, 0.15)',
-    borderRadius: 2,
-    borderStyle: 'solid',
-    borderWidth: 1,
-    color: colors.greyDark,
-    fontSize: normalize(10),
-    marginLeft: 8,
-    paddingLeft: 4,
-    paddingRight: 2,
-    paddingTop: 2,
+  privateIconContainer: {
+    marginLeft: 6,
   },
   description: {
     color: colors.primaryDark,
@@ -58,14 +51,22 @@ const styles = StyleSheet.create({
   },
 });
 
-const renderTitle = repository =>
+const renderTitle = (repository, showFullName) =>
   <View style={styles.wrapper}>
     <View style={styles.repositoryContainer}>
       <View style={styles.titleWrapper}>
         <Text style={styles.title}>
-          {repository.name}
+          {showFullName ? repository.full_name : repository.name}
         </Text>
-        {repository.private && <Text style={styles.private}>Private</Text>}
+        {repository.private &&
+          <View style={styles.privateIconContainer}>
+            <Icon
+              size={20}
+              name="lock"
+              type="octicon"
+              color={colors.greyDarkest}
+            />
+          </View>}
       </View>
       <Text style={styles.description}>
         {emojifyText(repository.description)}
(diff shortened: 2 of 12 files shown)


src/issue/issue.action.js
```diff
@@ -172,6 +173,28 @@ export const editIssueComment = (issueCommentId, owner, repoName, body) => {
   };
 };
 
+export const editIssueBody = (owner, repoName, issueNum, body) => {
```
housseindjirdeh: `editIssue` currently already exists and I'm definitely not a fan of how I'm passing `updateParams` to it. Because the comment is HTML, can't leverage that to pass in a new update parameter for the text to replace state. 

Will need to refactor how `editIssue` works later without a doubt and instead of passing in updateParams, just get what's returned from the API (and we can then just remove this)

lex111:
> Removes the user title from repository list item (@lex111 - with this it's unlikely we'll see the private box being squeezed anymore as most repo titles aren't nearly as long)

Let's remove this label and display this [icon](https://octicons.github.com/icon/lock/)?

lex111:
> Adds padding for no notifications message in notifications screen

And why are they needed? There in fact there is an alignment on the center, and after addition padding, any more will not be absolutely on the center to be displayed.

chinesedfan:
@housseindjirdeh User titles are needed, at least for searched result or stared repositories. This feature is added by #254 in particular. And I also mentioned the redundancy, but you missed it. :)

machour:
Seconding @lex111, let's use that emoji and bring back username/reponame :

If I'm on some user profile, then check his repos, I'll have to remember the user name, or navigate back and forth to have the repo full name to clone it

src/components/repository-list-item.component.js
```diff
@@ -63,7 +63,7 @@ const renderTitle = repository =>
     <View style={styles.repositoryContainer}>
       <View style={styles.titleWrapper}>
         <Text style={styles.title}>
-          {repository.full_name}
+          {repository.name}
```
lex111: As correctly noted, displaying only the name of the repository will not be convenient, so we need to add a check: if this is an auth screen, then show only the name of the repository, and if the user profile, then display the full name.

lex111: Although, maybe it can just add the `lock` icon to the end of the title? 
@chinesedfan so in fact will work?

chinesedfan: Yes, it doesn't matter the pending element is the private box or the lock icon, as long as the parent element is Text and the child view has width&height specified.

jouderianjr:
@housseindjirdeh What do you think about handling with edit issue body in another screen. Maybe we could create an `edit issue` for edit the description and the title.

src/components/repository-list-item.component.js
housseindjirdeh: Done

housseindjirdeh:
> And why are they needed? There in fact there is an alignment on the center, and after addition padding, any more will not be absolutely on the center to be displayed.

Honestly I just felt that the message looks a bit squeezed for English copy?

image

Probably just imagining things, reverted it back and removed that padding :)

> User titles are needed, at least for searched result or stared repositories. This feature is added by #254 in particular. And I also mentioned the redundancy, but you missed it. :)

Have no idea how I missed that point. Thank you, and I completely forgot that `search` and `starred` repos would have to show that. Updated it so now only the `authenticated` list of users repositories won't show (otherwise it will :))

> Let's remove this label and display this icon?

Done:

image

housseindjirdeh:
@jouderianjr so you make a good point, and right now if you head to the settings screen you can modify a few issue settings (but not title or description)

image

We could have issue title and description modifiable here but honestly I actually prefer at least the description being modified the same way as any other comment. It feels intuitive as I still think of it as a comment.

Definitely curious if you think there's a better option however, as well as how we can add functionality to edit the issue title in the future 🤔

lex111:
> Honestly I just felt that the message looks a bit squeezed for English copy?

Oops, did not notice if padding left and right, then this should not affect the alignment, although at first glance it seemed to me that this is so. Excuse me, let's it return, but can we make them smaller?

lex111:
> Let's remove this label and display this icon?

Do not you think that the size is big? In my opinion it was better if you reduce the size of the icon?

housseindjirdeh:
> Oops, did not notice if padding left and right, then this should not affect the alignment, although at first glance it seemed to me that this is so. Excuse me, let's it return, but can we make them smaller?

Cheers no worries. Added some padding but reduced it a bit :)

image

> Do not you think that the size is big? In my opinion it was better if you reduce the size of the icon?

And nice you're probably right. Reduced it:

image

machour:
@housseindjirdeh I'd rather use type="default" (or no type) for the "Mark all as read" button.
Let's keep "success" for **strong** actions, like merge

housseindjirdeh:
> @housseindjirdeh I'd rather use type="default" (or no type) for the "Mark all as read" button.
Let's keep "success" for strong actions, like merge

Done.

Although the default button looks interesting 🤔 Are these the intended colors? (just showing it defaulted for the merge button for demo, don't have any notifications in the moment that's why :))

image

machour:
@housseindjirdeh that icon should be black, otherwise it's the intended look.

jouderianjr:
> We could have issue title and description modifiable here but honestly I actually prefer at least the description being modified the same way as any other comment. It feels intuitive as I still think of it as a comment.
> 
> Definitely curious if you think there's a better option however, as well as how we can add functionality to edit the issue title in the future 🤔

I got it, make sense, so we can release that in v1.3 and then reorganize this screen in order to fit edit title and body 😄 

Nice job mate!

src/components/button.component.js
```diff
@@ -188,7 +191,12 @@ export class Button extends Component {
         }}
         disabledStyle={disabledStyle}
         icon={
-          icon && { ...defaultIconStyle, ...sizes[size].iconStyle, ...icon }
+          icon && {
+            ...defaultIconStyle,
```
housseindjirdeh: Extended this since didn't seem like icons were being styled in the component within @machour

housseindjirdeh:
> @housseindjirdeh that icon should be black, otherwise it's the intended look.

@machour fixed:

image

housseindjirdeh:
@jouderianjr Exactly what I was thinking :)

PR #410, opened 2017-10-02
Clean up braces

Ernoff:
Braces that were present as a result of keyboard malfunction has been cleaned up


Ernoff:
Possibly. I had to disable every installed extension in my VSCode until I ascertain and eliminate the root cause

PR #414, opened 2017-10-02
refactor(profiles): show user data when it is fully loaded

lex111:
This is a refactoring of refactoring :laughing:  Second attempt #366
In fact, in the past PR did not take into account the `auth profile`, and it was so that the block with the stars was displayed when there were none.

I need someone to check it, I need to know whether it's good now.

### DIFF
--- a/src/auth/screens/auth-profile.screen.js
+++ b/src/auth/screens/auth-profile.screen.js
@@ -121,9 +121,9 @@ class AuthProfile extends Component {
           renderContent={() =>
             <UserProfile
               type="user"
-              initialUser={hasInitialUser ? user : {}}
-              user={hasInitialUser ? user : {}}
-              starCount={hasInitialUser ? starCount : ''}
+              initialUser={hasInitialUser && !!starCount ? user : {}}
+              user={hasInitialUser && !!starCount ? user : {}}
+              starCount={hasInitialUser && !!starCount ? starCount : ''}
               language={language}
               navigation={navigation}
             />}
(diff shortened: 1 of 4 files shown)


src/components/user-profile.component.js
```diff
@@ -124,6 +125,7 @@ export const UserProfile = ({
       </Text>
     </View>
     <View style={styles.details}>
+      {user.hasOwnProperty('public_repos') &&
```
housseindjirdeh: Can we just do `user.public_repos` here?

src/auth/screens/auth-profile.screen.js
```diff
@@ -121,9 +121,9 @@ class AuthProfile extends Component {
           renderContent={() =>
             <UserProfile
               type="user"
-              initialUser={hasInitialUser ? user : {}}
-              user={hasInitialUser ? user : {}}
-              starCount={hasInitialUser ? starCount : ''}
+              initialUser={hasInitialUser && !!starCount ? user : {}}
```
housseindjirdeh: We have a lot of double negations in out app and I think we may be better to start leveraging `lodash` properly. Let's discuss this after our release however.

src/components/user-profile.component.js
lex111: There may be zero repositories, and then nothing will be shown, right?

housseindjirdeh: Yeah that's true 🤔 

Would again prefer using `_.has()` from lodash but let's discuss that later if we decide to include the whole library

PR #425, opened 2017-10-04
feat(issue): Disable merge button if not possible

### LINKED ISSUE #128: Don't show merge button if merge is not possible
Not sure how much of this data is available from the API, but the green merge button is still shown even if a PR is unable to be merged (due to conflicts or required tests not passing, for example). My suggestion would be to grey-out the merge button with a tooltip explaining why it's not mergeable.

chinesedfan:
Fixes #128.

The main idea is to leverage the field `mergeable` provided by [the pull request API](https://developer.github.com/v3/pulls/#get-a-single-pull-request). Besides, I refactor a little about the logic of fetching a issue. That's because I want to only fetch pull request if the issue is a pull request. At last, I find out some basic fields are still needed by using the issue API, so I give up the API replacement but keep the refactor.

### DIFF
--- a/src/components/issue-description.component.js
+++ b/src/components/issue-description.component.js
@@ -73,6 +73,7 @@ export class IssueDescription extends Component {
   props: {
     issue: Object,
     diff: string,
+    isMergeable: boolean,
     isMerged: boolean,
     isPendingDiff: boolean,
     isPendingCheckMerge: boolean,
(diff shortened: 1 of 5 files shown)


chinesedfan:
@housseindjirdeh The required field `mergeable` only exists in the actual pull request API.

PR #427, opened 2017-10-04
fix(markdown_style): Escape content string

jouderianjr:


housseindjirdeh:
No worries, thank folks will get this out as soon as possible 🙌

Issue #423, opened 2017-10-04
Edit Comment TextInput not rendering with fixed height on Android


housseindjirdeh:
On Android:

![image](https://user-images.githubusercontent.com/12476932/31159169-27861fb6-a895-11e7-9221-8b27fed9107d.png)

Not a huge deal and didn't want it to block release of course. CC @jouderianjr, I can look into this in the next few days :)

PR #428, opened 2017-10-04
fix(editissuecomment, textinput): Fix height rendering for Android

housseindjirdeh:
For #423 

`onContentSizeChange` doesn't seem to fire when the component mounts @jouderianjr so adding a height property seems to fix it

image


housseindjirdeh:
Going to merge this in as part of `1.3.1` (please feel free to comment if I should have done it differently however)

PR #435, opened 2017-10-06
fix(android): Hide tabbar when keyboard show

lex111:
This problem is only on Android, looked at the iOS-simulator, the tabbar is hidden there, but on Android no.


lex111:
@housseindjirdeh the app needs to be updated on Google Play.

PR #407, opened 2017-10-02
test: begin implementing basic component tests

andrewda:
Re-opening #148 as a branch of this main repo so it's easier for people to contribute to these tests while they're being developed.

Fixes #112.

PR #258, opened 2017-08-15
feat(commit): view commits

Antoine38660:
Fix #5 

- [x] create a commit screen
  - [ ] have the commiter name clickable (to open a profile)
- [x] have commits that show in the notifications screen be clickable
- [x] add number of commits in the repository screen - also clickable
- [x] have the number of commits when viewing the diff of a pull request be clickable

PR #438, opened 2017-10-06
feat: Issue Events

### LINKED ISSUE #212: Show issue events
Hey guys,

What about adding the issue events and pr events to gitpoint?

img_0494

So anybody knows who added the labels und requested the views and stuff.

Found this area in the api docs, that should be what is needed: https://developer.github.com/v3/issues/events/

brandly:
fixes #212. i've already [posted some updates](https://github.com/gitpoint/git-point/issues/212#issuecomment-334374537), but i'm close to finishing things up now. here's a screenshot:

screen shot 2017-10-06 at 4 01 06 pm

i made [a fake PR](https://github.com/brandly/git-point/pull/1) to render a bunch of events:

screen shot 2017-10-06 at 4 00 23 pm


brandly:
TODO:

- [ ] handle remaining events. [several are still unimplemented](https://github.com/gitpoint/git-point/pull/438/files?w=1#diff-291ee7c935dba219a987bfb7d7ff604bR96), but i need to find examples of those to test against.
- [ ] [combine simultaneous label changes into one event](https://github.com/gitpoint/git-point/issues/212#issuecomment-334680150). github's UI can get kinda crazy if you add/remove a lot of labels at once tho:
screen shot 2017-10-06 at 4 20 13 pm
- [ ] potentially add links to users/commits. i'm not sure how to implement this. seems like links use `TouchableOpacity` with is a `View` so it can't descend from a `Text` element. can i create links that embed in `Text` somehow?

lex111:
>  handle remaining events. several are still unimplemented, but i need to find examples of those to test against.

Of course they are [a lot of them](https://developer.github.com/v3/issues/events/#events-1), but you need to add support for everyone. We can create a test issue, and there apply different actions to show all the events for you.

>  combine simultaneous label changes into one event. github's UI can get kinda crazy if you add/remove a lot of labels at once tho:

And yet, it's better to show it the way it's presented on GitHub. You can do this, as I understand it, by comparing the date of the event?

> potentially add links to users/commits. i'm not sure how to implement this. seems like links use TouchableOpacity with is a View so it can't descend from a Text element. can i create links that embed in Text somehow?

This is not difficult, as an example you can see how it was done on the event screen. An example for the repository [cloning event](https://github.com/gitpoint/git-point/blob/master/src/auth/screens/events.screen.js#L249) is going to the repository screen.

src/components/issue-event-list-item.component.js
```diff
@@ -0,0 +1,520 @@
...
+  };
+
+  render() {
+    const { actor, milestone, created_at: createdAt } = this.props.event;
+    const action = this.props.demilestoned
+      ? 'removed this from'
+      : 'added this to';
+
+    return (
+      <View style={[styles.container, styles.header]}>
```
housseindjirdeh: Do you think we need to have all of these separate components for each specific use case? Will it be possible to have 2 or 3 and then with props, be able to extend their `icon`, `user` and `descriptions` for example?

brandly: yeah, i see exactly what you're saying. i was handling `actor`s poorly at first, so i wasn't quite sure what the more abstract component would look like yet. it makes sense now tho

housseindjirdeh: No worries, always easier to start laying down and then refactoring it slightly :)

Cheers, otherwise I can't see much I would change. This is really coming together nicely mate.

housseindjirdeh: Oh, and don't forget to add yourself as a contributor with `yarn contributors:add` :)

brandly:
thanks!! i've been using this app for a while on my phone, so i'm glad to help out. i'm really impressed at how everything's put together here.

> With regards to combining multiple label changes into one ListItem...

most of the time, it looks fine, but when there are a lot of label changes, grouping the labels is a big improvement. i started implementing this already. 

the logic for grouping label events together is working, but i'm still figuring out some of the styling. i created a new component for rending labels inline, but when i render a `<Text>`, it seems like `margin`/`padding` styles aren't respected. 

```js
export class InlineLabel extends Component {
  props: {
    label: Object,
  }

  render() {
    const { color, name } = this.props.label;

    return (
      <Text
        style={{
          backgroundColor: '#' + color,
          color: getFontColorByBackground(color),
          // Doesn't affect anything??
          padding: 10,
          margin: 10,
        }}
      >
        {name}
      </Text>
    );
  }
}
```

^ that component nested in some other `<Text>` ends up rendering this:

screen shot 2017-10-06 at 6 47 09 pm

is this the wrong approach? we could also postpone this for a later PR.

> Yep including links are relatively straightforward

that makes sense! is the `navigation` object exclusively provided to screens? if so, do you think it makes more sense to have `onUserPress` and `onCommitPress` callbacks passed down, or should i pass `navigation` down into the depths?

thanks for the feedback 🙏

brandly:
for the remaining unimplemented events, i'm having trouble getting my hands on the data i need to render what github shows in their UIs. 

for example, if you [hit the events api](https://api.github.com/repos/brandly/git-point/issues/1/events), there's an `added_to_project` event. however, there's no information about _which_ project it was added to. there's a `url` field, but following it [takes you here](https://api.github.com/repos/brandly/git-point/issues/events/1282488504) which still doesn't tell you!

i implemented `marked_as_duplicate`, but once again, i don't know _which_ issue it's a duplicate of. do you know how to handle this? should we just ignore certain events for now?

---

@housseindjirdeh i created an `<Event />` component to handle the base template, and then used that component to render everything. i still have quite a few individual components, however, since there are several "pairs" of events that very similar but often have slight wording or icon differences. do you like this approach? is there a slicker way to handle things?

---

with regards to (un)labeled events, in rare situations of changing several labels, it sure takes up a lot of vertical space:

screen shot 2017-10-07 at 12 10 12 pm

as mentioned in [my previous comment](https://github.com/gitpoint/git-point/pull/438#issuecomment-334903197), i'm not sure how to inline things and make them look nice. using `<LabelButton />`, as things are currently implemented, labels with obnoxiously long names don't render well:

screen shot 2017-10-07 at 12 11 49 pm

so i'm okay with postponing grouping (un)labeled events into a single row, but the current implementation has some shortcomings.

---

i'm excited to see these in the app! let me know what you think of all this!

brandly:
got the labels working 🎃 

screen shot 2017-10-07 at 6 24 23 pm

seems as tho if you have this:

```js
<Text>
  <Text></Text>
  <Text></Text>
  <Text></Text>
</Text>
```

the inner `Text`s no longer behave like a box. switching it to this allowed for `padding`/`margin`:

```js
<View>
  <Text></Text>
  <Text></Text>
  <Text></Text>
</View>
```

now, it looks like words only break/wrap between each `Text`. in more common situations, you'll get something like this, so i'm pretty happy with it:

screen shot 2017-10-07 at 6 32 21 pm

i'm under the impression that i can't put a `borderRadius` on `Text`. is there a way around this?

src/components/label-button.component.js
```diff
@@ -38,6 +38,31 @@ const styles = StyleSheet.create({
   },
 });
 
+export class InlineLabel extends Component {
```
andrewda: This might deserve its own file. We usually only create 1 component per file.

src/issue/screens/issue.screen.js
```diff
@@ -10,13 +10,14 @@ import {
   Platform,
 } from 'react-native';
 import { Icon } from 'react-native-elements';
-
+import moment from 'moment/min/moment-with-locales.min';
```
andrewda: Add a blank line after the moment import so we can separate external imports from local components.

src/issue/screens/issue.screen.js
```diff
@@ -306,6 +339,63 @@ class Issue extends Component {
...
 
+    const eventsToRender = fullEvents
+      .filter(({ event }) => event !== 'mentioned' && event !== 'subscribed')
+      .filter(({ event }, index, list) => {
+        if (index === 0) {
+          return true;
+        }
+
+        // Merge events are always followed by a closed event, but we don't
+        // want to render them.
```
andrewda: Good thinking 👍

src/issue/screens/issue.screen.js
```diff
@@ -306,6 +339,63 @@ class Issue extends Component {
...
+            labelGroup.unlabeled.push(event);
+          }
+
+          results.push(labelGroup);
+        } else {
+          results.push(event);
+        }
+
+        return results;
+      }, []);
```
andrewda: Some of this logic might be able to be moved out into a `src/utils` file so we can test it easier, but it's not the end of the world if that doesn't happen right now.

src/components/label-button.component.js
brandly: that's fair. i mainly wanted to share the `getFontColorByBackground` function. is there an ideal location i should move that two so i can share it between components?

src/issue/screens/issue.screen.js
brandly: yeah, this would be good to test. i don't see much testing going on currently. am i overlooking it?

andrewda:
#407 is where the tests are being added

src/issue/screens/issue.screen.js
andrewda: #407 is where the tests are being added

src/components/label-button.component.js
andrewda: Ooh, `getFontColorByBackground` might be a good thing to move to `src/utils` so we can access it from everywhere consistently!

src/issue/screens/issue.screen.js
```diff
@@ -55,10 +60,21 @@ const mapDispatchToProps = dispatch =>
...
+const compareCreatedAt = (a, b) => {
+  if (a.created_at < b.created_at) {
+    return -1;
+  } else if (a.created_at > b.created_at) {
+    return 1;
+  }
+
+  return 0;
+};
+
```
chinesedfan: How about `(a, b) => a.created_at - b.created_at`? So that we can omit this function.

src/api/index.js
```diff
@@ -379,3 +379,14 @@ export const fetchNotificationsCount = accessToken =>
 
 export const fetchRepoNotificationsCount = (owner, repoName, accessToken) =>
   v3.count(`/repos/${owner}/${repoName}/notifications?per_page=1`, accessToken);
+
+export const fetchIssueEvents = (
+  owner: string,
+  repoName: string,
+  issueNum: number,
+  accessToken: string
```
chinesedfan: Types make function declaration clearer. But in other functions we never write them.

src/issue/screens/issue.screen.js
brandly: i think `created_at` is a date string, so i don't think we can `-` them.

chinesedfan: @brandly Yes, my mistake. 🙈

brandly:
figured out the styling issues. labels are looking nice now, and that component lives in its own file. here are some samples:

screen shot 2017-10-07 at 10 30 32 pm

screen shot 2017-10-07 at 10 33 39 pm

screen shot 2017-10-07 at 10 34 26 pm

housseindjirdeh:
> thanks!! i've been using this app for a while on my phone, so i'm glad to help out. i'm really impressed at how everything's put together here.

Means a lot mate, thank you <3

> most of the time, it looks fine, but when there are a lot of label changes, grouping the labels is a big improvement. i started implementing this already.
> 
> the logic for grouping label events together is working, but i'm still figuring out some of the styling. i created a new component for rending labels inline, but when i render a <Text>, it seems like margin/padding styles aren't respected.

Nice, looking at that screenshot it definitely looks like grouping them together looks a lot nicer! And yep you got it, wrapping in a `View` makes it easier to add inner padding and outer margin :)

> that makes sense! is the navigation object exclusively provided to screens? if so, do you think it makes more sense to have onUserPress and onCommitPress callbacks passed down, or should i pass navigation down into the depths?

Hmm if we're talking low level components - it usually makes sense to have callbacks passed upwards so the parent can take care of navigating. Because you have specific actions `onUserPress` and `onCommitPress` rather than a typical button `onPress` -- then I think it's just fine to pass navigation down as these components are geared towards one thing only

> @housseindjirdeh i created an <Event /> component to handle the base template, and then used that component to render everything. i still have quite a few individual components, however, since there are several "pairs" of events that very similar but often have slight wording or icon differences. do you like this approach? is there a slicker way to handle things?

Yep I was just thinking I really don't know if it's possible to have a single component to take care of _all_ cases. We probably can if we extend more props but not sure if that makes things cleaner or not. Will go through your PR first thing tomorrow morning and let you know what I think :)

> figured out the styling issues. labels are looking nice now, and that component lives in its own file. here are some samples:

Labels are looking beautiful mate. Appreciate you taking the time to try and make them inline work nicely :dancer: 

Super excited to get this in the app soon :pray:

.all-contributorsrc
```diff
@@ -384,6 +384,13 @@
       ]
     },
     {
+      "login": "brandly",
+      "name": "Matthew Brandly",
+      "avatar_url": "https://avatars3.githubusercontent.com/u/820696?v=4",
```

src/issue/screens/issue.screen.js
housseindjirdeh: Agreed there's quite a bit of logic here that I would like to see tested (once we have our initial tests added). Definitely something we can do a bit later however.

brandly:
you can press usernames now! it looks like [commit pages are currently getting built](https://github.com/gitpoint/git-point/pull/258), so i'm under impression we can't make the `commit_id`s pressable yet.

besides that, i think the only real shortcoming here is that i can't figure out how to get the necessary data for certain events, [as described in an earlier comment](https://github.com/gitpoint/git-point/pull/438#issuecomment-334946805).

thanks for all the quick feedback! makes it easy to get involved here for the first time.

src/issue/screens/issue.screen.js
```diff
@@ -293,11 +324,14 @@ class Issue extends Component {
       isPendingIssue ||
       isDeletingComment
     );
-    const isShowLoadingContainer = isPendingComments || isPendingIssue;
-    const fullComments = !isPendingComments ? [issue, ...comments] : [];
+    const isShowLoadingContainer =
+      isPendingComments || isPendingIssue || isPendingEvents;
+    const fullEvents = !isPendingComments
```
housseindjirdeh: Since `fullEvents` is essentially everything (issue description, comments and events), maybe we can find a better name for this. `fullConversation` or something? (Not sure if that's any better however :joy: :sob:)

src/utils/color-helpers.js
```diff
@@ -0,0 +1,10 @@
+import { colors } from 'config';
+
+export const getFontColorByBackground = bgColor => {
+  const r = parseInt(bgColor.substr(0, 2), 16);
```
housseindjirdeh: :tada:

src/components/issue-event-list-item.component.js
```diff
@@ -0,0 +1,617 @@
...
+            {action} the <Bold>{milestone.title}</Bold> milestone
+          </Text>
+        }
+        createdAt={createdAt}
+      />
+    );
+  }
+}
+
+class Locked extends Component {
```
housseindjirdeh: I like how we have a base `Event` component that's being used for other scenarios, but still think we can probably simplify things here a bit. Instead of having each component here take in the event and decide which icon to use, etc.., we can probably extend this with props. For example: 

```
class EventInstance extends Component {
  props: {
    icon: string,
    iconColor: color,
    createdAt: string,
    textChildren: React.Element<*>,
  };

  render() {
    const {
      name,
      iconColor = '#586069',
      iconBackgroundColor = '#E6EBF1',
      createdAt,
      textChildren
    } = this.props;
    const action = this.props.unassigned ? 'unassigned' : 'assigned';

    return (
      <Event
        icon={
          <Icon
            iconStyle={{
              marginLeft: marginLeftForIconName(name),
              marginTop: 1,
              color: iconColor,
            }}
            containerStyle={[styles.iconContainer, { iconBackgroundColor }]}
            name={icon}
            type="octicon"
            size={16}
          />
        }
        text={
          <Text style={{ padding: 3 }}>
            {textChildren}
          </Text>
        }
        createdAt={createdAt}
      />
    );
  }
}
```

If you think it makes sense to have a `Event` component specified on it's own, then we can probably use something like this `EventInstance` everywhere instead of having `Locked`, `Headref`, etc.. which are essentially the same components with slight variations. Please let me know what you think.

housseindjirdeh: Notice how I'm just referencing `Icon` directly. Don't know if it's worth creating an `EventIcon` if we use a component like the above everywhere in my opinion as it'll only be used in one place (this component)

src/components/issue-event-list-item.component.js
```diff
@@ -0,0 +1,617 @@
...
+      // case 'converted_note_to_issue':
+      default:
+        return null;
+    }
+  }
+}
+
+class Event extends Component {
+  props: {
+    icon: Object, // TODO: enforce react instance
```
housseindjirdeh: `React.Element<*>` should do the trick :tada:

src/components/issue-event-list-item.component.js
```diff
@@ -0,0 +1,617 @@
...
+      <View style={styles.dateContainer}>
+        <Text style={styles.date}>
+          {moment(date).fromNow()}
+        </Text>
+      </View>
+    );
+  }
+}
+
+class Bold extends Component {
```
housseindjirdeh: For `Date` and `Bold`, if you consider my suggestion to begin to use `EventInstance` or a component name similarly everywhere and if these components aren't being used more than once, we can probably remove their isolations. Otherwise this is fine :+1:

src/components/issue-event-list-item.component.js
```diff
@@ -0,0 +1,617 @@
...
+                  actor={event.actor}
+                  onPress={this.onPressUser}
+                />{' '}
+                merged <Bold>{event.commit_id.slice(0, 7)}</Bold>
+              </Text>
+            }
+            createdAt={event.created_at}
+          />
+        );
+      // case 'referenced':
```
housseindjirdeh: Are the commented out cases conditions you just can't test? Noticing this when I view this issue #439 

image

Not sure why this is happening however :thinking:

brandly: yeah, [the most recent commit](https://github.com/gitpoint/git-point/pull/438/commits/5f80eddf834f2a87c1e4392c3fa2b5838a1e5900#diff-291ee7c935dba219a987bfb7d7ff604bL194) should've removed that. i had that in for development purposes.

src/issue/screens/issue.screen.js
brandly: agreed! i hesitated when i wrote that name but couldn't come up with anything better. `fullConversation` seems like the right move. i don't see it used in `git-point` yet, but it's used on github:

screen shot 2017-10-08 at 3 58 44 pm

src/components/issue-event-list-item.component.js
brandly: i think i see what you're saying. i'm happy to rearrange things, but i want to make my case first 😄. i'm not trying to argue or to even say that things are ideal how they're currently implemented, but i want to understand more of the _why_ you'd prefer things to be structured differently.

the more `react` i've written, the more i've tended to break things apart into a bunch of small components. these components can be composed together to build larger structures.

you seem to be somewhat against tiny components like `Bold`. in several different places surrounding inline text, i could write `<Text style={styles.boldText}>` or i could pull that out into `<Bold>`, allowing me to write nice things like

```js
changed the title from <Bold>{event.rename.from.trim()}</Bold>{' '}
to <Bold>{event.rename.to.trim()}</Bold>
```

isn't this encapsulation the point of react?

similarly, i _could_ get rid of `<EventIcon>`, but even if it's only in like 2 or 3 spots, i'd always be setting the same `name`, `size`, and `containerStyle` on an `Icon`. why not pull it out?

as a project grows, i tend to build up a library of these small components. even if the small components are relatively context specific, refactoring tends to feel easy, cause i can often just build a new component using a few existing building blocks and be done with it.

y'all have invested way more into this project than i have, so i'm happy to structure things as you'd like. it just feels as tho i'd end up with more duplication and larger, more complex `render` functions, neither of which i'm a big fan of. however, i could just be misunderstanding you or missing something altogether.

thanks again!!!

housseindjirdeh: Oh no please don't ever hesitate to let us know what you're thinking mate :)

Yeah everyone prefers to design their component renders in different fashions. I just happen to lean towards creating sub-components _if_ I feel like they could live as separate entity in the components directory and used throughout the app. But you're right, there's definitely benefit to break things down within the same component file and not have large render methods spread out across the app.

Definitely see where you're coming from trying to arrange things the way I suggested in causing large, duplicated logic. Perfectly fine with the way you've set things up if you feel like this is a nicer way to structure things in the file :+1:

brandly: i refactored some things in the direction you suggested, and i really like how things ended up! 

`<Event>` became a bit more powerful, and i was able to eliminate most of the smaller components. `<LabelGroup>` has to make quite a few decisions, so it held on. `<Bold>` and `<ActorLink>` still feel useful, so i kept them around too.

things feel more simple! let me know what you think 💯

src/issue/screens/issue.screen.js
brandly: i [took care of this naming](https://github.com/gitpoint/git-point/pull/438/commits/0bb54159e3abe39277524de9a004646f58b11f6c), while extracting that chunk of logic into `utils/` for future testing.

housseindjirdeh:
@brandly just wanted to check in - are you still adding commits to this or do you think it's in a state where we can merge in? If it's the latter, will pull it down again, give it one more run through and if all is good we can merge this baby in

brandly:
@housseindjirdeh hey, i think things are stable! [this comment got buried](https://github.com/gitpoint/git-point/pull/438#discussion_r143624069). 

let me know what you think of the current state of things! i'm pretty happy with it.

housseindjirdeh:
Oops that did get buried, I'm sorry for missing it.

Going through the codebase and it looks beautiful mate, appreciate you taking some of my considerations into account 🎉 . Going to pull this down and test it one more time tomorrow and keep you posted 🙌

brandly:
i was away from my computer for a few days, so i'm excited to come back to a merged PR! thanks again

Issue #289, opened 2017-09-05
Make the issue page height be the size of the visible view


CameronSamuels:
When I open the keyboard to type an issue title/comment, I can’t scroll to submit (I have to click off). But this is mainly needed to be changed because when the “issue comment” is too much text, you can’t see what you are typing. This should also be fixed because some people might not know they need to click to hide it.

PR #447, opened 2017-10-07
Show new issue submit btn when keyboard is open

acurat:
Fixes https://github.com/gitpoint/git-point/issues/289 

iOS screenshot

simulator screen shot - iphone 6 - 2017-10-07 at 12 24 34

Android screenshot

screenshot_1507399484


src/issue/screens/new-issue.screen.js
```diff
@@ -101,77 +109,91 @@ class NewIssue extends Component {
...
-                { height: Math.max(60, this.state.issueCommentHeight) },
-              ]}
-              value={issueComment}
-            />
-          </SectionList>
-
-          <SectionList>
-            <View style={styles.listItemContainer}>
+        <KeyboardAvoidingView
+          style={{ flex: 1 }}
```
lex111: Style need move to `styles`.

src/issue/screens/new-issue.screen.js
```diff
@@ -101,77 +109,91 @@ class NewIssue extends Component {
...
+          behavior={'padding'}
+          keyboardVerticalOffset={Platform.select({
+            ios: 65,
```
lex111: Why are the different values depending on the platform?

acurat: The keyboardVerticalOffset is different because the height of the following vary by platform.
- React navigation headers.
- Height of the auto suggestion bar.
So I figured both platforms needs separate offset heights.

src/issue/screens/new-issue.screen.js
acurat: Done.

andrewda:
@acurat What if the user is using a custom keyboard that isn't the default system size? GBoard for iOS, for example, is at least another 50px larger than the system default. There's probably a more fluid way to do this.

acurat:
Here is a screenshot with the GBoard keyboard. It looks fine.

simulator screen shot - iphone 6 - 2017-10-07 at 12 24 34

Now, that I re-read the KeyboardAvoidingView's documentation, it makes sense. Need to only offset the react navigation header, the height of which is fixed in this app.

_keyboardVerticalOffset: number 

This is the distance between the top of the user screen and the react native view, may be non-zero in some use cases._

Anyways, let me know if there is a better way to do this.

lex111:
@acurat it does not work for me:

image

housseindjirdeh:
Friendly ping @acurat, just wanted to check in and see if you need any help with this at all :)

acurat:
@housseindjirdeh Yes, I am not sure I know a better solution for this. Any help is appreciated.

machour:
Seems like React Native 0.55 will ship with something interesting for this PR:
  https://twitter.com/brunolemos/status/977037390429777920

sergggggggggg (approved):

Issue #421, opened 2017-10-03
Add Galician translation


siderio2:
I would like to see a [Galician](https://en.wikipedia.org/wiki/Galician_language) (my native language) version of the app.

siderio2:
I would love to contribute with this translation, please.

machour:
Hey, thanks you so much @siderio2.
Here's how translations are added in the app: https://github.com/gitpoint/git-point/blob/master/CONTRIBUTING.md#translations

And don't forget to add yourself as a contributor! 🙌

siderio2:
Yes, I already watched at it and will also go to gitter in case I found any problem. Thanks.

housseindjirdeh:
We would love to have Galician support, thank you so much @siderio2 <3

siderio2:
I'm almost done with the translation, but looking at the other files I must change, I have found that we are using  emojicode flags, but there's no emojicode for [Galicia...](https://en.wikipedia.org/wiki/Galicia_(Spain))

lex111:
@siderio2 how well will it be if we use the Spanish flag (for it there is emoji)? Or can we take the Galician flag and stylize it under the emoji flag so that they look similar like emoji flags?

cc @housseindjirdeh

housseindjirdeh:
@siderio2 THANK YOU so much mate 🎉 🎉 

It's unfortunate that autonomous regions don't have their own emojis 😞 

I'm perfectly fine with just using the Spanish flag if that's okay? @siderio2. If not, more than happy to rewrite our logic slightly and show circular flag images on the left hand side like this :)

![image](https://rlv.zcache.com/galicia_spain_flag_classic_round_sticker-r79c390697ded48dd83dfaee0d97a4626_v9waf_8byvr_630.jpg?view_padding=%5B285%2C0%2C285%2C0%5D)

Happy to take this on as a separate ticket however (think that would be easier than trying to stylize images to emojis @lex111). Either option works well with me 👍

lex111:
@housseindjirdeh Why circular flag? Then we will have different flags in style, you can make the same, by analogy with emoji, right?

siderio2:
[...]

PR #452, opened 2017-10-08
feat(translation): Add Galician translation

siderio2:
Add ga.js file to .../languages, plus modify some related files

BREAKING CHANGE: New translation

#421


siderio2:
Two things:
1. I have finally used the white flag emoji while the flag system is not changed, but I'm not sure if I'm using it right: all flags included until now have a shortcode like ':flag-pt:', but I have seen that for the white flag it is  ':flag_white:'. Does it work??
2. And I'm asking this because I'm not able to run the app locally on my computer, so it would be nice if someone can check it. I'm going to gitter to ask what am I doing wrong with the local deployment.

lex111:
> I have finally used the white flag emoji while the flag system is not changed, but I'm not sure if I'm using it right: all flags included until now have a shortcode like ':flag-pt:', but I have seen that for the white flag it is ':flag_white:'. Does it work??

In fact, it is not yet clear what to do with the flags. But in any case, it does not affect your translation in any way, so do not worry about it yet.

> And I'm asking this because I'm not able to run the app locally on my computer, so it would be nice if someone can check it. I'm going to gitter to ask what am I doing wrong with the local deployment.

Of course, let's try to figure out what's wrong.

src/auth/screens/language-settings.js
```diff
@@ -34,4 +34,9 @@ export default [
     emojiCode: ':flag-ru:',
     name: 'Русский',
   },
+  {
+    code: 'gl',
+    emojiCode: ':flag_white:',
```
housseindjirdeh: The list of emoji codes we can use are [here](). You're right, `flag_white` doesn't exist but you can use `:waving_white_flag:` to have this :white_flag:

housseindjirdeh:
@lex111 feel free to do a pass through, other than changing the emoji code to `:waving_white_flag:` it looks good to me.

@siderio2 we would love to help you get the app working on your machine as well. Please let us know whatever issue you might be seeing and we'll help :raised_hands:

siderio2:
Well, it took me more than expected, but I finally managed to run the app on my machine. I already changed the flag shortcode, but it doesn't work on my phone (Redmi note 4 with android 6.0). And also I see a string not translated in options: 'Open in Browser', but I can't find it...

housseindjirdeh:
Woo awesome glad to hear you got it working mate :tada: 

Yeah so the `Open in Browser` string was merged just recently so you'll have to rebase/merge from master to see the line in `en.js`. And hmm interesting how the icon doesn't show. Don't worry about that, @lex111 and I are already discussing replacing all those emojicodes with circular images anyway :)

ios/GitPoint.xcodeproj/project.pbxproj
```diff
@@ -71,6 +70,7 @@
 		DC5C35ED1E37AD1800F3F526 /* SimpleLineIcons.ttf in Resources */ = {isa = PBXBuildFile; fileRef = DC5C35E31E37AD1800F3F526 /* SimpleLineIcons.ttf */; };
 		DC5C35EE1E37AD1800F3F526 /* Zocial.ttf in Resources */ = {isa = PBXBuildFile; fileRef = DC5C35E41E37AD1800F3F526 /* Zocial.ttf */; };
 		EB0F176BE52B4561B9FF07AB /* libReactNativeConfig.a in Frameworks */ = {isa = PBXBuildFile; fileRef = D7043BDB577E427391ECFBB0 /* libReactNativeConfig.a */; };
+		CBC20ECDD7FF46A49DF646D6 /* Feather.ttf in Resources */ = {isa = PBXBuildFile; fileRef = 20A722EC74A74B50BD5DCCC9 /* Feather.ttf */; };
```
housseindjirdeh: Have noticed this locally when linking (I'm assuming RNE has a new icon font called `Feather`?) Not sure why I just didn't push it to master but perfectly okay to have it in this PR

Issue #141, opened 2017-07-21
GraphQL


alejandronanez:
We should start moving gradually towards GraphQL, it will save us a TON of requests. AFAIK GraphQL doesn't have available the same exact endpoints as Github V3 Rest API so we should have a mix between both.

andrewda:
While it's fresh on my mind, here's how we'd fetch an issue:

```
{
  repository(owner: "gitpoint", name: "git-point") {
    issue(number: 2) {
      title
      body
      author {
        avatarUrl
        login
      }
      labels(first: 5) {
        edges {
          node {
            name
            color
          }
        }
      }
      comments(first: 100) {
        edges {
          node {
            body
            author {
              avatarUrl
              login
            }
          }
        }
      }
    }
  }
}
```

andrewda:
And fetch a repository:

```graphql
{
  repository(owner: "gitpoint", name: "git-point") {
    description
    primaryLanguage {
      name
      color
    }
    owner {
      avatarUrl
      login
    }
    releases {
      totalCount
    }
    stargazers {
      totalCount
    }
    forks {
      totalCount
    }
    # Contributors:
    mentionableUsers(first: 100) {
      edges {
        node {
          avatarUrl
          login
        }
      }
    }
    openIssues: issues(first: 3, states: OPEN) {
      totalCount
      edges {
        node {
          title
        }
      }
    }
    closedIssues: issues(first: 3, states: CLOSED) {
      totalCount
      edges {
        node {
          title
        }
      }
    }
    openPullRequests: pullRequests(first: 3, states: OPEN) {
      totalCount
      edges {
        node {
          title
        }
      }
    }
    closedPullRequests: pullRequests(first: 3, states: [CLOSED, MERGED]) {
      totalCount
      edges {
        node {
          title
        }
      }
    }
  }
}
```

andrewda:
Fetching a user (query score: 1):

```graphql
{
  user(login: "andrewda") {
    name
    bio
    email
    location
    company
    websiteUrl
    viewerIsFollowing
  	followers {
      totalCount
    }
    following {
      totalCount
    }
    repositories {
      totalCount
    }
    organizations(first: 100) {
      edges {
        node {
          avatarUrl
          login
        }
      }
    }
  }
}
```

reyhansofian:
you can also use `viewer` nodes for current logged in user

```
{
  viewer {
    name
    bio
    email
    location
    company
    websiteUrl
    viewerIsFollowing
[...]

PR #455, opened 2017-10-08
wip: use GraphQL API to fetch issues and pull requests

cheshire137:
This is a work in progress to solicit some feedback about the approach. If this seems worthwhile, I'll continue working on this branch!

This branch partially addresses https://github.com/gitpoint/git-point/issues/141 by updating how a repository's issues and pull requests are fetched, so we use v4 of the GitHub API instead of v3. What's missing currently is the views for individual issues and pull requests are not working because they're still expecting data in the v3 format instead of v4. I added a new `getPullRequests` function because PRs are not automatically returned in the list of issues in v4 of the API.

There are a couple style changes I didn't intentionally make. 🤔 I believe when I commit my changes, a git hook is doing some automatic style fixes and it's adjusting lines I didn't otherwise touch. Specifically, it's removing extra parentheses from around components. Is this expected?

This branch also adds me to the contributors list. 🎉


lex111:
Well done, it's time to GraphQL!

@cheshire137 there is another approach to do GraphQL queries, we already [discussed](https://github.com/gitpoint/git-point/pull/347#discussion_r140639476) this in PR #347: you pass currently a ready query (with values variables), but instead you can only [pass the query and variables to it](https://github.com/gitpoint/git-point/pull/347/files#diff-284817f1b6aad47c0ac6d61bd72f474cR350). We considered this way, because GraphQL queries may be large and we wanted to store them in separate files, but by now we have not decided how and where to store them. @machour @housseindjirdeh is not it?

housseindjirdeh:
This looks solid!! :dancer: 

I'm still relatively new to GraphQL and I'm not sure @lex111. It does look like queries can be quite large so do you think we should keep them in the `actions` file like we have here or in the `api` file similar to your PR branch? Let us know what you think @cheshire137. Also, @machour will appreciate your thoughts on this as well.

And with regards to formatting changes, you're right :) We're using [prettier](https://github.com/prettier/prettier) as a pre-commit hook hence why you're noticing a few things different.

machour:
Wow wow, that's looking just amazing @cheshire137 ! 
I'll give your PR a try as soon as possible and see how it plays out.

I've been busy on my side with another PR that I just posted: #457.

It's a big refactoring of our current store, and it introduces a redux middleware and a completely new REST API that is aiming at replacing `v3`. ( 😄  )

Could you give it a read on your side, and see how this new store could be used for graphql?

cheshire137:
@lex111, I like what you're doing with the variable declarations in #347, like `query($login: String!)`.

> It does look like queries can be quite large so do you think we should keep them in the actions file like we have here or in the api file similar to your PR branch?

I agree GraphQL queries can get long, especially if they stay readable by being broken onto multiple lines. I like the idea of keeping multiple queries in a file, but only the ones related to each other. So not one giant file of all the GraphQL queries used in the whole app. I think src/api/queries/repositories.js or src/repository/repository.queries.js makes sense in this branch. I'll pick one style but I'm happy to change.

> I'll give your PR a try as soon as possible and see how it plays out.

Thanks! Beware right now when you click an issue or PR from a repo page, the issue/PR page will be empty of data.

> Could you give it a read on your side, and see how this new store could be used for graphql?

Yes! I'll look.

CONTRIBUTING.md
```diff
@@ -17,7 +17,9 @@ If you want to open a PR that fixes a bug or adds a feature, then we can't thank
 
 ### Working on issues
 
-Please feel free to take on any issue that's currently open. Feel free to resolve any issue that you would enjoy working on even if it happens to be a low priority.
+Please feel free to take on any issue that's currently open. You could look at
+[issues labeled "high priority"](https://github.com/gitpoint/git-point/issues?q=is%3Aopen+is%3Aissue+label%3A%22high+priority%22),
+but feel free to resolve any issue that you would enjoy working on even if it happens to be a low priority.
```
machour: not sure this should be a part of this PR

src/api/index.js
```diff
@@ -21,6 +21,36 @@ const METHOD = {
...
 
+export const v4 = {
+  root: 'https://api.github.com/graphql',
+  call: async parameters => {
+    const response = await fetch(v4.root, parameters);
+    const json = JSON.parse(response._bodyText);
+
+    if (response.status >= 200 && response.status < 300) {
+      return json.data;
+    }
```
machour: Could we test for failed response.statuses instead before doing the JSON.parse() ?

src/api/index.js
```diff
@@ -21,6 +21,36 @@ const METHOD = {
...
+    const error = new Error(json.message);
+
+    error.response = response;
+    throw error;
+  },
```
machour: What about `return Promise.reject(error.response)` instead?

src/api/index.js
```diff
@@ -21,6 +21,36 @@ const METHOD = {
...
+  parameters: (accessToken, body = {}) => {
+    return {
+      method: METHOD.POST,
+      headers: {
+        Authorization: `token ${accessToken}`,
+      },
```
machour: Haven't tried GraphQL on my side yet. Are this header enough? Shouldn't we be passing a User-Agent ? (GitPoint/Android-3.2 for example).

I'd like to have it shared between GraphQL & Rest

src/components/issue-list-item.component.js
```diff
@@ -55,15 +55,15 @@ const styles = StyleSheet.create({
 
 const getIconName = (type, issue) => {
   if (type === 'issue') {
-    return issue.state === 'closed' ? 'issue-closed' : 'issue-opened';
+    return issue.state === 'CLOSED' ? 'issue-closed' : 'issue-opened';
   }
```
machour: `CLOSED`, `OPEN`, .. should be moved in a string enum

src/api/index.js
```diff
@@ -21,6 +21,36 @@ const METHOD = {
...
+      body: JSON.stringify(body),
+    };
+  },
+  post: async (accessToken, body) => {
+    const response = await v4.call(v4.parameters(accessToken, body));
+
+    return response;
+  },
```
machour: is `post()` the only method that will exist to access GraphQL ?
If that's the case, maybe we can merge `call()` & `parameters()` inside it

src/issue/screens/issue.screen.js
```diff
@@ -176,20 +176,25 @@ class Issue extends Component {
...
+    let issueRepository;
+
+    if (params.issueURL) {
+      issueRepository = issueURL
+        .replace(`${v3.root}/repos/`, '')
+        .replace(/([^/]+\/[^/]+)\/issues\/\d+$/, '$1');
+    } else if (params.issue) {
+      issueRepository = params.issue.repository.nameWithOwner;
+    }
+    const repositoryUrl = `https://api.github.com/repos/${issueRepository}`;
```
machour: Should use `v3.root`

src/repository/screens/repository.screen.js
```diff
@@ -207,14 +210,9 @@ class Repository extends Component {
...
-    const pureIssues = issues.filter(issue => {
-      // eslint-disable-next-line no-prototype-builtins
-      return !issue.hasOwnProperty('pull_request');
-    });
 
-    const openPulls = pulls.filter(pull => pull.state === 'open');
-    const openIssues = pureIssues.filter(issue => issue.state === 'open');
+    const openPulls = pullRequests.filter(pull => pull.state === 'OPEN');
+    const openIssues = issues.filter(issue => issue.state === 'OPEN');
     const showFork =
```
machour: Using reactotron, I see that when we navigate to a repository screen, we pull 100 issues & 100 pull requests from the repo.
If the user doesn't click on "View All" for issues or PR, this is a waste of data & Github credit.

Could we have a single GraphQL query that retrieves only the 3 most recent & open issues & PR instead?

src/api/index.js
cheshire137: The GraphQL API returns JSON on error as well as success. I do the JSON parsing ahead of this line so I can get the message out of the body as to why the query failed, to put in the `Error`.

src/api/index.js
cheshire137: This is all that's necessary to use the endpoint, though I could add a user-agent header. See the example at https://developer.github.com/v4/guides/forming-calls/#communicating-with-graphql.

For trying the GraphQL API, I like https://developer.github.com/v4/explorer/, which lets you run queries as your GitHub user in the browser.

src/api/index.js
cheshire137: Yep, all requests whether they're reading or writing, are POSTs; see https://developer.github.com/v4/guides/forming-calls/#communicating-with-graphql.

I can merge these methods!

src/repository/screens/repository.screen.js
cheshire137: Yes! I wondered about this, but was trying to translate directly from how the v3 API was being used to v4. I'll change the code that's slicing the first three results in JavaScript so that instead we just pass `first: 3` to the API.

cheshire137:
> the issues/pulls screens seems to be empty when I visit them. Is that intended for now or did I hit a bug?

That's intended! It's where I left off.

src/api/index.js
machour: Didn't know that, thanks for clarifying it, it's perfect that way then.

cheshire137: I might be able to do `await response.json()` though instead of parsing the body myself, though. 🤔

machour:
Thanks for the feedback @cheshire137.

I remembered that @andrewda did an extensive study on GraphQL ressources limitation, and the queries scores in #141, and I wonder in what place should we use GraphQL, and where we should instead rely on REST to stay under the limits.

The way I see it, GraphQL is **great** for screens built with various data from different endpoint. The repository screen is a good example, here are the REST requests currently made:
 1. get the repository information
 2. get the repository forks count
 3. get the contributors
 4. check the existence of the README.md file
 5. get 100 issues
 6. get 100 pull requests

Do you think these six REST calls could be reduced to a single GraphQL query, that would return:
  * repo basic informations (the one displayed on that screen)
  * the 30 first contributors, with their login & avatarUrl
  * a boolean that tells if the README file exists
  * the 3 most recent issues with status OPEN
  * the 3 most recent pulls with status OPEN
  * the total number of open issues
  * the total number of open pulls
  * the total number number of forks

On the other hand, I'm not sure GraphQL would be preferable to REST if we just want to paginate the issues list for example, as we're able to get 100 issues in a single REST call, without consuming any credit available for GraphQL calls.

I'm curious about your thoughts on that

src/api/index.js
machour: And it makes more sense along with Promise.reject() 👌

cheshire137:
> the 30 first contributors, with their login & avatarUrl

To my knowledge, there's not a connection in GraphQL that returns a repository's contributors. How were you getting at this list via the REST API?

The other data in your list--repo basic info, if the README exists, etc.--all sound fine to fetch via GraphQL.

> I'm not sure GraphQL would be preferable to REST if we just want to paginate the issues list for example, as we're able to get 100 issues in a single REST call, without consuming any credit

This is something to consider. The biggest reason to use GraphQL to me is to get at the data we want more directly, and without getting a bunch of other data we don't need.

cheshire137:
> ...without consuming any credit available for GraphQL calls.

Can you share a link with info about GraphQL credits?

machour:
Contributors support seems to have been just added a few days ago: https://platform.github.community/t/cant-access-the-contributors-of-a-repository/1848/8?u=machour

With rest, we are using this endpoint https://api.github.com/repos/gitpoint/git-point/contributors

Agreed with you, being able to hand pick the returned fields is gold with GraphQL. Weird REST doesn't provide the same thing via a `?fields` GET parameters (gitlab & bitbucket do it afair)

Here's for GraphQL scoring system at GitHub: https://developer.github.com/v4/guides/resource-limitations/ (haven't done any math yet, so I'm maybe worrying for nothing)

cheshire137:
Looks like 100 issues in GraphQL should just cost 1, and that that rate limit should be shared with the REST API. So using 1 credit for GraphQL means there's 1 fewer for REST. It starts to jump if we nest connections (like grab 100 repos and for each of those, grab 100 issues).

We can also stick `rateLimit { cost }` in a GraphQL query to see how much the query will cost. We can do `rateLimit(dryRun: true) { cost }` to see the cost of a query without actually having it count or getting our requested data back.

cheshire137:
> Contributors support seems to have been just added a few days ago

Oh nice! Then your whole list can be fetched via GraphQL if we want.

CONTRIBUTING.md
housseindjirdeh: Fair enough, personally don't mind having this correction included however.

.all-contributorsrc
```diff
@@ -398,6 +398,15 @@
       "contributions": [
         "doc"
       ]
+    },
+    {
+      "login": "cheshire137",
```

G2Jose:
Have you guys also looked into [Relay](https://github.com/facebook/relay) and [Apollo](https://github.com/apollographql/apollo-client) ?

cheshire137:
Time got away with me on this one! If someone else wanted to take this over from me, I'd be happy to let them.

machour:
Closing this PR as it's superseded by the ongoing store refactoring. Thank you so much for all the efforts and explanations @cheshire137, the next implementation is inspired by your work!

Issue #430, opened 2017-10-04
Show API errors in toast


lex111:
At the moment, if we have an error when requesting something from the API, for example when searching, this is not handled in any way and the error is not shown. There is a suggestion to show an error in [toast](https://github.com/magicismight/react-native-root-toast). Or does anyone have any other thoughts on this matter?

nersoh:
I think that would be good. The more feedback users have, the better.

Harkindey:
Can i work on this?

housseindjirdeh:
Think this would be great to have. We can begin with thinking implementing `error` messages using toasts and then moving on to other types of notifications if we see fit (show a toast if request successful for example?)

Would be great if we can use a higher order component or something similar to simply have this toast _in every screen_ if an error is triggered :thinking: Don't know if this would be easy however given our error messages are in each part of state specifically for different requests.

Happy to hear any thoughts on this 💬

PR #457, opened 2017-10-09
refactor(redux): Switch to redux middleware & normalizr for Rest

machour:
Hi,

This is a first draft for a big refactoring of our current store implementation.
It relies on a redux middleware & normalizr to organise the store.

I started implementing it in the Organizations section, and I think it looks solid
enough to showcase this refactoring and start a discussion.

Start Reactotron & subscribe to the following reducers :
 * entities
 * pagination
 * errorMessage

Initial state:
screen shot 2017-10-09 at 2 47 20 am

Follow the described navigation, and start witnessing your subscriptions changes:

## The organization profile screen:

Go to the "GitPoint" organization profile screen.

Two api calls were made: `providers/github/repos.js@getById()` && `providers/github/repos.js@getMembers()`.

The first API call checked for the org in `entities`, did not find it, so dispatched an action interpreted by the result middleware. 
After the `fetch()` call in `performApiRequest()`, the returned JSON was normalized using the `orgSchema`, and stored by the `entities` reducers like so : 
screen shot 2017-10-09 at 2 48 44 am

The second API call, checked for the existence of "gitpoint" in `pagination.membersByOrg`, did not find it, so dispatched an action. The returned response from Github contained two things:
* A JSON array of users, which was reduced by `entities` using the `userSchema`:
screen shot 2017-10-09 at 2 57 36 am

* Pagination information in the `Link` header, which was parsed and stored by the `pagination` reducer in `pagination.membersByOrg.gitpoint`.
Because we specified that we expected an array of `userSchema`, normalizr was able to store only the usernames of the members in the pagination, instead of the full user information. (this is reason number 1 that normalizr() is awesome)
screen shot 2017-10-09 at 3 01 04 am

(I specifically asked for 8 users per page of result to demonstrate the pagination)

By passing the two functions (`getById` and `getMembers`)  as `mapDispatchToProps` in the OrganizationProfileScreen, we can define the following `mapStateToProps` function:

```js
const mapStateToProps = (state, ownProps) => {
  // We only need the organization id, so navigate() calls should be changed to get rid
  // of params.organization. Kept for now to avoid messing up things
  const orgId = ownProps.navigation.state.params.organization.login.toLowerCase();

  // We retrieve the things we're interested in following in state
  const { pagination: { membersByOrg }, entities: { orgs, users } } = state;

  // We initialize the members pagination as Fetching as it will start fetching right away
  // in componentWillMount()
  const membersPagination = membersByOrg[orgId] || {
    ids: [],
    isFetching: true,
  };
  // Here's how we transform our array of usernames to actual users
  const members = membersPagination.ids.map(id => users[id]);

  return {
    orgId,
    members,
    membersPagination,
    org: orgs[orgId],
  };
};
```

## Scrolling the members list

Remember how I said I limited the pagination to 8 members per page? Let's scroll to the right in the avatars list. By listening on `onEndReached`, we're able to re-trigger a call to `getMembers()` with an additional parameter this time: `getMembers(orgId, { loadMore: true })`

This second parameter makes the next call to Github use the URL stored in `pagination.memberByOrg.gitpoint.nextPageUrl` that appeared in last screenshot, and that asks for the next page of results. New results will be reduced using the same process as earlier, and our usernames array will grow with the new references to users.

screen shot 2017-10-09 at 3 16 27 am

As this was the last page of results, `nextPageUrl` got set to null, and no more apis calls will be made.

## In and out of the organization screen

Go back in navigation, then forward to the organization screen again.
Both `getById()` and `getMembers()` were called again, but as everything was already in the store, the calls to Github api were blocked. (see `getById()` code)

## Refreshing the organization screen

Unfold `entities.orgs.gitpoint` and observe the `_fetchedAt` property. It's a timestamp of the moment we normalized the entity in the schema.

Pull down the screen to refresh. `getById(orgId)` & `getMembers(orgId)` will be called again, this time with a `forceRefresh` option, that will perform an API call to Github, regardless of what we have or not in our store.

Once the response is fetched, the usual reducing/normalization takes places, and the entity is swapped in the store. See how `_fetchedAt` changed? Pull the screen again. Such amaze. Much wow.

(:warning: I still haven't fully implemented forceRefreshing on paginated list like getMembers(), so its pagination in the store will look off.)

## Browsing the organization repositories

Click the number of repos on the organization screen.
This time, we'll be fetching `entities.repos` entities, and paginating using the `reposByOrg` entry in the store.

If you use the search, we we'll be using `reposBySearch` instead, and indexing with the complete query string sent to GitHub. Here's an example searching for `site`

screen shot 2017-10-09 at 3 24 10 am

##  Moar data

Well, ok, 2 repositories & 12 members are not enough to really enjoy this baby.
Let's navigate to the "search" tab, and search for "ionic", select the first repo in results, and then click on the owner.

You should now be on the Organization page for the "Ionic" team.

Wait. Navigate back to the tab you came from.
Notice how data is still related to gitpoint, instead of being swapped to Ionic?

You can also try and go to 4 different organizations on the 4 tabs, scroll down their repositories, and switch beetween tabs.

One of our major flaws just got fixed ! 🕺 

### Schemas are wonderful

1. By extensively using the `processStrategy` of normalizr schemas, we are able to craft our stored entities with care to only keep attributes that are used by the app, and minimize the store size (we don't store the dozen of urls that comes with every response from github for example)

2. We can standardize some useful internal attributes:

You already know about `_fetchedAt` (that we will be able to use later to leverage [conditional request](https://developer.github.com/v3/#conditional-requests).

There is also `_entityUrl` which points to the entity on github website. This will be usefull for #285 

There is also `_isComplete` which tells us if the entity was completely or partially received. (For example, the members returned by getMember() are incomplete users, in comparison to what you get on the `/user/:username`). (Usage still WIP)

### Isn't that too much for the store ?

If we're worried about the store growing too much in size, we can add a `_lastUsed` property.
In `getById()` for example, when we check the store and find that `gitpoint` is already in the store, we could dispatch a `UPDATE_USED_AT` action to update the `_lastUsed` attribute.

Then, we could implement an other action, triggered on low memory warning, (or when we have more than N entities) to go through entities and remove the old entries.

### Not persisted

`entities`, `pagination` and `errorMessage` are blacklisted from the persistent store. Fresh data every time we open the app. 

### Gitlab? Bitbucket?

With this new redux middleware, and the use of schemas, we could add support for Gitlab & Bitbucket by keeping things standardized.

By implementing provider-specific schemas, and adopting an internal convention for those, we will be able to produce entities with the exact same structure for all 3 providers.

Which means that all screens/components using these entities, can display the data regardless of where they came from.

Of course, for now, our first objective is to bring the Github provider up and 100% making use of this new redux store.

I intendedly added an extra `providers/github` path in this PR, for us to keep standardization in mind.

Voilà!


src/api/rest/providers/github/endpoints/search.js
```diff
@@ -0,0 +1,23 @@
...
+import Schemas from '../schemas';
+
+const _searchRepos = (query, nextPageUrl) => ({
+  id: query,
+  [CALL_API]: {
+    types: REPOS_BY_SEARCH,
+    endpoint: nextPageUrl,
+    schema: Schemas.REPO_ARRAY,
+    normalizrKey: 'items',
+  },
```
machour: Added `normalizrKey` here because the repositories array is in the `item` keys of the JSON response.
We may need to come up with something cleaner if there are more than a level to cross

alejandronanez:
Just wow, will take a look this week! Thanks for the huge effort!

src/api/rest/providers/github/endpoints/orgs.js
```diff
@@ -0,0 +1,65 @@
...
+  },
+});
+
+export const getMembers = (orgId, options) => {
+  return handlePaginatedApi(
+    `orgs/${orgId}/members?per_page=8`,
+    { name: 'membersByOrg', key: orgId, call: _getMembers },
+    options
+  );
+};
```
machour: Without the `handlePaginatedApi` shorthand, `getMembers` would look like this:

```js
export const getMembers = (orgId, { loadMore = false, forceRefresh = false } = {}) => (dispatch, getState) => {
  const firstPageUrl = `orgs/${orgId}/members?per_page=8`;

  let { nextPageUrl = firstPageUrl } = getState().pagination.membersByOrg[orgId] || {};
  const { pageCount = 0, isFetching = false } =
    getState().pagination.membersByOrg[orgId] || {};

  if (forceRefresh) {
    // TODO: how to reset the state ? dispatch(clearPagination('paginationId')) ?
    nextPageUrl = firstPageUrl;
  } else if (isFetching || (pageCount > 0 && !loadMore) || !nextPageUrl) {
    return null;
  }

  return dispatch(_getMembers(orgId, nextPageUrl));
};
```

src/api/rest/providers/github/schemas/orgs.js
```diff
@@ -0,0 +1,152 @@
+import { schema } from 'normalizr';
```
machour: This schema is the one extensively & accurately (no copy/paste) documented.

Private flags setting should be mututalized between all schemas with a helper function: 
const processed = initSchema(entity.id); // { _isComplete: false, _isAuth: false, _entityUrl: false, _fetchedAt: `now`, id: `the passed id`

src/api/rest/reducers/entities.js
```diff
@@ -0,0 +1,17 @@
+import merge from 'lodash.merge';
+
+// Updates an entity cache in response to any action with response.entities.
+export const entities = (
+  state = {
+    users: {},
+    orgs: {},
+    repos: {},
+  },
```
machour: + `commits` indexed by `repo`+`hash id`
+ `issues` indexed by `repo`+`issue id`
+ ...

src/api/rest/reducers/errorMessage.js
```diff
@@ -0,0 +1,14 @@
+const RESET_ERROR_MESSAGE = 'RESET_ERROR';
```
machour: @lex111 in here we could store errors from the API and toast them as you mentioned in #430

cheshire137:
Just a wee little PR. 😅 

refactor_redux___switch_to_redux_middleware___normalizr_for_rest_by_machour_ _pull_request__457_ _gitpoint_git-point

src/components/users-avatar-list.component.js
```diff
@@ -0,0 +1,127 @@
...
+  noMembersMessage: string,
+  containerStyle: Object,
+  smallTitle: string,
+  navigation: Object,
+  authUser: Object,
+  loadMore: Function,
+};
+
+const size = 120;
+
```
machour: Made avatars bigger on purpose to test pagination

src/organization/screens/organization-profile.screen.js
```diff
@@ -1,51 +1,20 @@
 import React, { Component } from 'react';
```
machour: The amount of red in this diff <3
Screens gets simplified

src/organization/screens/organization-profile.screen.js
```diff
@@ -84,23 +56,29 @@ class OrganizationProfile extends Component {
...
-  getOrgData = () => {
-    const organization = this.props.navigation.state.params.organization;
+  refreshData = () => {
+    const { navigation, getMembers, getById } = this.props;
+    const orgId = navigation.state.params.organization.login;
+
+    navigation.setParams({ refreshing: true });
+    getById(orgId, { forceRefresh: true });
+    getMembers(orgId, { forceRefresh: true });
+  };
```
machour: This can be a mutualized behavior in our screens, where we just call loadData() with force refresh

src/organization/screens/organization-profile.screen.js
```diff
@@ -109,83 +87,82 @@ class OrganizationProfile extends Component {
...
     const initialOrganization = this.props.navigation.state.params.organization;
     const organizationActions = [translate('common.openInBrowser', language)];
 
+    if (!org) {
+      return (
+        <Text>
+          Loading organization {orgId} .. TODO: Make me look nicer
+        </Text>
+      );
+    }
```
machour: We should instead be drawing the screen with loaders everywhere where we don't have the info in store yet

src/organization/screens/organization-profile.screen.js
```diff
@@ -109,83 +87,82 @@ class OrganizationProfile extends Component {
...
-            <UserProfile
-              type="org"
-              initialUser={initialOrganization}
-              user={
-                initialOrganization.login === organization.login
-                  ? organization
+          renderContent={() =>
+            <OrgProfile
+              initialOrg={initialOrganization}
+              org={
```
machour: what is this initalOrganization / initialUser being passed everywhere? 💭

src/organization/screens/organization-profile.screen.js
```diff
@@ -109,83 +87,82 @@ class OrganizationProfile extends Component {
...
+            !membersPagination.pageCount &&
             <LoadingMembersList
               title={translate('organization.main.membersTitle', language)}
-            />
-          )}
+            />}
 
-          {!isPendingMembers && (
-            <MembersList
+          {(!membersPagination.isFetching || membersPagination.pageCount > 0) &&
```
machour: This pagination test should be extracted to a helper function, that tells if a paginator already have done an initial load

src/organization/screens/organization-profile.screen.js
```diff
@@ -205,6 +182,32 @@ class OrganizationProfile extends Component {
   }
 }
 
-export const OrganizationProfileScreen = connect(selectors, actions)(
-  OrganizationProfile
-);
+const mapStateToProps = (state, ownProps) => {
+  // TODO: This should be normalized to params.id
+  const orgId = ownProps.navigation.state.params.organization.login.toLowerCase();
+
```
machour: Once implemented everywhere, screens should take just an "id" instead of an object like "user" or "organization".

The id suffice to get the data from the redux store, and we will standardize & simplify navigation greatly

src/organization/screens/repository-list.screen.js
```diff
@@ -0,0 +1,223 @@
+/* eslint-disable no-shadow */
```
machour: refactor: This screen should be extracted to `common/screens/` and be used in **all** repositories list screen.

src/organization/screens/repository-list.screen.js
```diff
@@ -0,0 +1,223 @@
...
+
+      this.setState({
+        searchStart: true,
+        query,
+      });
+
+      const { searchRepos, navigation } = this.props;
+
+      navigation.setParams({ searchedKeyword });
+      searchRepos(searchedKeyword);
```
machour: I used `navigation.setParams` to store the searchedKeyword, later used in `mapStateToProps()`.
Is that okay, or considered a bad practice?

lex111:
@machour I will try to see this PR in detail as soon as possible, but with a cursory examination it is clear that a very cool work! :clap:

housseindjirdeh:
Thanks so much for this @machour, this is going to take some time to digest so I'll have some time blocked hopefully in a few days to go through this in detail. :tada:

machour:
Since last comment, I did the following:

Clean up:
- removed old & unused actions, reducer, types from organization
- moved `src/organization/screen/*` to `src/screens/organization/` (proposed re-organisation)

Refactor:
- kept only the redux logic in the organization repos screen (container)
- created a `<RepositoryList />` components that only deals with the UI (dumb).

UI ❤️ :
- added a loading indicator at the bottom of the list which disappears when last page is reached

src/components/users-avatar-list.component.js
```diff
@@ -0,0 +1,127 @@
...
+    marginBottom: 10,
+    paddingLeft: 15,
+  },
+  flatList: {
+    paddingLeft: 15,
+    paddingRight: 15,
+  },
+});
+
+const UsersAvatarListComponent = ({
```
lex111: Partially the last avatar is not displayed:

image

src/screens/organization/organization-profile.js
```diff
@@ -0,0 +1,198 @@
...
+  OrgProfile,
+  SectionList,
+  ParallaxScroll,
+  LoadingMembersList,
+  EntityInfo,
+  UsersAvatarList,
+} from 'components';
+import { emojifyText, translate, openURLInView } from 'utils';
+import { colors, fonts } from 'config';
+import { getById, getMembers } from 'api/rest/providers/github/endpoints/orgs';
```
lex111: Just for further consideration: it would be cool to just write :thinking: 
```js
import { getById, getMembers } from 'api';
```

More precisely, then it would be so:
```js
import { getOrganizationById, getOrganizationMembers } from 'api';
```

Proposal: instead of `getOrganizationById` maybe was simply `getOrganization`

routes.js
```diff
@@ -61,6 +60,12 @@ import {
 } from 'issue';
 
 const sharedRoutes = {
+  OrgRepositoryList: {
```
lex111: Perhaps it's better to use the full name here (`OrganizationRepositoryList`), because the other screens start with the `Organization`, not the `Org` (to support consistency)

src/screens/organization/repository-list.js
```diff
@@ -0,0 +1,97 @@
+/* eslint-disable no-shadow */
+import React, { Component } from 'react';
+import { connect } from 'react-redux';
+
+import { RepositoryList } from 'components';
+
+import { getRepos } from 'api/rest/providers/github/endpoints/orgs';
+import { searchRepos } from 'api/rest/providers/github/endpoints/search';
```
lex111: How about doing something like this?

```js
import { getOrganizationRepos, searchRepos } from 'api/rest/providers/github/endpoints';
```

src/components/repo-list-item.component.js
```diff
@@ -0,0 +1,132 @@
...
+          color={languageColors[repository.language]}
+        />}
+
+      <Text style={styles.extraInfoSubject}>
+        {repository.language}
+      </Text>
+    </View>
+  </View>;
+
+export const RepoListItem = ({ repository, showFullName, navigation }: Props) =>
```
lex111: Maybe use full name `RepositoryListItem`, because there is already a `RepositoryList`?

root.reducer.js
```diff
@@ -2,16 +2,21 @@ import { combineReducers } from 'redux';
 import { authReducer } from 'auth';
 import { userReducer } from 'user';
 import { repositoryReducer } from 'repository';
-import { organizationReducer } from 'organization';
 import { issueReducer } from 'issue';
 import { searchReducer } from 'search';
 import { notificationsReducer } from 'notifications';
 
+import { entities } from './src/api/rest/reducers/entities';
```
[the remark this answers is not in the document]
machour: it's already there, but somehow it doesn't work
used src/ to move forward with coding instead of banging my head on the desk :D

root.store.js
```diff
@@ -6,9 +6,10 @@ import reduxThunk from 'redux-thunk';
 import { composeWithDevTools } from 'redux-devtools-extension';
 import 'config/reactotron';
 import { rootReducer } from './root.reducer';
+import restApi from './src/api/rest/middleware';
```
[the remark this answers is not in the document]
machour: There will be a similar middleware in `src/api/graphql/middleware/` I think
Still have to test a GraphQL implementation and see if maybe we can re-use it

routes.js
```diff
@@ -12,6 +12,8 @@ import { NotificationIcon } from 'components';
 import { colors } from 'config';
 import { translate } from 'utils';
 
+import screens from 'screens';
```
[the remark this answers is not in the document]
machour: yes, because `actions`, `reducers` and `types` are now gone from `organization/` (make sure you pulled my latest changes)

routes.js
machour: Absolutely

src/api/rest/actions/orgs.js
```diff
@@ -0,0 +1,7 @@
+import { createActionSet } from 'utils';
+
+export const ORGS_GET_BY_ID = createActionSet('ORGS_GET_BY_ID');
+export const ORGS_GET_REPOS = createActionSet('ORGS_GET_REPOS');
```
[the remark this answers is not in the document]
machour: When I started this PR, I was trying to have the following namespace set up:

```
client.orgs.* //everything related to orgs operation, like
client.orgs.getById(orgId)
client.orgs.getMembers(orgId)

client.users.get() // gets the authenticated user
```

The goal is to namespace the API by entities. It seems to me more clear and well dispatched.

But `connect()` only supports plain functions, so I had to import { getById } from './orgs/';

I was thinking about renaming `orgs.getById()` to `orgsGetById`, `users.get` to `usersGet`, etc, so that it's more clear what is the real function we're calling

src/api/rest/middleware/index.js
```diff
@@ -0,0 +1,62 @@
...
+  }
+
+  if (typeof types !== 'object') {
+    throw new Error('Expected an object containing the three action types.');
+  }
+
+  const accessToken = store.getState().auth.accessToken;
+
+  const actionWith = data => {
+    const finalAction = Object.assign({}, action, data);
```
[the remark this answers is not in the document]
machour: Definitely. FYI, I stole the base logic from this sample: (https://github.com/reactjs/redux/tree/master/examples/real-world) and kept some stuff untouched

src/api/rest/providers/github/endpoints/orgs.js
```diff
@@ -0,0 +1,65 @@
...
+  orgId,
+  { requiredFields = [], forceRefresh = false } = {}
+) => (dispatch, getState) => {
+  const org = getState().entities.orgs[orgId];
+
+  if (!forceRefresh && org && requiredFields.every(key => has(org, key))) {
+    return null;
+  }
+
+  return dispatch(_getById(orgId));
```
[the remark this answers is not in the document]
machour: I avoided using an anonymous function for now, so that the code is understandable.

The proposed renaming is: 
 - exported function: orgsGetById()
 - private function getById() (no need to prefix with orgs, we're inside orgs.js)

src/api/rest/reducers/pagination.js
```diff
@@ -0,0 +1,75 @@
...
+// and a function telling how to extract the key from an action.
+const paginate = types => {
+  if (typeof types !== 'object') {
+    throw new Error('Expected types to be an object of three props.');
+  }
+
+  const updatePagination = (
+    state = {
+      isFetching: false,
+      nextPageUrl: undefined,
```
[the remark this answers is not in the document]
machour: nextPageUrl === null means there is no more pages left, pagination is ended

src/components/org-profile.component.js
```diff
@@ -0,0 +1,146 @@
...
+            org.avatarUrl &&
+            <ImageZoom
+              uri={{
+                uri: org.avatarUrl,
+              }}
+              style={[styles.avatar]}
+            />}
+          {(!org || !org.avatarUrl) &&
+            <Image
+              source={require('../assets/images/loading.gif')}
```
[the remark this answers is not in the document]
machour: Didn't know about it, and I was thinking of having an animated version of gitpoint logo for image loading..

src/components/users-avatar-list.component.js
machour: Yes, that's the case in `master` too

src/screens/organization/repository-list.js
machour: I would like to have an `orgs` prefix for organization related methods (like a namespace, but without the '.')

I inspired my naming from https://octokit.github.io/node-github/ which seems to me like it's nailing it.

src/components/repo-list-item.component.js
machour: Yup yup.

Again, naming/imports paths were not really polished, and we need to focus on more critical architectural stuff for the store ;)

Did you get a chance to watch the Redux store being populated? What do you think about the schemas thing?

src/screens/organization/repository-list.js
lex111: This comment should be considered with [this](https://github.com/gitpoint/git-point/pull/457#discussion_r144004561).

If we assume that there will be other providers, including.

src/api/rest/providers/github/endpoints/orgs.js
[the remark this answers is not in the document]
lex111: Agree, but why `orgsGetById` instead of `getOrgsById`?

src/components/repo-list-item.component.js
lex111: Yes, I looked through the debugger, it's simple and understandable. Results are saved (cached), very good with our navigation. And most importantly, you can refuse now from `reducers` and `actions`, delete them! I could not imagine, but it turned out to be not so difficult, all ingenious is simple :smile: 

Of course, we still need to test, but in general it's very good.

Regarding the scheme, I thought it would be a general (basic) scheme for each entity, because we will have a own scheme for REST (v3) and GraphQL (v4)? I wanted them to follow the basic scheme, where by the way the types of fields would be spelled out eg.

src/api/rest/providers/github/endpoints/orgs.js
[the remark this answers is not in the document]
machour: Just pushed a refactoring that namespaces the client like I originally intended.

Now from a screen, you simply import `client` from `providers/github`, then you pass the functions you need as props in `connect()`, renaming them as you want for the screen code.

See here: https://github.com/machour/git-point/blob/82495a62accfc4b7f383bcb3e5862592a3efd7ff/src/screens/organization/repository-list.js#L94

Does that sound better to you?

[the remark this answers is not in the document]
lex111: Now, of course! At once it became much clearer, you too?

src/components/repo-list-item.component.js
machour: Completely. The basic scheme would also apply to future Bitbucket & Gitlab schema.
I wanted to use flow to add type checking, but I'm unable to get it to work correctly

src/api/rest/providers/github/endpoints/orgs.js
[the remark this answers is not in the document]
lex111: And still, I do not really like that the spelling of entities is differently used: `orgs` or `organizations`, `repos` or `repositories`. I want to just come to the same spelling, and I would choose the full name, and it used it everywhere.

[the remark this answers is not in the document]
machour: Yes! I was struggling for hours trying to use a Proxy() on the parameters passed to `connect()`, before thinking of doing this trivial thing 😭 

Just like you mentioned for schema, I'd like to have a base client, used only to document endpoints & methods (and to typecheck future implementations later)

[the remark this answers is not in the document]
machour: @lex111 on my side, I'm more tempted to use the short forms everywhere in the app :D 

We'll have a vote on that if needed. #teamShort

src/components/repo-list-item.component.js
lex111: To implement Flow I think it may be already in a separate PR?

But as I understand, in `normalizr` there is no way to do the basic schema?

src/screens/activity/events.screen.js
```diff
@@ -79,25 +80,23 @@ const styles = StyleSheet.create({
 
 class Events extends Component {
   componentDidMount() {
-    const { user: { login }, getUser } = this.props;
+    const { getEvents, user: { login } } = this.props;
 
-    if (login) {
-      this.getUserEvents();
-    } else {
-      getUser();
```
lex111: Maybe I missed something, there are too many changes, but did we get the `user` in advance? The point is that this check is needed when we just enter the application (when only logged in), we do not have the `user` (because of lazy loading tabs), so we get data on it, and only then we receive events for it.

machour: Absolutely buddy, I have a TODO comment for that line 95.

I'm thinking about moving the getUser() call to the "Preparing GitPoint.." screen right after the login. There is no point in opening the App if that special call fail, and we don't have `getState().auth.user`.

 What do you think about that?

lex111: Yeah, totally agree :+1:

src/components/repo-list-item.component.js
machour: We normally already have flow configured in the repo (.flowconfig).

I'm not sure I understand your question about the basic schema for normalizr.

Currently, I implement a Schema by entity. In that schema, I handle  the JSON => entity attributes mapping. 
As the JSON structure from GraphQL will probably be different, we will have to implement new Schema, that will transform GraphQL JSON to the same entity structure we have for REST.

lex111: > As the JSON structure from GraphQL will probably be different, we will have to implement new Schema, that will transform GraphQL JSON to the same entity structure we have for REST.

I meant it, the schema fields will be the same, but because REST and GraphQL return them in different ways (`avatar_url` and `avatarUrl`), I wanted to have a basic schema where all these fields were specified, and they were all in these two schemes (REST and GraphQL).

machour: I don't think it's possible with Normalizr, that's why I was counting on some sort of type checker instead

housseindjirdeh:
I can't thank you enough @lex111 for spending the time to review this.

@machour, apologies are in order. I said I would review it today but unfortunately had so little time outside of work. Will definitely go through this tomorrow and pick your brain as I run and test this throughout.

machour:
Changed the following things in last commit:

### Switched back to class style for the client

- got rid of `github/endpoints/*` in favor of a single `Github` class.
- implemented `Rest` a base class that takes care of `fetch()`
- implemented `Github` that extends Rest and use Rest.fetch() for the API calls

### Got rid of the Redux middleware, in favor of a `withReducers()` "decorator" 
*(not sure of both naming)*

**Motivation**: there are time when we need to call the API, but without anything getting reduced in the store. (post a comment, get a count, ..).

With this refactoring `Github` is now a class extending `Rest`, that can be used like this to perform a direct API call:
```js
const client = new Github();

client.setAccessToken('########');
client.orgs.getRepos('gitpoint').then(struct => {
  // struct contains the following keys:
  // - response: The Response() object obtained from the fetch call
  // - schema: A schema that may be used to normalize() the response (if the developer needs it)
});
```

On the contrary, if we want the call to dispatch actions, and the response to be reduced automagically, we instantiate and use the client like this: 

```js
const client = withReducers(Github);

client.orgs.getRepos('gitpoint');
```

This call will automatically:
 - compute the actions set to be used. The convention is: namespace_function_in_snake_case uppercased (client.users.getFollowers => USERS_GET_FOLLOWERS)
 - check if we already have the requested data in store, if so, block the api call
 - if not perform the API call, and automatically paginate it if needed, and store the data

This *greatly* simplifies the initial version and reduce the amount of code needed to implement new features in the app.


I'm also thinking about a `withAuth()` decorator that would fetch the accessToken from store and do the `setAccessToken()` call automatically for us (for the first example of use).

### Reverted directories changes for screens

Let's keep this PR focused on redux so that changes are more clear. (thanks @jouderianjr)

If this PR is accepted, a new folder structure will be proposed in a second PR, with screens laid out as follow:
```
.
├── api
├── assets
├── components
├── config
├── locale
├── screens
│   ├── activity
│   ├── auth
│   ├── issues
│   ├── orgs
│   ├── repos
│   ├── search
│   └── user
└── utils
```

App.js
```diff
@@ -62,7 +62,7 @@ class App extends Component {
       {
         storage: AsyncStorage,
         transforms: [encryptor],
-        blacklist: ['user'],
+        blacklist: ['entities', 'pagination', 'errorMessage', 'user'],
```
housseindjirdeh: With this and after some thought, I'm really not too concerned with adding more and more normalized entities to our store. If the data isn't persisting, I really don't think this would be too much of an issue 👍

src/api/rest/decorators/with-reducers.js
```diff
@@ -0,0 +1,129 @@
...
+      return new Proxy(client[namespace], {
+        get: (endpoint, call) => (...args) => (dispatch, getState) => {
+          // Used as a key for state.pagination
+          const actionName = `${namespace}_${call}`
+            .replace(/([A-Z])/g, '_$1')
+            .toUpperCase();
+
+          const action = Actions[actionName];
+
+          if (typeof action === 'undefined') {
```
housseindjirdeh: Will `if(action) {...}` be sufficient?

src/api/rest/decorators/with-reducers.js
```diff
@@ -0,0 +1,129 @@
...
+            const { loadMore = false } = magicArg;
+            const { pageCount = 0, isFetching = false, nextPageUrl } =
+              paginator[pureArgs.join('-')] || {};
+
+            if (
+              isFetching ||
+              (pageCount > 0 && !loadMore) ||
+              (loadMore && !nextPageUrl)
+            ) {
+              return Promise.resolve(); // Already fetching, don't retrigger a call
```
housseindjirdeh: 👍

src/api/rest/providers/github/schemas/users.js
```diff
@@ -0,0 +1,89 @@
...
+      processed.avatarUrl = entity.avatar_url;
+
+      // These flags should be in all our schemas.
+      processed._isComplete = false; // entity not fully fetched yet
+      processed._isAuth = false; // entity doesn't belong to the auth user
+      processed._entityUrl = false; // The github url for the entity. To be used in openInBrowser()
+      processed._fetchedAt = moment().format('X');
+
+      // name is only present in full mode, we base our full parsing on its presence
+      if (typeof entity.name !== 'undefined') {
```
housseindjirdeh: Same here. Pretty much anywhere with `undefined` checks, can we not just simplify them to like in this case `if(entity.name){...}`

src/api/rest/providers/github/schemas/users.js
```diff
@@ -0,0 +1,89 @@
...
+export const userSchema = new schema.Entity(
+  'users',
+  {},
+  {
+    idAttribute: user => user.login.toLowerCase(),
+    processStrategy: entity => {
+      const processed = {};
+
+      // These are provided in both mini & full modes
+      processed.id = entity.login.toLowerCase(); // id should be always used for navigation
```
housseindjirdeh: Think we we should be able to simplify this logic with spread.

```
const processed = {
   ...entity,
   id: login.toLowerCase(),
   avatarUrl: avatar_url,
    // ....
}
```

And for keys with the same names (like `login`) we shouldn't need to have to include it as it'll be part of the spread

src/api/rest/providers/github/schemas/users.js
```diff
@@ -0,0 +1,89 @@
...
+      }
+
+      return processed;
+    },
+  }
+);
+
+/**
+
+const userFull = {
```
housseindjirdeh: Love this 👍

src/organization/screens/organization-repository-list.screen.js
```diff
@@ -0,0 +1,99 @@
+import React, { Component } from 'react';
```
housseindjirdeh: Is the idea that we'll need two separate screens (org-repo-list-screen and repo-list-screen) even if we've added normalized data throughout our store? Or is this just temporary?

src/components/repo-list-item.component.js
```diff
@@ -0,0 +1,132 @@
+import React from 'react';
```
housseindjirdeh: Why do we have this component if we already have `repository-list-item`?

src/api/rest/providers/github/client.js
```diff
@@ -0,0 +1,132 @@
...
+     * @param {string} userId
+     */
+    getEventsReceived: async (userId, params) => {
+      return this.fetch(
+        `users/${userId}/received_events`,
+        {
+          schema: Schemas.EVENT_ARRAY,
+        },
+        params
+      ).then(struct => {
```
housseindjirdeh: Don't really understand what's going on here fully 😞 

You're passing `schema` as an object? And why is `struct` defined as the data type returned?

package.json
```diff
@@ -48,11 +48,15 @@
   "dependencies": {
     "entities": "^1.1.1",
     "fuzzy-search": "^1.4.0",
+    "lodash.has": "^4.5.2",
+    "lodash.merge": "^4.6.0",
```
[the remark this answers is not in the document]
machour: I was gonna argue that we only use 4 operators, but I had a look at the node_modules/ folder, and found out that 36 operators were already included along with full lodash itself.
I'll switch to using the whole library

src/api/rest/decorators/with-reducers.js
```diff
@@ -0,0 +1,129 @@
+import * as Actions from 'api/rest/actions';
```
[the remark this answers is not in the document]
machour: I'll be available this week-end to pair and walk you through it.

Here's some bits. I'm assuming that "pure" api call is a call made without the `withReducers()` wrapper.

Basically, adding a pure api call is as simple as defining the method in the client. (in the accurate namespace).

Here's client.activity.getEventsReceived() braked down:
```js
  activity = {
    // First arguments are the ones truly required to craft the URL that will be requested.
    // 
    // The last argument should ALWAYS be params, in all methods. It's referred to as
    // "magicalArg" in withReducers(). It's an object with the following keys (for now):
    //  - loadMore: boolean. Can be passed by the user to request the next page of results
    //  - url: string. This will be automatically provided by `withReducers` when paginating, 
    //                       and will contain the next page url. If present, then fetch()
    //                       will call this url instead of `users/${userId}/received_events`
    getEventsReceived: async (userId, params) => {
      // notice that we call THIS.fech() and not the native fetch()
      return this.fetch(
         // The url that should be initially requested
         // this.fetch() will chose to ignore it and use params.url instead if provided
        `users/${userId}/received_events`, 
        {
          // the schema to be used against the returned json to normalize it
          schema: Schemas.EVENT_ARRAY,
        },
        // always pass this down to this.fetch()
        params
      ).then(struct => {
        // this.fetch() returns a structure with the following keys
        // - response: The fetch() Response object
        // - schema: The schema that was passed to it
        return {
          ...struct,
          // if we're calling a paginated endpoint, automatically compute and return next page url
          nextPageUrl: this.getNextPageUrl(struct.response),
        };
      });
    },
  };
```

> Now you may wonder why we're passing schema to this.fetch(), and then retrieve it back? 
> I'm wondering too 🤣. I'll clean that up.
> 
> The thing is, we need to retrieve the schema to normalize against along with the response :
> - If used as a pure call, the developer may find handy to have the expected response, described by the schema. (optional)
> - When used with `withReducers()`, the proxy uses this schema to normalize the JSON.
 (required))

Second, if we intended to use this newly added pure api method with `withReducers()`, we have two extra steps:
1. define an action set in actions/ (with a specific naming: ACTIVITY_GET_EVENTS_RECEIVED)
2. if the call is to be paginated, add a reducer in pagination.js as follow:
     `ACTIVITY_GET_EVENTS_RECEIVED: paginate(Actions.ACTIVITY_GET_EVENTS_RECEIVED)`,

That's it.

src/api/rest/providers/github/schemas/users.js
machour: We don't want to use the spread here, as you'll be copying *all* the original JSON attributes in the normalized entity.

I've willingly did an extra verbose function because we need to take really care of:

1. handpicking the attributes we will need, and only them => helps keep the store as slim as possible
2. name them carefully & clearly => because they will be used in all the app
3. construct them by hand if needed:
> imagine if Bitbucket returns { first_name: 'John', last_name: 'Doe' }, Gitlab returns { name: 'John', family_name: 'Doe' } and Github { fullname: 'John Doe' }
> we will have 3 schemas that all construct the following attribute: fullName: 'John Doe'

Of course, we can use some more syntax sugar here. For now I'm keeping things as readable as possible while this PR is being reviewed/challenged.

src/api/rest/providers/github/schemas/users.js
machour: This can be tricky and error prone in a schema.

If the typeof *is* undefined, then the field wasn't available in the original JSON. We use that to determine what form of the entity we received from the API (complete, minimal, etc)

If for some reason entity.name is provided, but as an empty string, the test will not do what we want.

I'll add a `isProvided()` helper to make things clearer and have only one typeof undefined check.

src/organization/screens/organization-repository-list.screen.js
machour: This was temporary but may come with a plot-twist :)

As you can see OrganizationRepositoryListScreen is now a simple container that calls a dumb `<RepositoryList />` component.

OrganizationRepositoryListScreen implements the store logic, and knows, being an organization screen, that it should read from the ORGS_GET_REPOS pagination.

For a user, we will be reading from USERS_GET_REPOS instead.

I guess it makes sense to have as many containers as "virtual" screens, and a single dumb component used by all of them. What do you think about it?

src/components/repo-list-item.component.js
machour: This is a temporary thing. I wanted to avoid cluttering `repository-list-item` with if/else to target the correct attribute. Turns out not so many things changed.

`repo-list-item` will disappear in favor of `repository-list-item`.

src/api/rest/providers/github/client.js
machour: I explained this a bit more in my reply to your first review comment.

Today, I'll be expanding this PR to the notifications screens, in order to proof test pure vs reduced api calls: 
  - retrieving notifications should be reduced and paginated
  - marking notifications read should not involve the reducer/store at all, but be a pure call

I'll let you know how it plays out!

src/api/rest/actions/activity.js
```diff
@@ -0,0 +1,5 @@
+import { createActionSet } from 'utils';
+
+export const ACTIVITY_GET_EVENTS_RECEIVED = createActionSet(
```
alejandronanez: What do you think about using `redux-actions` instead of this `createActionSet` util?

src/components/repo-list-item.component.js
```diff
@@ -0,0 +1,132 @@
...
+import { StyleSheet, Text, View } from 'react-native';
+import { ListItem, Icon } from 'react-native-elements';
+
+import { emojifyText, abbreviateNumber } from 'utils';
+import { colors, languageColors, fonts, normalize } from 'config';
+
+type Props = {
```
alejandronanez: Curious, how do we check for `flow` if we don't have `// @flow` on line 1? I think you already answered this but I don't know exactly where.

alejandronanez:
One concern since this is a huge piece of work.
What you guys think about using `redux-actions` for the sake of reducing boilerplate?

src/api/rest/decorators/with-reducers.js
```diff
@@ -0,0 +1,128 @@
...
+import { Alert } from 'react-native';
+
+const displayError = error => {
+  Alert.alert('API Error', error);
+};
+
+export const withReducers = Provider => {
+  const client = new Provider();
+
+  return new Proxy(withReducers, {
```
alejandronanez: I'm sorry but I don't understand what is going on here =[ I think it's way too hard to follow, what about splitting functionality in small pieces that are easier to test/follow?

src/api/rest/actions/activity.js
machour: The way it's done in this PR, actions are barely seeable already : 
 - defined in [rest/actions/](https://github.com/gitpoint/git-point/pull/457/files#diff-40791c1b0969e259b980e54d84481d84)
 - mapped to paginations in [rest/reducers/paginations.js](https://github.com/gitpoint/git-point/pull/457/files#diff-0188bb8ed2ff83caf951adfd9b83aa7c)

What would be the benefit from using `redux-actions`? (I'm not quiet familiar with it)

src/api/rest/decorators/with-reducers.js
lex111: @machour by the way, it's also hard for me to understand this yet, but I'll study it on the weekend. And why `withReducers`?

src/components/repo-list-item.component.js
lex111: @alejandronanez we have a lot where there is not this, we need to create a separate PR with the addition of full support Flow throughout the project.

alejandronanez: Oh...got it! Thanks @lex111

src/api/rest/actions/activity.js
alejandronanez: Here's a PR that uses `redux-actions` #269

alejandronanez: Redux Actions will reduce verbosity introduced by default by Redux. You can see the motivation here https://redux-actions.js.org/docs/introduction/Motivation.html

src/api/rest/decorators/with-reducers.js
machour: @alejandronanez I'm currently extracting functions on my working copy :)

Here's some explanations:

The basic client, used like that:
`client.repos.getById("git-point")`
fetches() the url from github, and returns the Response (+ some meta data).
No redux, store, actions, anything involved. A simple Rest client.

`withReducers()` is a proxy around `client`, that serves the following purpose:

1. intercept the client call (`client.repos.getById("git-point")`)
2. compute the corresponding action name/pagination name automatically (reduce boiler plate, standardize naming)
3. check if the requested data is already in store, if that's the case, return; 
4. if the data is not available:
  - If it's a paginated call, check the pagination status, and optionally override the API URL that will be called
5. dispatch PENDING & trigger the real request (as it would have been the case without `withReducers()`
6. normalize the received JSON data, and trigger SUCCESS for it to be stored

It's a lot of things indeed, but it's actually the WHOLE stuff, centralized in one place.

Don't hesitate if you need clarifications mate (/cc @lex111)

src/api/rest/actions/activity.js
machour: I understand the motivation here, but against master, not this PR :/

alejandronanez:
@all Guys, I'm stepping aside on this PR, I think I don't have enough context right now to make a good decision about what should/should not be done and this is becoming larger (100+ comments + 1.8k lines is already quite difficult to catch up)

machour:
@alejandronanez I totally understand your concern about the size of the PR..
It's a big piece, and it have to be actually pulled and tested step by step by the reviewer.
The initial comments I initially made were there to answer questions that would may have arise while trying it.
Sorry if it or I got too verbose.

I'm supposed to have a one on one review with @housseindjirdeh tomorrow about this proposal.
I'll do that and see how he feels about it. Maybe things can be split across various PRs and be clearer.
Or maybe my use of redux is unholy, and it should be dismissed all together.

src/api/rest/decorators/with-reducers.js
```diff
@@ -0,0 +1,128 @@
...
+
+          // Get accessToken from state
+          client.setAccessToken(getState().auth.accessToken);
+
+          let finalArgs = args;
+
+          if (paginator) {
+            const { loadMore = false } = magicArg;
+            const { pageCount = 0, isFetching = false, nextPageUrl } =
+              paginator[pureArgs.join('-')] || {};
```
lex111: Why need `.join('-')`?

src/api/rest/decorators/with-reducers.js
```diff
@@ -0,0 +1,128 @@
...
+  Alert.alert('API Error', error);
+};
+
+export const withReducers = Provider => {
+  const client = new Provider();
+
+  return new Proxy(withReducers, {
+    get: (c, namespace) => {
+      return new Proxy(client[namespace], {
+        get: (endpoint, call) => (...args) => (dispatch, getState) => {
```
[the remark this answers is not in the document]
machour: You're right, that should be "method"

Could you check the last version I just pushed for namings ?

src/api/rest/decorators/with-reducers.js
```diff
@@ -0,0 +1,128 @@
...
+          const action = Actions[actionName];
+
+          if (!action) {
+            return displayError(
+              `Unknown action. Did you forget to define Actions.${actionName}?`
+            );
+          }
+
+          // Identify if we have our magical last argument
+          const declaredArgsNumber = endpoint[call].length;
```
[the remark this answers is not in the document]
machour: Yes.

The only sad thing, is that Function.length doesn't return accurate numbers for optional parameters:

```js
const = (foo, bar); // Function.length = 2
const = (foo, bar = {}); // Function.length = 1

File: src/api/rest/decorators/with-reducers.js [hunk 46]
```diff
@@ -0,0 +1,128 @@
...
+          const isMagicArgAvailable = args.length === declaredArgsNumber;
```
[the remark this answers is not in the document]
machour: I really didn't know how to name this extra last argument which should be implemented in all methods definition.. 
This is an argument that is always at the end, and is used for both internal & external stuff. AdditionalArg doesn't seem to describe it enough

src/api/rest/decorators/with-reducers.js
machour: example:

```js
client.activity.getNotifications(false, false); // unread
client.activity.getNotifications(true, false); // participating
```

We need to paginators here, and to identify them uniquely we use .join('-') to create the key: 

screen shot 2017-10-15 at 6 38 12 pm

src/api/rest/providers/github/client.js
```diff
@@ -0,0 +1,132 @@
+import { Client } from '../base';
+import Schemas from './schemas';
+
+export class Github extends Client {
```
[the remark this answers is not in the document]
machour: fine with me

src/api/rest/providers/github/client.js
```diff
@@ -0,0 +1,132 @@
...
+  API_ROOT = 'https://api.github.com/';
+
+  setAccessToken = accessToken => {
```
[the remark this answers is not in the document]
machour: This was made thinking about future Gitlab / Bitbucket implementation.
We `setAccessToken` and the client is responsible for using it in headers, query string, cookie ..

src/api/rest/providers/github/client.js
```diff
@@ -0,0 +1,132 @@
...
+    this.authHeaders = { Authorization: `token ${accessToken}` };
+  };
+
+  getNextPageUrl = response => {
```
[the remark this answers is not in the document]
machour: This is correct for Github, but for Gitlab for example, next link information is contained in the JSON response. This is why I decided to pass `Response`

src/api/rest/providers/github/client.js
```diff
@@ -0,0 +1,132 @@
...
+
+  /**
+   * The organizations endpoint
+   */
+  orgs = {
+    /**
+     * Gets an organization by its id
+     *
+     * @param {string} orgId
+     */
```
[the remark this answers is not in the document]
machour: Description for `params` will be the same for all methods. I wanted to avoid duplicating doc comments before finding a good name.

The JSDoc is in Github for now, but I'll move it to the Rest class for later. (We're going to add all the blanks prototypes there along with the JSDocs to be able to generate an API documentation like this: https://octokit.github.io/node-github/#api-activity-markNotificationThreadAsRead)

Does that sound ok to you?

src/api/rest/providers/github/client.js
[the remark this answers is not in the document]
lex111: I understand, but in fact we set the Auth header, just if it was renamed, it was clearer.

In Bitbucket or Gitlab, the token is not specified in the header?

src/api/rest/providers/github/client.js
```diff
@@ -0,0 +1,132 @@
...
+  /**
+   * The activity endpoint
+   */
+  activity = {
+    /**
+     * Gets received events
+     *
+     * @param {string} userId
+     */
+    getEventsReceived: async (userId, params) => {
```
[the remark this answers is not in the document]
machour: activity is a reflection of GitHub's endpoint, which contains events, notifications, watching, etc : https://developer.github.com/v3/activity/

I'm sticking with node-github naming mostly for now, as it almost reflects the URL called

src/api/rest/decorators/with-reducers.js
lex111: Oh, I see, although it's not very clear.

src/components/entity-info.component.js
```diff
@@ -143,6 +143,22 @@ export const EntityInfo = ({ entity, orgs, language, navigation }: Props) => {
           onPress={() => Communications.web(getBlogLink(entity.blog))}
           underlayColor={colors.greyLight}
         />}
+
+      {!!entity.webSite &&
```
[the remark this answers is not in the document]
machour: The block above should be removed once the new API is implemented on Users.
For now I duplicated the block instead of putting 3 `entity.web_site ? entity.web_site : entity.webSite` in the existing block

src/api/rest/decorators/with-reducers.js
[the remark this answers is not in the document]
lex111: I think it would be easier if we considered that there could be more than one `magic` parameters, the more it is quite possible, given how it is currently used:

```js
const { loadMore = false } = magicArg; // then maybe `param2`, `param3`... and these are additional parameters
```

src/components/entity-info.component.js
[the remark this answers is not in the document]
machour: @lex111 left you some more comments on gitter to avoid writing another page here :)

src/api/rest/providers/github/schemas/events.js
```diff
@@ -0,0 +1,32 @@
...
+export const eventSchema = new schema.Entity(
+  'events',
+  {
+    actor: userSchema,
+    org: orgSchema,
+    repo: repoSchema,
+  },
+  {
+    idAttribute: event => event.id,
+    processStrategy: entity => {
```
jouderianjr: I think that can be refactor in order to avoid `processed.` repetition

```
proccessStrategy: entity => ({
    ...initSchema(),
   id: entity.id,
   ....
```

src/api/rest/providers/github/schemas/repos.js
```diff
@@ -0,0 +1,162 @@
...
+    idAttribute: repo =>
+      (isInMinimalisticForm(repo) ? repo.name : repo.full_name).toLowerCase(),
+    processStrategy: entity => {
+      const processed = initSchema();
+
+      // Repo received from events
+      if (isInMinimalisticForm(entity)) {
+        processed.id = entity.name.toLowerCase();
+        processed.fullName = entity.name;
+        processed.shortName = entity.name.substring(
```
jouderianjr: It is possible have more than one '/' in the string ?

machour: not for a repo

src/api/rest/providers/github/schemas/events.js
machour: Thanks a lot for your review man! Just went through your remarks and applied the suggestions. I didn't know about `String.includes`, nice one :)

I just kept EventSchema like that for the moment, cause it still needs some work that may require that `processed` const. All of this will be polished at the end of the PR.

machour:
Dears,

After validating this PR principle with @housseindjirdeh & @lex111, and after hearing from @jouderianjr & @alejandronanez I came up with a smoother plan to integrate this PR:

- This big and barely readable/reviewable PR is closed as of now
- I'll wait for tests to be merged in master (#407)
- I'll do a new & smaller PR which includes:
  - The `api/rest/` folder, which handles all the API/Redux logic
  - An internal documentation in markdown format
  - Tests for schemas transformations
  - The notifications section of the app redone with the new redux approach
  - 0 initial comments from my part (😆) 

Once this new PR is digested by everyone & merged in, we'll set a plan to migrate the rest of the App.

Thank you all for your constructive feedback!

<3

Issue #309, opened 2017-09-06
Hide readme button if no readme


CameronSamuels:

housseindjirdeh:
Agreed this should be relatively simple as well :+1:

chinesedfan:
Because no existing field in `repository` object, we should move `getReadMeByDispatch` from `read-me.screen` to `repository.screen`.

- If blocking loading, `repository.screen` may become slower.
- If hiding readme button after the request is returned, users may find ui changed suddenly.

I prefer the second, how about you?

machour:
@chinesedfan I prefer the second option too, but starting with a hidden README button. If there's a readme, we update the UI to show it (like we do for the "Merge" button).

PR #459, opened 2017-10-09
fix(translation): add missing string to translations

FlorianTaut:
Hey guys,

i found a string in `readme.screen.js`
I added it to the translation files but the translators need to translate it now.

Im not sure, probably @housseindjirdeh wanted this string to be not translate or it was just overseen somehow. If its wanted this way please close this pr.

Thank you guys

### DIFF
--- a/src/locale/languages/es.js
+++ b/src/locale/languages/es.js
@@ -222,6 +222,7 @@ export const erase = {
       fileRenamed: 'File renamed without any changes',
     },
     readMe: {
+      readMeActions: 'README Actions',
       noReadMeFound: 'No README.md found',
     },
   },
--- a/src/locale/languages/es.js
+++ b/src/locale/languages/es.js
@@ -222,8 +222,8 @@ export const erase = {
       fileRenamed: 'File renamed without any changes',
     },
     readMe: {
-      readMeActions: 'README Actions',
-      noReadMeFound: 'No README.md found',
+      readMeActions: 'Acciones README',
+      noReadMeFound: 'No se ha encontrado README.md',
     },
   },
   organization: {
--- a/src/locale/languages/es.js
+++ b/src/locale/languages/es.js
@@ -221,6 +221,9 @@ export const erase = {
       deleted: 'DELETED',
       fileRenamed: 'File renamed without any changes',
     },
+    readMe: {
+      noReadMeFound: 'No README.md found',
+    },
   },
   organization: {
     main: {
--- a/src/locale/languages/fr.js
+++ b/src/locale/languages/fr.js
@@ -228,7 +228,7 @@ export const fr = {
     },
     readMe: {
       readMeActions: 'Actions sur le README',
-      noReadMeFound: 'No README.md found',
+      noReadMeFound: 'Pas de README.md trouvé',
     },
   },
   organization: {
--- a/src/locale/languages/fr.js
+++ b/src/locale/languages/fr.js
@@ -228,6 +228,7 @@ export const fr = {
     },
     readMe: {
       readMeActions: 'Actions sur le README',
+      noReadMeFound: 'No README.md found',
     },
   },
   organization: {
--- a/src/locale/languages/pt.js
+++ b/src/locale/languages/pt.js
@@ -229,8 +229,8 @@ export const pt = {
       fileRenamed: 'O ficheiro mudou de nome sem qualquer alteração',
     },
     readMe: {
-      readMeActions: 'README Actions',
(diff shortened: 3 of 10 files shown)


FlorianTaut:
So i need you guys to translate this string please:
`repository.readMe.noReadMeFound`

Thank you @siderio2 @Jpfonseca @jouderianjr @machour @lex111 @alejandronanez @RolfKoenders

src/locale/languages/es.js
```diff
@@ -221,6 +221,9 @@ export const erase = {
       deleted: 'DELETED',
       fileRenamed: 'File renamed without any changes',
     },
+    readMe: {
+      noReadMeFound: 'No README.md found',
```
Antoine38660: Could you add the `readMeActions` key? It exist in en.js :)

src/locale/languages/pt.js
```diff
@@ -228,6 +228,9 @@ export const pt = {
       deleted: 'REMOVIDO',
       fileRenamed: 'O ficheiro mudou de nome sem qualquer alteração',
     },
+    readMe: {
+      noReadMeFound: 'No README.md found',
```
Antoine38660: Same here about the `readMeActions`

src/locale/languages/es.js
FlorianTaut: Yeah, the hole `readMe` part was missing in es, not sure why.

FlorianTaut:
Thank you @Antoine38660 now we only need the translations for this string or should we merge without and fix it with following pr's? 👍

src/locale/languages/ru.js
```diff
@@ -229,6 +229,7 @@ export const ru = {
     },
     readMe: {
       readMeActions: 'Действия с README',
+      noReadMeFound: 'No README.md found',
```
vitalkanev: Should be `Не удалось найти README.md`

src/locale/languages/fr.js
```diff
@@ -228,6 +228,7 @@ export const fr = {
     },
     readMe: {
       readMeActions: 'Actions sur le README',
+      noReadMeFound: 'No README.md found',
     },
```
machour: French translation: `Pas de README.md trouvé`

src/locale/languages/ru.js
FlorianTaut: Commited! Please check it out 👍

src/locale/languages/fr.js
FlorianTaut: Thank you @machour 

Also done, please check it out! 👍

chinesedfan:
@FlorianTaut Thanks for your work! But we may remove this string because of #309.

FlorianTaut:
Hey @chinesedfan, im not sure why this should get in any conflict with #309, could you explain me why please? 👍

chinesedfan:
@FlorianTaut Sorry, I didn't write very clear. I mean, with #309, it is nearly impossible that no ReadMe happens. Because users have no chance to see the entry button first.

FlorianTaut:
@chinesedfan Yeah its possible that this is never shown but we have no string directly written into code, so i think we should do it at this point as well. Dont you think so either?

chinesedfan:
@FlorianTaut If we can confirm, I prefer to remove noReadMe related logic. Then we can skip translating this string. Now we have so many languages to support. 😆  (cc @housseindjirdeh)

FlorianTaut:
@chinesedfan im okay with both, lets wait for @housseindjirdeh‘s opinion and then we can go on with this pr

siderio2:
Should I add the line to the translation or wait??? And if I have to add, how? New PR??

FlorianTaut:
@siderio2 just leave a comment with the translated string please! 👍

siderio2:
This is the Galician translation:
readMe: {
       readMeActions: 'Accións do README ',
       noReadMeFound: 'Non se atopou o README.md',
      },

jouderianjr:
pt-br `noReadMeFound: 'Nenhum README.md foi encontrado'` :dancer:

housseindjirdeh:
Thanks everyone <3

@FlorianTaut what do you think of just leveraging Google Translate for the remainder of the languages and leaving a comment? Comment should allow us to come back to them later just to make sure strings are okay

FlorianTaut:
@housseindjirdeh yes, i will do that today, tried to answer yesterday but gitpoint crashes on every comment 😞

housseindjirdeh:
Thank you @FlorianTaut <3

Are you noticing the constant crashing only on Android? :(

PR #463, opened 2017-10-10
fix(issue): Consolidate issue link parsing

### LINKED ISSUE #462: Crash on issue comments
Hey guys,

i tried to open a notification with gitpoint.
The app crashed immeditely.

I tried it with 2 different notifications.

Somebody can reproduce this crash?

Have iOS 11.0.2 on an iPhone 7 running latest GitPoint version

chinesedfan:
~~Fixes~~ Discovered in #462.

The regular expression can't match anything if `issueURL` is a issue comment, like `https://github.com/octokit/discussions/issues/7#issuecomment-327616339`. It causes `[1]` throws an error.

```
const issueNumberRegex = /issues\/([0-9]+)$/;
const { issue, issueURL, isPR, language } = navigation.state.params;
const number = issue ? issue.number : issueURL.match(issueNumberRegex)[1];
```

### DIFF
--- a/routes.js
+++ b/routes.js
@@ -132,7 +132,7 @@ const sharedRoutes = {
   Issue: {
     screen: IssueScreen,
     navigationOptions: ({ navigation }) => {
-      const issueNumberRegex = /issues\/([0-9]+)$/;
+      const issueNumberRegex = /issues\/([0-9]+)(#|$)/;
       const { issue, issueURL, isPR, language } = navigation.state.params;
       const number = issue ? issue.number : issueURL.match(issueNumberRegex)[1];
       const langKey = isPR ? 'pullRequest' : 'issue';


machour:
Nice catch buddy, but I think you're fixing another unknown bug here.

#462 is a bug happening when reading an issue. There's no navigation problem related there, but an Animated related exception (see my comments on that issue). If I dismiss the red screen, I do see the issue screen along with some of its comments. The bug there is supposedly with a TouchOpacity element. Still haven't investigate it yet.

machour:
Oh, actually you fixed a bug happening when clicking on the link I added in my comment :D

machour:
Ok, I understood what happened here :D 

There is the bug reported in #462, which is still un-resolved.
And there's a new bug you spotted by clicking my comment on #462.

chinesedfan:
@machour Right. I saw the crash you mentioned for a while, but was attracted by this bug. 😆 

I have changed the PR description so that this PR doesn't close #462.

PR #466, opened 2017-10-10
feat(translation): Add Polish translation

peterblazejewicz:
This commits adds the first version of the files required to provide a Polish translation
strings for the GitPoint. Tested locally using iOS simulator.

The commit is rebased against recent addition to l10n support.

Thanks!


FlorianTaut:
@peterblazejewicz Thank you very much!

PR #468, opened 2017-10-11
chore: disable OS_ACTIVITY_MODE in Debug mode

machour:
When using xcode, logs are flooded with this message (X incrementing every 2 seconds):
`[] nw_connection_get_connected_socket_block_invoke X Connection has no connected handler`

setting OS_ACTIVITY_MODE = disable fixes it, and logs are bearable again


machour:
I feel you, those continuous logs made me nuts! Glad I don't have to reconfigure it everytime now <3

Issue #439, opened 2017-10-06
Translation for other languages


lex111:
Hello everyone!

We need new translations, **the instruction to add is [here](https://github.com/gitpoint/git-point/wiki/I18N#adding-a-new-language)**.

At the moment **existing** translations:
- English
- French
- Dutch
- Turkish
- Brazilian Portuguese
- Russian
- Galician (thanks to @siderio2!)
- Spanish
- Portuguese (thanks to @Jpfonseca!)
- Portuguese Brazilian
- German (thanks to @tautf!)
- Polish (thanks to @peterblazejewicz!)
- Esperanto (thanks to @kodisto!)
- Ukrainian
- Chinese Traditional
- Chinese Simplified
- Basque
- Swedish (thanks to @SpaceLenore)
- Tagalog (thanks to @geeofree)
- Serbian (thank to @cobisimo)

We are looking forward to the _new_ translations, this will be a good start for Hacktoberfest, so welcome!

**P.S. Only native speakers, please.**

khvilaboa:
Spanish coming soon... :)

lex111:
@khvilaboa it is very good! However, we already have PR #282 for translating into Spanish, but it is not finished, and you can take it as a basis and create your own PR with already ready translation, I hope, @alejandronanez  will not mind, right? @alejandronanez  will then check the accuracy of the translation in that case.

stephenpfaff:
So you need **_any_** language which does not already exist then or specific ones?

lex111:
@stephenpfaff we will be glad to **any** language for which there is no translation yet.

tminussi:
I would be happy to help with pr-br

housseindjirdeh:
Thank you all so much, _any language_ not present will be AMAZING

Rishabh04-02:
I'd like to help adding **de** - German.

andrewda:
@Rishabh04-02 Awesome! Thanks a lot!

FlorianTaut:
@Rishabh04-02 I'm already working on german Translation, but you can make this your Job if you want to. Thank you :+1:

Rishabh04-02:
@FlorianTaut actually I've started to work on this. But so far have done only 10% of the job. So, if your work has progressed more than mine then you can work on it, I'll work on something else. But please let me know, if you're working on this.

viettrung9012:
I can help with translation to Vietnamese

phillip055:
Thai!

[...]

PR #471, opened 2017-10-13
feat(translation): add Esperanto translation

kodisto:
For issue [ #439](https://github.com/gitpoint/git-point/issues/439), translation of Esperanto


src/auth/screens/language-settings.js
```diff
@@ -35,6 +35,10 @@ export default [
     name: 'Русский',
   },
   {
+    code: 'eo',
+    emojiCode: ':star:',
```
andrewda: I think this should probably be a flag, even if it's `:waving_white_flag:`.

housseindjirdeh: Yeah @lex111 and i have discussed how we can possibly change this, but in the meantime `:waving_white_flag:` should do :)

housseindjirdeh:
Thank you @kodisto 🙏  Only thing is the flag emoticon, everything else looks solid 🎉

chinesedfan:
@housseindjirdeh This PR breaks ESLint. I will fix it in #478.

PR #476, opened 2017-10-14
fix(locale): Check before configuring locale

### LINKED ISSUE #436: Crash if the device language is Chinese
Bad news! After updated to 1.3, I can't open GitPoint any more on my iPhone. It's my mistake that only testing in simulator.

The following [function](https://github.com/gitpoint/git-point/blob/master/src/locale/index.js#L24) is the culprit. In my iPhone, `I18n.locale` returns `zh-Hans-CN`, but `zh` is not a valid locale for `moment`.
```
export const getLanguage = () => I18n.locale.substr(0, 2);
```

Then it causes a crash.

11507295307_ pic

A workaround is to set device language to English first, open GitPoint successfully and choose any language in options manually. Then recover the device language.

@housseindjirdeh I find it's hard to implement `getLanguage`, so I suggest to just return `en` in it.

chinesedfan:
Fixes #436.

If the provided locale is not supported by `moment`, just use `en`.

### DIFF
--- a/src/utils/method-helpers.js
+++ b/src/utils/method-helpers.js
@@ -35,11 +35,13 @@ export const delay = (delayed, ms) =>
 export const translate = (key, lang, interpolation = null) =>
   I18n.t(key, { locale: lang, ...interpolation });
 
-export const configureLocale = language => {
-  I18n.locale = language;
+export const configureLocale = locale => {
+  const validLocale = moment.locales().indexOf(locale) < 0 ? 'en' : locale;
+
+  I18n.locale = validLocale;
   I18n.fallbacks = true;
 
-  moment.updateLocale(language, {
-    relativeTime: translate('common.relativeTime', language),
+  moment.updateLocale(validLocale, {
+    relativeTime: translate('common.relativeTime', validLocale),
   });
 };


lex111:
@chinesedfan thanks for the fix, but I'm already preparing PR with fixing this bug.

In your PR, Chinese language will never be used, because its locale needs to be changed separately, therefore it is only the elimination of the error, but nevertheless the problem with the fact that the language does not apply will not be fixed, it will not be Chinese language, but English. It took me more time to fix, because I'm still doing the refactoring of the current localization, tomorrow I will send PR.

chinesedfan:
@lex111 In fact, I am a little confused about the relationship between GitPoint locale and moment locale. If GitPoint doesn't support Chinese, why should we set moment locale as Chinese?

lex111:
@chinesedfan you are right,  I will consider this.

housseindjirdeh:
Thanks a million for this @chinesedfan, was this or something similar addressed in your PR @lex111?

chinesedfan:
@housseindjirdeh Ha, this thing has been fixed in #479.

PR #479, opened 2017-10-15
fix(locale): make correct locale use

lex111:
Fixes #463

First of all, the correct definition of the locale is fixed. For example, we have a `pt-br` locale, but because the locale used to be splited to just `pt`, we actually had used an English locale.
Further, specific locales for the Chinese language are added, or there was an error with the moment.
And also replaced everywhere `language` with `locale`, as this is more correct and moved the methods of localization in a separate helper.


src/utils/localization-helper.js
```diff
@@ -0,0 +1,49 @@
...
+    'zh-hans': 'zh-cn',
+    'zh-hans-cn': 'zh-cn',
+    'zh-hans-sg': 'zh-cn',
+    'zh-hans-hk': 'zh-hk',
+    'zh-hant': 'zh-tw',
+    'zh-hant-tw': 'zh-tw',
+    'zh-hant-hk': 'zh-tw',
+  };
+
+  return specialLocales[locale] || locale;
```
chinesedfan: @lex111 We still need something like `I18n.locale.substr(0, 2)`. Maybe separate by '-' and remove the last part.

If I change my iPhone device language to French, `I18n.locale` returns `fr-zh`. But it is not supported by `moment`.

lex111: @chinesedfan No, it is not necessary, it was a mistake. For example, in the case of `pt-br`, it will be just `pt`, but we need `pt-br`, because translation is only for him.

> If I change my iPhone device language to French, I18n.locale returns fr-zh. But it is not supported by moment.

There is a [check](https://github.com/gitpoint/git-point/pull/479/files#diff-aa84d914cd1a32be6d7b8abcb41cf186R37) if there is such a locale in `moment`, and if not, then the default locale (English).

chinesedfan: @lex111 I mean, `fr-zh` should also result in `fr`, instead of `en`.

machour:
@lex111 Seems really good to me all in all, but are we *obliged* to have that `language` => `locale` terminology change?

I understand the motivation; and using "locale" is indeed more correct, but it's going to cause conflicts in almost all opened PRs just for semantics.. (I'm looking at #457 and silently crying 🤣 )

src/utils/localization-helper.js
lex111: @chinesedfan I fixed it, look now.

lex111:
> @lex111 Seems really good to me all in all, but are we obliged to have that language => locale terminology change?

Yes, I understand that diff has become big, but it seems to me important to rename it, because it eliminates confusion currenly. Languages are now just a list items on the screen, and the locale is the key in the store.

machour:
@lex111 a little late to the party, sorry buddy:

I'm curious, why is `getAuthLocale` used only in organization-profile.screen.js ?

I had a error after merging master in my redux PR, switched back to reading from state.auth.locale in mapStateToProps and it's working again. Is `getAuthLocale` really needed?

lex111:
@machour I'm not sure, I just renamed this, it's @alejandronanez did, it's better to ask him.

PR #487, opened 2017-10-17
Hacktoberfest

gilemerson:
Commit One

### DIFF
--- a/__tests__/index.android.js
+++ b/__tests__/index.android.js
@@ -1,3 +1,5 @@
+/*Commit One*/
+
 import 'react-native';
 import React from 'react';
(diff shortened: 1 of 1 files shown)


andrewda:
Thanks for your interest in GitPoint! Please don't create spam PR's though – that's not what Hacktoberfest is about. The whole point of Hacktoberfest is for quality PR's to be contributed to as many repositories as possible, but submitting spam cheats the system and yourself.

That being said, we'd gladly accept any PR's – big or small – which benefit not only the app itself, but also the documentation or development experience.

GantMan:
https://twitter.com/emerg5/status/920081081554026496

PR #490, opened 2017-10-17
fix: Fix the last avatar being truncated

machour:
Also made size a variable


machour:
Woops, this included an unintended commit

Issue #406, opened 2017-10-02
Write action/API tests


andrewda:
I'd like to keep this separate from #148 – ideally we'd get that merged in first. Afterwards, we can work on testing actions and GitHub API interactions. Obviously we won't want to use the production GitHub API as we'd get rate limited very quickly, so we might want to run our own mock GitHub API, similar to https://github.com/mzabriskie/mock-github-api.

machour:
Just came across [jest-fetch-mock](https://www.npmjs.com/package/jest-fetch-mock) which may come really handy too

machour:
@andrewda Could we integrate tests for now without upgrading to React 16?

From what I understand, enzyme is the show stopper here, and we have to wait for all our dependencies to be R16 compatible being able to use it.

Jest snapshots could be a quiet solid alternative while waiting for packages to be upgraded by their maintainers. The only thing that we won't be able to test for now, as far as I understand it, is clicking/touching elements & their callbacks. 

I would happily live without that for now, and we'd finally have tests implemented in master.

🙏  🙏  🙏  ❤️

andrewda:
@machour Actually, #407 should be almost ready to go. Only thing that's missing there is more tests, but we can add those as we go. There should be enough there for people to get started writing more tests.

machour:
@andrewda when I pull the `tests` branch and run it on the simulator, I get this error

![screen shot 2017-10-17 at 12 43 31 pm](https://user-images.githubusercontent.com/304450/31663141-49ba6c34-b341-11e7-912b-512b9c2f9cf2.png)

rn-app-intro uses a really old fork & unmaintained fork of react-swiper. Is that the problem?

andrewda:
:thinking: Yea looks like that's caused by something in `rn-app-intro`. It looks like there's a new version out (`0.0.5`), so could you try upgrading to that?

machour:
That new version is completely broken: https://github.com/Ernoff/rn-app-intro/issues/1

andrewda:
:frowning: Alright, hopefully that can be fixed soon! As for snapshot tests which you mentioned earlier, I'm not a huge fan primarily because it requires redoing the tests after every minor change to a component. This could be fine to prevent unwanted changes, but eventually people will get into a habit of automatically re-snapshotting without testing first. It's better than nothing, I guess, but I'd like to get Enzyme tests incorporated as soon as possible.

soytjan:
[...]

PR #493, opened 2017-10-17
refactor: Drop rn-app-intro in favor of react-native-swiper

machour:
We weren't using any of the advanced features of react-native-app-intro.
This should fix the compatibility issue for #406

### DIFF
--- a/src/auth/screens/login.screen.js
+++ b/src/auth/screens/login.screen.js
@@ -14,7 +14,7 @@ import {
   Modal,
 } from 'react-native';
 import { Button, Icon } from 'react-native-elements';
-import AppIntro from 'rn-app-intro';
+import Swiper from 'react-native-swiper';
 import queryString from 'query-string';
 
 import { ViewContainer } from 'components';
@@ -98,6 +98,10 @@ const styles = StyleSheet.create({
     flex: 2,
     justifyContent: 'center',
     alignItems: 'center',
+  },
+  signInContainer: {
+    position: 'absolute',
+    bottom: 80,
   },
   slide: {
     flex: 1,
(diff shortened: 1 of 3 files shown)


machour:
Failing on android (all white)

src/auth/screens/login.screen.js
```diff
@@ -229,130 +243,120 @@ class Login extends Component {
...
-                        this.setModalVisible(!this.state.modalVisible)}
-                    />
-                  </View>
-                </View>
-              </Modal>
-              <View style={styles.miniSection}>
-                <Image
-                  style={styles.logo}
-                  source={require('../../assets/logo.png')}
+        {this.shouldDisplayIntro() && (
```
lex111: How about moving the code to a separate component, such as `Intro`?

```js
{this.shouldDisplayIntro() && <Intro />}
```

src/auth/screens/login.screen.js
```diff
@@ -229,130 +243,120 @@ class Login extends Component {
...
+              <Text style={styles.title}>
+                {translate('auth.login.issuesTitle', locale)}
+              </Text>
+              <Text style={styles.message}>
+                {translate('auth.login.issuesMessage', locale)}
+              </Text>
+            </View>
+          </Swiper>
+        )}
+        {this.shouldDisplayIntro() && (
```
lex111: Why this check? Can we combine into one?

src/auth/screens/login.screen.js
machour: I'm not sure about this, it would be used only for the login screen (+comment below)

src/auth/screens/login.screen.js
machour: We sadly cannot, `<Swiper>` don't work at all in Android when nested into a view, that's why I need these multiple checks/rendering

PR #495, opened 2017-10-18
feat(styling): add styled components

alejandronanez:
Adding `styled-components` initial work! 🙌

### DIFF
--- a/App.js
+++ b/App.js
@@ -1,13 +1,7 @@
 import React, { Component } from 'react';
 import { Provider } from 'react-redux';
-import {
-  AppRegistry,
-  AsyncStorage,
-  Image,
-  StyleSheet,
-  View,
-  LayoutAnimation,
-} from 'react-native';
+import styled from 'styled-components/native';
+import { AppRegistry, AsyncStorage, LayoutAnimation } from 'react-native';
 import { persistStore } from 'redux-persist';
 import createEncryptor from 'redux-persist-transform-encrypt';
 import DeviceInfo from 'react-native-device-info';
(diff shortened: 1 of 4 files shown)


App.js
```diff
@@ -20,18 +14,17 @@ import { determineLanguage } from 'locale';
...
-    alignItems: 'center',
-  },
-  logo: {
-    width: 100,
-    height: 100,
-  },
-});
+const Container = styled.View`
+  align-items: center;
+  background-color: ${() => colors.white}
```
chinesedfan: @alejandronanez I am not sure. Can it just be `background-color: 'white';` ?

andrewda: Doing it this way makes it so we can change the white color everywhere by just changing the a single value. However, why can't we just do `${colors.white}` @alejandronanez?

alejandronanez: Hey @chinesedfan 
- We should avoid at all costs using plain strings for colors, that's not a good practice.
- We already have a `colors` file that we should keep using.

chinesedfan: @alejandronanez Yes, my point is the anonymous function, as @andrewda mentioned. (You merged so quickly. 😆 )

alejandronanez: @andrewda you're right! There you go https://github.com/gitpoint/git-point/pull/497/files

alejandronanez: Oh, I put a fix for that too @chinesedfan! :)

ericadamski: Understanding that this is already merged, there is no need for a function here, you can just use string interpolation

```javascript
background-color: ${colors.white};
```

Issue #492, opened 2017-10-17
Better way to indicate members list can be scrolled


chinesedfan:
Now it only depends on the last avatar is not shown completely, and is not guaranteed in different screen widths of devices.

machour:
Done in #499, what do you think about it?

machour:
Closed in #529 🎉

PR #499, opened 2017-10-18
feat: Flash members list when it's scrollable

machour:
Addresses #492 by flashing the FlatList. Scrolling indicator is disabled afterward.

gif

### DIFF
--- a/src/components/members.component.js
+++ b/src/components/members.component.js
@@ -90,7 +90,19 @@ const MembersListComponent = ({
     <FlatList
       style={styles.flatList}
       data={members}
-      showsHorizontalScrollIndicator={false}
+      ref={ref => {
+        this.listRef = ref;
+      }}
+      showsHorizontalScrollIndicator
+      onContentSizeChange={() => {
+        const list = this.listRef;
+
+        list.flashScrollIndicators();
+        setTimeout(
+          () => list.setNativeProps({ showsHorizontalScrollIndicator: false }),
+          1000
+        );
+      }}
       renderItem={({ item }) => (
         <TouchableHighlight
           onPress={() => {


machour:
@housseindjirdeh yes, only if the list contains hidden elements.
This is to response to @chinesedfan concern on #492, but maybe this isn't the best solution yet :)

housseindjirdeh:
Ah just read that issue now.

Hmm I agree that's not immediately intuitive that the list can be scrolled _but_ I don't think flashing the indicator is a good idea. If the indicator was slightly padded away from the avatars than maybe but I think it looks kind of weird the way it's squeezed in there at the bottom of the avatars. Personally I would rather we think about a different solution.

CC @chinesedfan, let us know what you think.

chinesedfan:
@machour Thanks for your quick effort! But I am sorry that the flash design is very satisfied for me. My current simple idea is try to add a [triangle](https://octicons.github.com/icon/triangle-right/). What about ideas of you guys?

cc @housseindjirdeh.

machour:
I guess flashing scrollbars is a dead UX pattern nowadays :D 

Will happily change to something else!

ocarreterom:
How about to add a small transparent-to-white gradient at the end/start of scrollview? I can share a screenshot if it's necessary.

machour:
I'd love to see that screenshot @ocarreterom

ocarreterom:
Before:
img_4e39490dc472-1

After:
after

And show/hide de gradient left/right if the start/end of the scroll is reached.

machour:
@ocarreterom now that's way nicer! Do you want to work on the new PR? I'll happily close this one.

andrewda:
@ocarreterom that looks awesome! Only thing I'd have to add is that if a specific side is not scrollable, don't show the overlay opacity.

ocarreterom:
@andrewda I have in mind :)

ocarreterom:
@machour I can try to do it! 👍

machour:
Closing this one

PR #501, opened 2017-10-18
feat(markdown): Add support for quoted emails

### LINKED ISSUE #467: markdown: Add support for quoted text
When replying from an email, the notification content is embedded in the reply.

On github.com, it's displayed as a foldable section, folded by default:
screen shot 2017-10-11 at 12 22 21 pm

Currently the app doesn't take care of it:
screen shot 2017-10-11 at 12 22 05 pm

Should be handled in the same way.

machour:
Introduced a `<ToggleView>` component which will also come handy when we add PR review comments.

gif

Closes #467

### DIFF
--- a/src/components/github-htmlview.component.js
+++ b/src/components/github-htmlview.component.js
@@ -6,7 +6,7 @@ import SyntaxHighlighter from 'react-native-syntax-highlighter';
 import { github as GithubStyle } from 'react-syntax-highlighter/dist/styles';
 import entities from 'entities';
 
-import { ImageZoom } from 'components';
+import { ImageZoom, ToggleView } from 'components';
 import { colors, fonts, normalize } from 'config';
 
 const textStyle = Platform.select({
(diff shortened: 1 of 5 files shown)


src/components/github-htmlview.component.js
```diff
@@ -65,6 +65,17 @@ const styles = {
   a: linkStyle,
 };
 
+const quotedEmailToggleStyle = {
+  backgroundColor: colors.greyMid,
+  paddingLeft: 4,
+  paddingRight: 4,
```
lex111: If use `paddingHorizontal` instead `paddingLeft` and `paddingRight`, will it work?

machour: yes, thank you for the reminder 👍

src/components/github-htmlview.component.js
```diff
@@ -246,6 +261,33 @@ export class GithubHtmlView extends Component {
...
+                    siblings,
+                    parent,
+                    defaultRenderer
+                  )}
+                </ToggleView>
+              );
+            }
+          }
+
+          return undefined;
```
lex111: I can not understand why `undefined` instead of `null`?

machour: From react-native-htmlview documentation: 
> renderNode: a custom function to render HTML nodes however you see fit. If the function returns undefined (not null), the default renderer will be used for that node.

lex111: I'm understood, thank you.

PR #502, opened 2017-10-18
refactor(auth): get user data after login

lex111:
Previously, information about the user was obtained on events screen, which is not entirely correct. This PR fixes this: the user gets after a successful login.


housseindjirdeh:
This is a good catch and agreed, it shouldn't be there :O

Just a tiny merge conflict 🕺

lex111:
@housseindjirdeh done :heavy_check_mark:

Issue #503, opened 2017-10-18
Convert to styled-components


lex111:
After discuss in the issue #480, we begin the gradual migration of the codebase to `styled-components`. This is a good and easy task for those who want to participate in Hacktoberfest or just start contributing to the project. The only rule is: **one** component or screen converted to `styled-components` in your PR.

You can see the example of migration in PR #495.

agatac:
Hi, I'm a new contributor. I can start with Badge Component.

lex111:
@agatac fine, we will wait for PR, if there are problems, please contact to chat on Gitter.

housseindjirdeh:
Agreed @lex111, I love the one component rule for now 🙌 

Welcome @agatac! The `badge` component is a great place to start :)

As per our discussion in #480, we can keep those styles in the same file for now. We can definitely look into moving them into separate files at a later point if need be. Any questions you have at any time, please don't hesitate to ask 🙏

nimadera:
I'd like to help with this. Should I choose a component or is there a recommended component to tackle next?

lex111:
@nimadera no, you can choose any component you like, just write it here so that others know about it.

pdong:
I took a shot at the view container, label-list-item, and notification icon styles.  I went through some other ones but got stuck at some spots or wasn't sure where the components were used.  I'm not sure if there is an easy way to reload certain views either as I'm pretty new to react native.

A couple of notes:
For fonts, the project is currently bringing them in as a spread operator in the StyleSheet.  I'm assuming we'd now be importing them as `${fonts.fontPrimarySemiBold.fontFamily}` (I looked at the config and fontFamily was the only property so this seems safe).

For a component like label-button-component which takes props to conditionally apply styles, I wasn't really sure what approach should be made here.  The props can be passed into the styled container (e.g. ${( { largeWithTag }: Props) => {}}) but I was thinking an approach like

```
const ButtonContainer = styled(Button).attrs({
  'padding-right': props => props.large ? 10 : 5,
  'padding-left': props => props.large ? 10 : 5,
  'padding-top': props => props.large ? 5 : 3,
  'padding-bottom': props => props.large ? 5 : 3,
  'margin-right': props => props.large ? 0 : 10,
  'min-width': props => props.large ? 'unset' : 70,
})`
  border-radius: 3;
  margin-left: 0;
  padding-right: ${props => props['padding-right']}
  padding-left: ${props => props['padding-left']}
[...]

PR #508, opened 2017-10-19
chore(*): style view-container.component.js

pdong:
Converted view-container.component.js to use styled components as part of #503


.all-contributorsrc
```diff
@@ -490,6 +490,13 @@
       "contributions": [
         "code"
       ]
+    },
+    {
+      "login": "pdong",
```

PR #475, opened 2017-10-14
feat(fonts): Remove useless fonts in iOS

chinesedfan:
Related to #402.

Picked out necessary fonts(FontAwesome, MaterialIcons and Octicons), according to 
[react-native-vector-icons docs](https://github.com/oblador/react-native-vector-icons#option-manually).

@machour I only have iOS development environment, hope you can tweak app.gradle as well.

PR #513, opened 2017-10-19
chore: fix `yarn run link`

chinesedfan:
With `rn-cli.config.js`, we can ignore assets that `react-native-vector-icons` defines in [its rnpm config](https://github.com/oblador/react-native-vector-icons/blob/master/package.json#L57). So that `react-native link` will not link those assets into our project again.

I removed some useless fonts for iOS in #475, but @housseindjirdeh and @machour pointed that `yarn run link`(which runs `react-native link` in fact) will add them back. It is really hard to find a thorough document about `react-native link`. Fortunately, I find the way to solve this problem after dig into RN source codes[[1](https://github.com/facebook/react-native/blob/master/local-cli/link/link.js#L198), [2](https://github.com/facebook/react-native/blob/master/local-cli/link/getDependencyConfig.js#L9)].

By the way, I added the font `Menlo` back. It doesn't belongs to `react-native-vector-icons`. Sorry for removed by mistake.

### DIFF
--- a/rn-cli.config.js
+++ b/rn-cli.config.js
@@ -1,0 +1,18 @@
+const defaultConfig = require('react-native/local-cli/core/default.config.js');
+
+const config = {
+  getDependencyConfig(packageName) {
+    if (packageName === 'react-native-vector-icons') {
+      // we do not need link all fonts from it, #402
+      return {
+        assets: [],
+        commands: {},
+        params: [],
+      };
+    }
+
+    return defaultConfig.getDependencyConfig(packageName);
+  },
+};
+
+module.exports = config;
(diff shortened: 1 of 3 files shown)


rn-cli.config.js
```diff
@@ -0,0 +1,18 @@
+const defaultConfig = require('react-native/local-cli/core/default.config.js');
+
```
machour: This errors out: Error: Cannot find module 'react-native/local-cli/core/default.config.js'

chinesedfan: @machour It's my fault. RN 0.48+ is hard to be installed in China due to network issues, so I usually install 0.44.3. I will update it to suit our RN version.

chinesedfan:
@machour Finally, I managed to setup my RN 0.48 environment. Now everything seems OK.

PR #517, opened 2017-10-19
fix: Remove undefined var & fix typo

machour:
`window` was used without ever being defined & setting `width` is not needed (tested on iOS & Android while changing device orientation)

This only causes an error with RN 0.49.3, really curious Oo.

### DIFF
--- a/src/components/parallax-scroll.component.js
+++ b/src/components/parallax-scroll.component.js
@@ -34,13 +34,11 @@ const styles = StyleSheet.create({
   background: {
     position: 'absolute',
     top: 0,
-    width: window.width,
     backgroundColor: colors.primaryDark,
   },
   stickySection: {
     height: STICKY_HEADER_HEIGHT,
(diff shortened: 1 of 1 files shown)


lex111:
> This only causes an error with RN 0.49.3, really curious Oo.

In version 0.49.4 the same thing? I think it needs to be somehow marked to know that it needs to be checked when the version of RN is updated.

machour:
@lex111 I meant it's weird that our current RN, or even eslint didn't catch that error.

lex111:
@machour I thought that in RN 0.49.3 this only causes an error, we do not have a new version.

andrewda:
Yea, surprises this didn't crash in earlier version of RN. ESLint probably didn't catch because this is valid web JavaScript.

andrewda:
Unless maybe RN artificially injected a `window` variable before?

housseindjirdeh:
Interesting, was thinking the same thing @andrewda :thinking:

PR #524, opened 2017-10-20
feat(events): add scroll to top button

lex111:
Demo:

image

### DIFF
--- a/src/auth/screens/events.screen.js
+++ b/src/auth/screens/events.screen.js
@@ -6,7 +6,12 @@ import { bindActionCreators } from 'redux';
 import { StyleSheet, Text, FlatList, View } from 'react-native';
 import moment from 'moment/min/moment-with-locales.min';
 
-import { LoadingUserListItem, UserListItem, ViewContainer } from 'components';
+import {
+  LoadingUserListItem,
+  UserListItem,
+  ViewContainer,
+  BackToTop,
+} from 'components';
 import { colors, fonts, normalize } from 'config';
 import { emojifyText, translate } from 'utils';
 import { getUserEvents } from 'auth';
@@ -77,6 +82,14 @@ const styles = StyleSheet.create({
 });
 
 class Events extends Component {
+  constructor() {
+    super();
+
+    this.state = {
+      offsetY: null,
+    };
+  }
+
   componentDidMount() {
     this.getUserEvents();
   }
@@ -482,6 +495,14 @@ class Events extends Component {
     return item.id;
   };
 
+  scrollTop = () => {
+    this.eventsList.scrollToOffset({
+      x: 0,
+      y: 0,
+      animated: true,
+    });
+  };
+
   renderDescription(userEvent) {
     return (
       <Text style={styles.descriptionContainer}>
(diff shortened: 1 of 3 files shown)


src/auth/screens/events.screen.js
```diff
@@ -77,6 +82,14 @@ const styles = StyleSheet.create({
 });
 
 class Events extends Component {
+  constructor() {
+    super();
```
chinesedfan: `super(props)` may be better.

src/components/back-to-top.component.js
```diff
@@ -0,0 +1,72 @@
...
+  render() {
+    const { onPress, bottom } = this.props;
+    const extraStyles = {
+      bottom,
+      opacity: this.state.animatedOpacity,
+    };
+
+    return (
+      <Container style={extraStyles}>
+        <TouchableOpacity onPress={() => onPress()}>
```
chinesedfan: Can here be written as `onPress={onPress}` directly?

src/auth/screens/events.screen.js
```diff
@@ -77,6 +82,14 @@ const styles = StyleSheet.create({
...
+
+    this.state = {
+      offsetY: null,
```
chinesedfan: If `offset` is number, should it be zero by default?

src/components/back-to-top.component.js
```diff
@@ -0,0 +1,72 @@
...
+          <Icon color={colors.white} size={32} name="expand-less" />
+        </TouchableOpacity>
+      </Container>
+    );
+  }
+}
+
+BackToTop.defaultProps = {
+  offsetY: null,
```
chinesedfan: Similar here.

src/components/back-to-top.component.js
```diff
@@ -0,0 +1,72 @@
...
+  opacity: 1,
+  bottom: 10,
+  distance: 200,
+  onPress: undefined,
```
chinesedfan: The default value is `undefined`, is it not meaningful?

ocarreterom:
This functionality would not be necessary in iOS devices, because touching the status bar, the view scrolls to top. 🤔 
What do you thing?

andrewda:
Good point @ocarreterom. I still kinda like the button, but I guess either way works for iOS.

machour:
So, do we keep this for iOS, or do we restrict it for Android only? 🤔

andrewda:
@machour I still like it as is – keep the button on iOS.

ocarreterom:
@andrewda in my opinion if you want quick action to scroll to top, touching the current active tab is the best improvement. Android and iOS have this feature in common. I don't use any app that use a floating button to scroll.

housseindjirdeh:
Since we haven't come to a consensus here, going to close this for now (thank you thank you @lex111 for putting the work to make this work).

Let's keep the discussion flowing if somebody happens to think this would be nice to have and worthwhile. Thoughts @machour @chinesedfan?

chinesedfan:
I came up with a compromised solution that only setting the button visible when users scroll to top.

machour:
To be honest, I've never been a fan of this kind of button on websites, but I'm fine living with it if it's implemented (maybe a togglable configuration?)

PR #525, opened 2017-10-20
feat(markdown): Support to render del tag

chinesedfan:
For example, when you visit #479.

### DIFF
--- a/src/components/github-htmlview.component.js
+++ b/src/components/github-htmlview.component.js
@@ -235,11 +235,6 @@ export class GithubHtmlView extends Component {
             </Text>
           </View>
         ),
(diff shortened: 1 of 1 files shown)


src/components/github-htmlview.component.js
```diff
@@ -233,6 +235,11 @@ export class GithubHtmlView extends Component {
             </Text>
           </View>
         ),
+        del: (node, index, siblings, parent, defaultRenderer) => (
+          <Text key={index} style={styles.del}>
+            {defaultRenderer(node.children, node)}
+          </Text>
+        ),
```
machour: This is not needed since it's the default behavior of `<HTMLView />` (it will read the `del` style from its `stylesheet` prop).

PR #527, opened 2017-10-20
fix: Fix button transparency on Android

machour:
Seems like this is now needed by `react-native-elements` button to correctly render a transparent button absolutely positioned.

(we bumped from "^0.9.3" to "^0.17.0" in #407)

### DIFF
--- a/src/auth/screens/login.screen.js
+++ b/src/auth/screens/login.screen.js
@@ -55,7 +55,10 @@ const styles = StyleSheet.create({
     borderColor: colors.white,
     paddingVertical: 10,
     paddingHorizontal: 30,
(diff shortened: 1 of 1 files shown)


jouderianjr:
Hey @machour Looking  this pr (https://github.com/react-native-training/react-native-elements/pull/382 ) looks like they added a outline props that should do the same thing, Could you test it? Maybe fits in our user case 😄 

Ps. This is not documented yet on react-native-elements site :(

machour:
@jouderianjr already tested it (and transparent too), same white background on Android:

screen shot 2017-10-20 at 3 40 15 pm

jouderianjr:
oh, got it 😄

PR #529, opened 2017-10-20
feat(members): add gradient to members list

ocarreterom:
Adds a gradient to the right and left side in the members scroll list.

Also removes the `marginRight` from the last avatar in the list.

captura de pantalla 2017-10-20 a las 18 16 55


src/components/members.component.js
```diff
@@ -60,8 +60,13 @@ const styles = StyleSheet.create({
     paddingLeft: 15,
   },
   flatList: {
-    marginLeft: 15,
-    marginRight: 15,
+    paddingLeft: 15,
+    paddingRight: 15,
```
lex111: This can be simplified: instead of `paddingLeft` and `paddingRight`, use `paddingHorizontal`.

andrewda:
This looks outstanding! Thanks so much @ocarreterom!!

.all-contributorsrc
```diff
@@ -497,6 +497,15 @@
       "avatar_url": "https://avatars2.githubusercontent.com/u/5158502?v=4",
       "profile": "http://www.phidong.com",
       "contributions": []
+    },
+    {
+      "login": "ocarreterom",
+      "name": "Óscar Carretero",
```

PR #533, opened 2017-10-21
refactor(diff-blocks): convert diff-blocks to styled component

machour:
My first converted component!

I've found the need to define a new `config.fontsStyled` object with converted styles rules.
Is that ok?

Also renamed styles to avoid colors in names.


src/config/fonts.js
```diff
@@ -6,3 +6,12 @@ export const fonts = {
   fontPrimaryBold: { fontFamily: 'Nunito-Bold' },
   fontCode: { fontFamily: 'Menlo' },
 };
+
+export const fontsStyled = {
```
alejandronanez: I think @josenaranjo is working on something related to fonts as he mentioned [here](https://github.com/gitpoint/git-point/issues/503#issuecomment-337936548) AFAIK he created a new file called `styledFonts.js` inside `/config`, maybe you can talk to him before merging this PR so you don't get any merge conflict in the future. FWIW I like the idea of having a new fresh file with fonts.

machour: `styledFonts.js` looks good to me. I'll wait on @josenaranjo PR to get merged before changing this one

alejandronanez: @machour ok, @josenaranjo's PR was merged https://github.com/gitpoint/git-point/pull/561 You can start using his font file.

andrewda: 🤔  I actually like having all the fonts in one file. Not sure though.

machour: seconding @andrewda, keeping both variables in the same file will help making sure we change both when updating something.

@alejandronanez  Is it okay of I merge it back in `fonts.js` ?

alejandronanez: Sounds good to me, do you mind removing the file @josenaranjo added on his PR and doing a small refactor on his component so we don't introduce more tech debt in here?

machour:
Replaced by #692

Issue #532, opened 2017-10-20
Migration to styled-components


alejandronanez:
Here's a list based off @andrewda's list for adding unit tests to the app (https://github.com/gitpoint/git-point/issues/518), feel free to add/remove components if you think it's necessary.

**For newcomers:**
1. Choose the component you want to style
1. Comment here which component you'd like to work on
1. We'll update this list with your name
1. Make sure to include "before" and "after" screenshots in your submitted PR

# Components 
- [ ] Button (unclaimed)
- [ ] IssueListItem (@josenaranjo - #578)
- [ ] LabelButton (@chinesedfan - #727)
- [ ] MarkdownWebView (@josenaranjo - PR Pending)
- [ ] RepositoryProfile (@shmesa22 - PR Pending)
- [ ] RepositorySectionTitle (@bg-k – PR Pending)
- [ ] SectionList (@jamesg1 - #608)
- [ ] StateBadge (@rkpasia - #715)
- [ ] ToggleView (unclaimed)
- [ ] UserProfile (@chinesedfan – #724)

Done:
- [x] Badge (@jpls93 – #535)
- [x] CodeLine (@josenaranjo – #561)
- [x] CommentInput (@shmesa22 – #596)
- [x] CommentListItem (@chinesedfan - #723)
- [x] DiffBlocks (@machour – #692)
- [x] EntityInfo (@MrLoh - #558)
- [x] ImageZoom (@MrLoh - #559)
- [x] InlineLabel (@binkpitch – #544)
- [x] IssueDescription (@josenaranjo - #563)
- [x] IssueEventListItem (@josenaranjo - #564)
- [x] LabelListItem (@pdong – #509)
- [x] Members (@shmesa22 - #606)
- [x] MentionArea (@binkpitch – #553)
- [x] NotificationIcon (@pdong – #510)
- [x] NotificationListItem (@LeoCp - #610)
- [x] ParallaxScroll (@shmesa22 - #607)
- [x] RepositoryListItem (@apoeco - #605)
- [x] ViewContainer (@pdong – #508)
- [x] UserListItem (@shmesa22 - #601)

# Screens
- [ ] Repository (unclaimed)
- [ ] Search (@Arjun-sna - #710)
- [x] Auth (@ZahraTee - #702)
- [ ] Issue (@remarks - PR Pending)

Done:
- [x] Notifications (@ZahraTee - #684 )
- [x] Organization (@ZahraTee - #682)
- [x] User (@ZahraTee - #683)

[...]

PR #535, opened 2017-10-21
refactor(badge): refactor badge component to use styled-components

jpls93:
Related to #532


src/components/badge.component.js
```diff
@@ -1,5 +1,5 @@
 import React from 'react';
-import { StyleSheet, View, Text } from 'react-native';
+import styled from 'styled-components';
```
chinesedfan: Please import from 'styled-components/react-native'.

jpls93:
@chinesedfan 🎉

src/components/badge.component.js
```diff
@@ -10,39 +10,29 @@ type Props = {
...
+const BadgeContainer = styled.View`
+  flex: 1;
+  align-items: 'center';
+  justify-content: 'center';
+  border-radius: 18;
+  width: 18;
+  height: 18;
+  border-color: ${colors.alabaster};
+  border-width: 1;
+  ${({ backgroundColor }) => `background-color: ${backgroundColor};`};
```
alejandronanez: I think you can do `background-color: ${backgroundColor}` instead,

src/components/badge.component.js
```diff
@@ -10,39 +10,29 @@ type Props = {
...
+`;
+
+const BadgeText = styled.Text`
+  ${{ ...fonts.fontPrimaryBold }};
+  background-color: 'transparent';
+  ${({ largeText }) =>
```
alejandronanez: Same goes for this `font-size: ${largeText ? normalize(9.5) : normalize(7)}`

jpls93: Hi @alejandronanez , might not work since unlike `border-color: ${colors.alabaster};`, `colors` is a file-wide variable whereas `backgroundColor` and `largeText` are props passed down to the styled components. https://www.styled-components.com/docs/basics#adapting-based-on-props

jpls93: @alejandronanez , let me know if I can help you with anything else

alejandronanez: Oh yeah, you're right. I missed that, sorry.
Thanks!

MrLoh: the usual way would be this though: 
```
...
    font-size: ${({largeText}) => largeText ? normalize(9.5) : normalize(7)};
```

jpls93: I agree that would be more concise. I guess I just made the code style more consistent by making it parallel to how I wrote BadgeContainer's `backgroundColor` which doesn't have a default value. Since writing something without a default value that way would make it `background-color: undefined` which, without tree-shaking, would yield dead code.

Issue #537, opened 2017-10-21
state-badge.component - Migration to styled-components


ruskakimov:
Claiming the state-badge component.
Issue #532

chinesedfan:
@ruskakimov I am afraid that #535 has done the same thing. Can you pick another component?

ruskakimov:
@chinesedfan sure thing

PR #543, opened 2017-10-21
refactor(state-badge): switch to styled components

ruskakimov:
state-badge component refactored with styled components. #537 

~Please, check it in your local environment first. I have made the changes blindfolded without testing them on a phone, as I have failed to setup the react-native environment. :cry:~

Managed to set up react native and debug it on my android device. :tada:

### DIFF
--- a/src/components/state-badge.component.js
+++ b/src/components/state-badge.component.js
@@ -13,25 +13,26 @@ type Props = {
   locale: string,
 };
 
+const colorOfIssueState = issueState => {
+  switch (issueState) {
+    case 'merged':
+      return colors.purple;
+    case 'open':
+      return colors.green;
+    case 'closed':
+      return colors.red;
+    default:
+      return colors.black;
+  }
+};
+
 const Badge = styled.View`
-  padding: 12;
+  padding-left: 12;
+  padding-right: 12;
   padding-top: 3;
   padding-bottom: 3;
   border-radius: 20;
-  background-color: ${({ issueState }) => {
-    switch (issueState) {
-      case 'merged':
-        return colors.purple;
-      case 'open':
-        return colors.green;
-      case 'closed':
-        return colors.red;
-      default:
-        break;
-    }
-
-    return colors.black;
-  }};
+  background-color: ${({ issueState }) => colorOfIssueState(issueState)};
 `;
 
 const BadgeText = styled.Text`
--- a/src/components/state-badge.component.js
+++ b/src/components/state-badge.component.js
@@ -1,5 +1,5 @@
 import React from 'react';
-import { Text, View, StyleSheet } from 'react-native';
+import styled from 'styled-components/native';
 
 import { colors, fonts, normalize } from 'config';
 import { translate } from 'utils';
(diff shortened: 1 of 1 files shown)


src/components/state-badge.component.js
```diff
@@ -13,28 +13,32 @@ type Props = {
...
-    ...fonts.fontPrimarySemiBold,
-    color: colors.white,
-  },
-});
+const Badge = styled.View`
+  padding: 12;
+  padding-top: 3;
+  padding-bottom: 3;
+  border-radius: 20;
+  background-color: ${({ issueState }) => {
```
alejandronanez: I'd like to see this logic moved into another function - that will allow us to test it

src/components/state-badge.component.js
```diff
@@ -13,28 +13,32 @@ type Props = {
...
+    switch (issueState) {
+      case 'merged':
+        return colors.purple;
+      case 'open':
+        return colors.green;
+      case 'closed':
+        return colors.red;
+      default:
+        break;
```
alejandronanez: You can return `colors.black` here instead of returning it on line 33.

src/components/state-badge.component.js
```diff
@@ -58,19 +62,9 @@ export const StateBadge = ({
...
-    issueStyle = styles.openIssue;
-  } else if (issueState === 'closed') {
-    issueStyle = styles.closedIssue;
-  }
-
   return (
-    <View style={[styles.badge, style, issueStyle]}>
-      <Text style={styles.text}>{issueText}</Text>
-    </View>
+    <Badge issueState={issueState} style={style}>
```
chinesedfan: I think `style` should be integrated into `Badge`, or we may lose the meaning of styled component.

src/components/state-badge.component.js
```diff
@@ -13,28 +13,32 @@ type Props = {
...
-    backgroundColor: colors.red,
-  },
-  text: {
-    fontSize: normalize(12),
-    ...fonts.fontPrimarySemiBold,
-    color: colors.white,
-  },
-});
+const Badge = styled.View`
+  padding: 12;
```
chinesedfan: Be careful with shorthands properties. You can check [here](https://www.styled-components.com/docs/basics#react-native). 😆

src/components/state-badge.component.js
ruskakimov: completed!

ruskakimov:
Btw, current `StateBadge` tests are no longer compatible with the implementation. I have searched around and found [this](https://github.com/styled-components/jest-styled-components) library. What do you think about it?

alejandronanez:
If I’m not wrong, that library is built for snapshot testing + styled components. And here we’re not using snapshots. What are the issues with the current tests?

src/components/state-badge.component.js
ruskakimov: From what I found `style` prop is passed by `repository-section-title` and `issue-description` components.


`Repository-section-title` passed style object:
```
{
  marginLeft: 10,
}
```

`Issue-description` passed style object:
```
{
    alignItems: 'flex-end',
    justifyContent: 'center',
}
```

ruskakimov:
The following line in tests breaks down:
`expect(wrapper.prop('style')[2].backgroundColor).toEqual(colors.green);`

With this error:
`TypeError: Cannot read property '2' of undefined`


Full test:
```
  it('correctly renders with open issue', () => {
    const wrapper = shallow(<StateBadge issue={openIssue} language="en" />);

    expect(
      wrapper
        .childAt(0)
        .childAt(0)
        .text()
    ).toEqual('Open');
    expect(wrapper.prop('style')[2].backgroundColor).toEqual(colors.green);
  });
```

src/components/state-badge.component.js
ruskakimov: I could not find any examples of passing new style properties to styled component directly. Probably not a good idea? So, what about making `bool` props that would control if the mentioned styles apply?

ruskakimov: Oh, I think I have found a [solution](https://www.styled-components.com/docs/basics#extending-styles). We could just extend `StateBadge` with the mentioned styles in `issue-description` and `repository-section-title` locally.

chinesedfan:
@ruskakimov I am not sure whether current tests is compatible or not, but I think you forget to handle `style` props in the styled component. Maybe something like the following.

```
const Badge = styled.View`
    // blabla, please fix the grammar
    ${({style}) =>...}
`;
```

src/components/state-badge.component.js
```diff
@@ -13,28 +13,30 @@ type Props = {
...
+
+const Badge = styled.View`
+  padding: 3px 12px;
+  border-radius: 20px;
+  background-color: ${({ issueState }) => colorOfIssueState(issueState)};
+`;
+
+const BadgeText = styled.Text`
+  font-size: ${normalize(12)};
+  font-family: ${fonts.fontPrimaryBold.fontFamily};
```
machour: Instead of:
`font-family: ${styledFonts.fontPrimaryBold};`

use:
`${fonts.fontPrimaryBold}`

ruskakimov:
I am sorry guys, could you please reassign this task to somebody else? Sorry for wasting your time :sweat_smile:

machour:
Oh man, sorry I missed your feedback yesterday. There are no apologies needed and time wasn't wasted at all! Please feel free to come back & help us with our ongoing tasks whenever you feel like it! ❤️

PR #546, opened 2017-10-21
Fix: Updated yarn.lock hash for react-native-mock

jglover:
Updates hash for react-native-mock to match expected `96bee7b072cea66f5dfa21c39b2416ca6fb6c0da`.  
This also causes some cascading changes in the lock file after running `yarn` (install) again so should be treated with caution.   
Tested on Android (physical) and iOS simulator.

### DIFF
--- a/yarn.lock
+++ b/yarn.lock
@@ -151,7 +151,7 @@ anchor-markdown-header@^0.5.5:
   dependencies:
     emoji-regex "~6.1.0"
 
-ansi-escapes@^1.0.0, ansi-escapes@^1.1.0, ansi-escapes@^1.4.0:
+ansi-escapes@^1.0.0, ansi-escapes@^1.1.0:
   version "1.4.0"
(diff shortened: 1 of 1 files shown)


alejandronanez:
What command did you run to update the `yarn.lock` file? Modifying directly `yarn.lock` is not good.

jouderianjr:
hey @alejandronanez I think he just deleted the `yarn.lock` and ran `yarn install` again, I did that and worked in my env 😄

jglover:
Just to give some background, I updated the hash via `yarn remove` and `yarn add`, I believe the hash may have been wrong originally because someone updated the file manually, probably resolving a conflict gone bad?
The other changes were caused by running `yarn` afterwards, which I believe has caused it to add dependencies for `react-native-mock` whilst also making sure it has all the versions required by peer dependencies.

andrewda:
Guessing the issue was b/c I used the fork instead of the original and didn't do it through yarn. Thanks for the fix!

PR #557, opened 2017-10-22
Refactor comment-list-item to use styled-components

MrLoh:
related to #532


MrLoh:
the `CONTRIBUORS.md` file is somehow broken, so I couldn't add myself. I think the last contributor might have tried to add himself manually.

andrewda:
@MrLoh I'll try to fix that.

andrewda:
@MrLoh try adding yourself now

MrLoh:
cool, thant worked. I rebased of master and added myself. Should be ready to merge

.all-contributorsrc
```diff
@@ -554,6 +554,13 @@
       "contributions": [
         "code"
       ]
+    },
+    {
+      "login": "MrLoh",
+      "name": "Tobias Lohse",
+      "avatar_url": "https://avatars0.githubusercontent.com/u/1285032?v=4",
+      "profile": "http://MrLoh.se",
+      "contributions": []
```
chinesedfan: 🎉 You contributed codes in fact.

src/components/comment-list-item.component.js
```diff
@@ -13,71 +13,71 @@ import moment from 'moment/min/moment-with-locales.min';
...
-  },
-});
+const Container = styled.View`
+  padding: 10px 10px 0 0;
+  background-color: transparent;
+`;
+
+const Header = styled.View`
+  flex-direction: row;
+  margin-left: 10px;
```
chinesedfan: Nice catch for those shorthands properties. But shall we always add units for properties? (cc @alejandronanez)

src/components/comment-list-item.component.js
```diff
@@ -13,71 +13,71 @@ import moment from 'moment/min/moment-with-locales.min';
...
+`;
+
+const DateContainer = styled.View`
+  flex: 0.1;
+  align-items: flex-end;
+  justify-content: center;
+`;
+
+const LinkDescription = styled.Text`
+  ${{ ...fonts.fontPrimaryBold }};
```
chinesedfan: Can we just write as `${fonts.fontPrimaryBold}`?

src/components/comment-list-item.component.js
```diff
@@ -13,71 +13,71 @@ import moment from 'moment/min/moment-with-locales.min';
...
+const CommentContainer = styled.View`
+  padding-top: 5px;
+  margin-left: 54px;
+  border-bottom-color: ${colors.greyLight};
+  border-bottom-width: 1px;
+  padding-bottom: ${props => (props.bottomPadding ? 15 : 0)}px;
+`;
+
+const CommentTextNone = styled.Text`
+  ${{ ...fonts.fontPrimary }};
```
chinesedfan: Same here.

src/components/comment-list-item.component.js
MrLoh: I don’t know. It’s up to you. I never use units for 0 in css.

src/components/comment-list-item.component.js
MrLoh: I don’t kno. I’m not sure how this actually works. Just copied it from the other already concerted components. I handle fonts differently in my apps.

src/components/comment-list-item.component.js
alejandronanez: I think we should stick to unitless, that's what we've been doing in other PRs

MrLoh:
@alejandronanez I don't think it's a good idea to go unitless, as it is against the [styled-component docs](https://www.styled-components.com/docs/basics#react-native) and will break percentage units. 

I'm also a bit confused as to who is in charge of reviewing and merging pull requests as two other of my pull-requests with the same style issues where accepted and merged allready. #559 #558

chinesedfan:
@MrLoh Yes, styled-component requires units in sometimes, but it is only for shorthands properties. And sorry about previous PR reviews are inconsistent.

src/components/comment-list-item.component.js
machour: This should be:

`font-family: ${styledFonts.fontPrimaryBold}`

PR #563, opened 2017-10-23
refactor(styles): IssueDescription component styles update

josenaranjo:
Migrating StyleSheet styles to styled components

Related #532

### DIFF
--- a/src/components/issue-description.component.js
+++ b/src/components/issue-description.component.js
@@ -1,8 +1,9 @@
 import React, { Component } from 'react';
-import { StyleSheet, View, ActivityIndicator } from 'react-native';
+import { ActivityIndicator } from 'react-native';
 import { ListItem } from 'react-native-elements';
 import Parse from 'parse-diff';
 import moment from 'moment/min/moment-with-locales.min';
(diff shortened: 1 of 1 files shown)


src/components/issue-description.component.js
```diff
@@ -12,62 +13,67 @@ import {
...
+
+const DiffBlocksContainer = styled.View`
+  flex-direction: row;
+  align-items: center;
+  justify-content: flex-end;
+  padding-right: 10;
+  padding-bottom: 10;
+`;
+
+const StateBadgeStyled = styled(StateBadge)`
```
andrewda: 🤔  This is failing the test for some reason.

```
Cannot create styled-component for component: undefined
```

andrewda: Actually, why are we styling the StateBadge at all? Seems a bit weird. Shouldn't this all be done in the component?

josenaranjo: Couldn't agree more @andrewda . I reverted these styles to use StyleSheet while we refactor StateBadge component, so it doesn't require any style props.

Issue #518, opened 2017-10-20
Add more tests!


andrewda:
We can use this issue to divide up some work. I'll add a checklist below with what needs to be done and who it's assigned to. Please comment (or edit yourself in) if you'd like to take on a test!

### Components

- [x] Badge (@andrewda – #407 )
- [x] Button (@andrewda – #407)
- [x] CodeLine (@andrewda – #407)
- [ ] CommentInput (unclaimed)
- [ ] CommentListItem (unclaimed)
- [ ] DiffBlocks (unclaimed)
- [x] EntityInfo (@andrewda – #407)
- [ ] GithubHtmlView (unclaimed)
- [ ] GithubStyle (unclaimed)
- [ ] ImageZoom (unclaimed)
- [ ] InlineLabel (unclaimed)
- [ ] IssueDescription (unclaimed)
- [ ] IssueEventListItem (unclaimed)
- [ ] IssueListItem (unclaimed)
- [x] LabelButton (@andrewda – #407)
- [ ] LabelListItem (unclaimed)
- [ ] MarkdownWebView (unclaimed)
- [ ] Members (unclaimed)
- [ ] MentionArea (unclaimed)
- [ ] NotificationIcon (unclaimed)
- [ ] NotificationListItem (unclaimed)
- [ ] ParallaxScroll (unclaimed)
- [ ] RepositoryListItem (unclaimed)
- [ ] RepositoryProfile (unclaimed)
- [ ] RepositorySectionTitle (unclaimed)
- [ ] SearchBar – Android (unclaimed)
- [ ] SearchBar – iOS (unclaimed)
- [ ] SectionList (unclaimed)
- [x] StateBadge (@andrewda – #407)
- [x] ToggleView (@machour – #407)
- [ ] UserListItem (unclaimed)
- [ ] UserProfile
- [ ] ViewContainer

jouderianjr:
I'll start slow :) Could you give me ImageZoom and CommentListItem? 💃

alejandronanez:
I can grab userProfile and RepositoryProfile.

machour:
I'd like to implement tests for `<GithubHtmlView />` but I'm not sure how to implement them with Enzyme. (I'm fairly new to testing :relaxed:)

This component does two things:
 1. Transform a given HTML to a new HTML with some modifications
 2. Hands out this new HTML to `<HTMLView />`

Should I stick to only testing the first part, or should the output result of <HTMLView /> be tested too?
[...]

PR #565, opened 2017-10-23
test: Add test for Color Helpers (#518)

kenitive:
Hello,
I've added tests for the color helper :)


__tests__/tests/utilities/color-helpers.js
```diff
@@ -0,0 +1,33 @@
+import { getFontColorByBackground } from 'utils';
+
+describe('Color Helpers', () => {
+  it('should return a black font color for the light background #1', () => {
```
andrewda: Could you say something like `'should return a black font color for the light yellow background'` and apply that format to the rest of these tests?

kenitive:
@andrewda done :)

Issue #569, opened 2017-10-24
v1.4.0 Release - Final Checks


housseindjirdeh:
Releasing `1.4.0` has been long overdue, and my apologies :O. Let's have this out before the end of this week ⭐️ 

Couple of things I've noticed so let's use this issue to track down all the changes we'll need before release 🚀 

**For any thing mentioned on the list here and on this issue you want to tackle, just tag this issue number on your PR.**

On iOS && Android:
----

### Profile images don't seem to be showing on profile screens

![image](https://user-images.githubusercontent.com/12476932/31921322-da7eaf0a-b83c-11e7-8878-17ae8188be2f.png)

### A few list items throughout the app have borders bolded a lot more

![image](https://user-images.githubusercontent.com/12476932/31921499-ac0b161c-b83d-11e7-96f4-cba1b7dd6494.png)

### Notifications badge count has styling issues

View both screenshots above

Anything else anybody sees, please add it to this ticket. Shouldn't be too much regressions (and now that we're adding tests we should definitely be seeing less)

alejandronanez:
@housseindjirdeh I can take a look at the profile picture bug.

housseindjirdeh:
Thank you matey @alejandronanez <3

housseindjirdeh:
⚠️ Thanks for logging this one @machour:

## Android

![image](https://user-images.githubusercontent.com/12476932/31922003-875a78be-b840-11e7-8573-2b9017c9f550.png)

App intro buttons not showing smoothly

machour:
@housseindjirdeh The app intro button should have been fixed by #527 
Are you still getting this?

machour:
@jouderianjr did you fix the cropped buttons for issues and pulls? I don't remember anymore :D

jouderianjr:
yeah, #545 😄  Sorry for don't help much in this release, I'm pretty busy in this days 😢

housseindjirdeh:
Thanks @machour, didn't run it and thought it was still showing for Android users. Thank you <3

housseindjirdeh:
@yannbf was kind enough to log this in #536. Not sure if it's still occurring but logging it here to remind us to double check if it hasn't been resolved yet :)

![image](https://user-images.githubusercontent.com/12476932/31967362-af1a608a-b8db-11e7-83df-3dbf6f4e1429.png)

alejandronanez:
[...]

PR #572, opened 2017-10-24
fix(build): fix styled components errors

alejandronanez:
This was breaking `master` for 1.4
Related to #569


machour:
It's not the first time a styled-component merged PR breaks the app. We should be more careful with manual tests *before* merging.

andrewda:
Getting [`detox`](https://github.com/wix/detox) set up would be awesome.

PR #578, opened 2017-10-25
refactor(styles): StyleSheet to Styled Components

josenaranjo:
I left some StyleSheet styles without migrating because those styles are so specific about a particular behavior that 1. I'm not sure they are going to display the component as expected and 2. It has more meaning as it is. Let me know what you think.

#532

### DIFF
--- a/src/components/issue-list-item.component.js
+++ b/src/components/issue-list-item.component.js
@@ -1,7 +1,8 @@
 import React from 'react';
-import { StyleSheet, TouchableHighlight, View, Text } from 'react-native';
+import { StyleSheet, TouchableHighlight } from 'react-native';
 import { ListItem, Icon } from 'react-native-elements';
 import moment from 'moment/min/moment-with-locales.min';
+import styled from 'styled-components/native';
 
 import { colors, fonts, normalize } from 'config';
 import { translate } from 'utils';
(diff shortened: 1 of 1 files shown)


src/components/issue-list-item.component.js
```diff
@@ -13,16 +14,28 @@ type Props = {
   locale: string,
 };
 
+const Container = styled.View`
+  flex-direction: row;
+  align-items: flex-start;
+  justify-content: center;
+  padding-right: 10;
+  padding: 5;
```
chinesedfan: Shorthands properties should have units.

src/components/issue-list-item.component.js
```diff
@@ -13,16 +14,28 @@ type Props = {
...
-  container: {
-    flexDirection: 'row',
-    alignItems: 'flex-start',
-    justifyContent: 'center',
-    paddingRight: 10,
-    paddingVertical: 5,
-    borderBottomWidth: 1,
-    borderBottomColor: colors.greyLight,
-  },
   closedIssue: {
```
chinesedfan: I think it can be removed, because nowhere uses it.

src/components/issue-list-item.component.js
josenaranjo: Fixed

src/components/issue-list-item.component.js
josenaranjo: The closeIssue style is used in Line 70.

chinesedfan: Oh yes, I forget to expand those collapsed codes.

src/components/issue-list-item.component.js
```diff
@@ -99,10 +103,10 @@ export const IssueListItem = ({ type, issue, navigation, locale }: Props) => (
         titleStyle={styles.title}
         subtitleStyle={styles.subtitle}
       />
```
machour: Once #619 is merged, this `<ListItem>` can be converted as a styled component using this:

```js
const Issue = styled(ListItem).attrs({
  containerStyle: {
     // old style attributes
  },
  titleStyle: {
     // old style attributes
  },
  subtitleStyle: {
     // old style attributes
  }
})``;
```

src/components/issue-list-item.component.js
```diff
@@ -99,10 +103,10 @@ export const IssueListItem = ({ type, issue, navigation, locale }: Props) => (
...
-      <View style={styles.commentsContainer}>
+      <CommentsContainer>
         <Icon name="comment" type="octicon" size={18} color={colors.grey} />
```
machour: I'm wondering if we should define a new styled component here (same way as ListItem), and have all this static props in the styled component.
The JSX code would then simply be something like: `<CommentIcon />`

What do you think of this? 🤔

src/components/issue-list-item.component.js
machour: #619 is merged, you should now be able to convert the react-native-elements components!

PR #579, opened 2017-10-25
test: Added tests for auth reducer and refactored some other tests

jglover:
Adds test for auth reducer and refactoring.
Also adds default fallback for 'en' inside `localization-helper` incase location cannot be determined/promise to get locale in `app.js` has not fulfilled yet and somehow user has navigated, requiring a value (this was breaking unit tests).


andrewda:
Also - don't end data files in `.data.js`. If we're going to change the extensions for the files, we should do it in a new PR. For now keep things consistent.

andrewda:
There's a _lot_ of data in `data/api/users.data.js` file. Maybe we could shorten it down to 1 event per type so it's easier to follow if something breaks?

jglover:
I'll rename those files, yeah it's quite a bit of data; I just took the GitHub example responses verbatim from the API docs. Since the file isn't used in a build is it a good idea to just keep data as a copy of the example responses incase it ever needs to be changed?

jglover:
@andrewda done. What's the reasoning behind not using `*.spec.js` though? I'm curious about that one, it's the accepted convention in just about every framework or project I've worked with and helps differentiate the file when using fuzzy search, triggers test syntax highlighting in most editors etc

andrewda:
@jglover Several reasons: 1) when I began adding tests, I didn't use it, and we've stayed consistent with that, 2) everything's already in a single `__tests__/` directory, so it's not really necessary to define something as a specification. That being said, I'd love to completely redo our test naming scheme in another PR (an issue for discussion would be great, too).

src/utils/localization-helper.js
```diff
@@ -8,7 +8,8 @@ export const translate = (key, locale, interpolation = null) =>
   I18n.t(key, { locale, ...interpolation });
 
 export const getLocale = () => {
-  const locale = I18n.locale.toLowerCase();
+  // If for some reason a locale cannot be determined, fall back to 'en'.
+  const locale = (I18n.locale && I18n.locale.toLowerCase()) || 'en';
```
alejandronanez: 💯

jglover:
@andrewda Anything left outstanding? Keen to get this merged before it gets conflicts.

src/utils/localization-helper.js
chinesedfan: @jglover Please read `en` from `src/config/common`, as well as the comment.

jglover: @chinesedfan good tip, updated.

PR #558, opened 2017-10-22
refactor: use styled-components in entity-info

MrLoh:
related to #532

PR #586, opened 2017-10-26
fix(ui): Fix entity-info styling

machour:
Fixes unwanted UI changes introduced by #558 

| Before | After |
| -------|-------|
| screen shot 2017-10-26 at 4 22 12 pm  | screen shot 2017-10-26 at 4 21 18 pm |

⚠️ ⚠️ ⚠️ ⚠️ ⚠️ ⚠️ ⚠️ ⚠️ 

It seems that using styled-components the way it's done here does more harm than good. We didn't need to specify `font-weight 'normal'`, or `wrapperStyle` in the past. 
It's like using styled-component completely reset the style of the react-native-elements component, and we have to specify defaults again. Really weird.

Another painful thing is that before using SC, the `hideChevron` prop produced consistent margins no matter of its value. With SC, I got this difference for the first element:

screen shot 2017-10-26 at 4 11 18 pm

I had to display the chevron, but transparent for that element, to produce a consistent margin.

The big question here: Is SC used as it should be here? Is this a bug of SC ?


andrewda:
A bit random but @housseindjirdeh could you configure Coveralls to pass the Status Check as long as there's less than a 0.25% reduction in coverage? It's a bit silly to have it fail for this PR, or any PR that adds a tiny bit of complexity.

andrewda:
Actually not sure how I feel about that... Maybe we should keep it as is so we can be aware of when we're lowering coverage. Input?

machour:
+1 on keep in it like that for now @andrewda

andrewda:
Alright, this should be fine to be merged. I'm adding a lot more tests for EntityInfo, but will follow those up in a future PR.

machour:
Finally had the time to investigate a bit more, there's an incompatibility between Styled-Components & React Native Elements where the default container style get lost.

More info in this bug report: https://github.com/react-native-training/react-native-elements/issues/679

machour:
Closing this as #619 fixes the root cause

PR #589, opened 2017-10-27
test: add NotificationListItem component tests

alejandronanez:
100% coverage for `NotificationListItem` component
screen shot 2017-10-28 at 11 29 20 am


__tests__/tests/components/NotificationListItem.js
```diff
@@ -0,0 +1,95 @@
...
+    const wrapper = shallow(
+      <NotificationListItem {...defaultProps} notification={notification} />
+    );
+
+    const result = wrapper.instance().getComponentType();
+
+    expect(result).toBe(TouchableOpacity);
+  });
+
+  it('should call navigation when user press TitleComponent', () => {
```
andrewda: I would say `'should call navigation when TitleComponent is pressed`'. Talking about the user is more appropriate in an E2E test.

andrewda: Actually, I take that back. Talking about the user is fine. If you want to do it that way, just switch "press" to "presses".

alejandronanez:
Oh. Makes sense. Will update 😎 thanks!

Issue #588, opened 2017-10-27
Remove Sinon.js


alejandronanez:
We should remove `sinon.js` from our stack, Jest has its own way to do spies and mocks.

When you remove this, don't forget to update the tests.

andrewda:
👍

andrewda:
Apologies for using Sinon in the first place – I had seen it and was excited to use it without looking to Jest's prebuilt solutions 😂

alejandronanez:
haja np!

PR #591, opened 2017-10-27
refactor(tests): use jest mocks instead of sinon spies

stupergenius:
Per #588, removes the dependency on sinon and uses jest builtin mocks.


andrewda:
Will merge this as soon as tests pass 🎉

PR #600, opened 2017-10-29
fix(ui): Fix Bio section in auth-profile

jouderianjr:
| Devices tested?  | Android version Nougat |
| Related ticket?  | #569         |

---

## Screenshots

| Before   | After    |
| -------- | -------- |
| beforeauth | subtitlewithfontweight |

## Description

After the react-native-elements update, we've lost some styles,
this commit removes ellipsize from bio text and set `fontWeight` to normal.

I'm still using the text as subtitle but we could just use as a title, has a similar style that the old subtitle.

| Title   | Subtitle    |
| -------- | -------- |
| title | subtitlewithfontweight |

### DIFF
--- a/src/auth/screens/auth-profile.screen.js
+++ b/src/auth/screens/auth-profile.screen.js
@@ -45,10 +45,6 @@ const mapDispatchToProps = dispatch =>
 
 const styles = StyleSheet.create({
   listTitle: {
-    color: colors.black,
-    ...fonts.fontPrimary,
-  },
-  listSubTitle: {
     color: colors.greyDark,
     ...fonts.fontPrimary,
   },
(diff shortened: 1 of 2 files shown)


machour:
@jouderianjr echoing your suggestion, it would make more sense to use `title` instead of `subtitle`.

alejandronanez:
@jouderianjr please feel free to merge either you use `title` or not.

jouderianjr:
I'm just changing to `title` right now 💃 but I've noticed that I forgot about `user-profile.screen`, I'll fix it in that screen also.

jouderianjr:
Merging, Thank you everyone 💃

PR #601, opened 2017-10-29
refactor: use styled-components in user-list-item

shmesa22:
Related https://github.com/gitpoint/git-point/issues/532


shmesa22:
screen shot 2017-10-28 at 20 38 19
screen shot 2017-10-28 at 20 38 34

src/components/user-list-item.component.js
```diff
@@ -111,52 +124,44 @@ class UserListItemComponent extends Component {
             user.type === 'User' ? { user } : { organization: user }
           )}
         underlayColor={colors.greyLight}
-        style={!noBorderBottom && styles.borderContainer}
+        noBorderBottom={!noBorderBottom}
```
chinesedfan: Here is wired.

alejandronanez: hmm what do you mean by _wired_?

chinesedfan: @alejandronanez The prop name should be something like `hasBorderBottom`, because the value is `!noBorderBottom`.

src/components/user-list-item.component.js
```diff
@@ -1,65 +1,72 @@
...
-    marginLeft: 10,
-  },
-  iconContainer: {
-    flex: 0.15,
-    alignItems: 'flex-end',
-    justifyContent: 'center',
-  },
-});
+const ViewBorderContainer = styled.View`
+  border-bottom-color: ${props => props.hasBorderBottom && colors.greyLight};
```
machour: I think this will evaluate to a boolean and not work as expected .. shouldn't you be providing two different return values (just like the line below this one?)

src/components/user-list-item.component.js
```diff
@@ -1,65 +1,72 @@
...
+  border-bottom-width: ${props => (props.hasBorderBottom ? 1 : 0)};
+`;
+const TouchableBorderContainer = ViewBorderContainer.withComponent(
+  TouchableHighlight
+);
```
machour: nicely done 👍

machour:
Ping here to @shmesa22 😄 
Let me know if you need anything!

src/components/user-list-item.component.js
chinesedfan: @shmesa22 Right. It will cause a `invalid prop` error.

I am eager to merge styled-components PRs and your PR is very very close to be merged. So forgive me to fix it in your branch directly. Normally I shouldn't do that.

PR #610, opened 2017-10-30
refactor: NotificationListItem to used styled-components

LeoCp:
| Before   | After    |
| -------- | -------- |
|simulator screen shot - iphone 6 - 2017-10-30 at 12 26 57|simulator screen shot - iphone 6 - 2017-10-30 at 12 32 50|

## Description
Migrate NotificationListItem component to styled-components (#532)

### DIFF
--- a/src/components/notification-list-item.component.js
+++ b/src/components/notification-list-item.component.js
@@ -1,7 +1,8 @@
 /* eslint-disable no-nested-ternary */
 
 import React, { Component } from 'react';
-import { StyleSheet, View, TouchableOpacity, Text } from 'react-native';
+import styled from 'styled-components/native';
+import { View, TouchableOpacity } from 'react-native';
 import { Icon } from 'react-native-elements';
 
 import { colors, fonts, normalize } from 'config';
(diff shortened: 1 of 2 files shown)


src/components/notification-list-item.component.js
```diff
@@ -12,41 +13,48 @@ type Props = {
...
+  align-items: center;
+`;
+
+const TitleContainer = styled.View`
+  flex: 1;
+`;
+
+const Title = styled.Text`
+  color: ${colors.black};
+  ${{ ...fonts.fontPrimary }};
```
chinesedfan: Please try to use `font-family: ${styledFonts.fontPrimary}`, while `styledFonts` is imported from `config`.

src/components/notification-list-item.component.js
```diff
@@ -12,41 +13,48 @@ type Props = {
...
+  justify-content: center;
+`;
+
+const IconStyled = styled(Icon).attrs({
+  color: colors.grey,
+  size: 22,
+  type: 'octicon',
+})``;
+
+const CheckIcon = styled(IconStyled).attrs({ name: 'check' })``;
```
chinesedfan: Can the empty block be removed?

src/components/notification-list-item.component.js
```diff
@@ -59,9 +67,9 @@ export class NotificationListItem extends Component {
 
   getIconName = type => {
     switch (type) {
-      case 'commit':
+      case 'Commit':
```
chinesedfan: Is this a case bug that you find out?

LeoCp: yes, the type was wrong.

src/components/notification-list-item.component.js
```diff
@@ -70,41 +76,36 @@ export class NotificationListItem extends Component {
 
   render() {
     const { notification, iconAction, navigationAction } = this.props;
-
-    const TitleComponent = this.getComponentType();
+    const tag = this.getComponentType();
     const iconName = this.getIconName(notification.subject.type);
+    const titleComponentProps =
```
alejandronanez: Hey @LeoCp welcome!
Can you please move this into a class method and add proper tests to it? I don't feel comfortable with all this logic under `render() { ... }`

src/components/notification-list-item.component.js
alejandronanez: optional: I think it will be good to transform `type` to lowercase and to the check, this way we cover ourselves from any future change.
What do you think?

andrewda: Agree, but should be done in another PR.

machour:
@alejandronanez could you verify that the latest changes are okay with you and merge this baby in? ❤️

machour:
Validated on gitter with Alejandro 🕺

PR #611, opened 2017-10-31
fix(translation): correct typo and some words in french

ShiiFu:
## Description

I did some correction on French translation.


machour:
@ShiiFu merci beaucoup! 🎉 

Could you add yourself as a contributor using `yarn contributors:add`? Use the `translation` key.

machour:
@ShiiFu We have an ongoing migration to styled components if you want to contribute some more: #532
Thank you so much for the french proof reading buddy!

PR #612, opened 2017-10-31
fix(sc): Fix by stylelint

chinesedfan:
~~WIP.~~

Modified by stylelint's suggestions. Including:

- add missing commas
- fix unexpected token
- fix unsupported properties, like `padding-horizonal`
- fix unitless shorthands properties
- replace multi properties by shorthands

The last thing is that, is there a better way for empty tagged template? `stylelint` will throw a warning. Of course, maybe we can add a rule to ignore it.

```
const StyledListItem = styled(ListItem).attrs({
  ...
  hideChevron: props => props.unknown,
})``;
```

### DIFF
--- a/src/components/entity-info.component.js
+++ b/src/components/entity-info.component.js
@@ -25,7 +25,7 @@ const StyledListItem = styled(ListItem).attrs({
   },
   underlayColor: props => (props.unknown ? null : colors.greyLight),
   hideChevron: props => props.unknown,
-})``;
+});
 
 const getBlogLink = url =>
   url.substr(0, 4) === 'http' ? url : `http://${url}`;
--- a/src/components/entity-info.component.js
+++ b/src/components/entity-info.component.js
@@ -25,7 +25,7 @@ const StyledListItem = styled(ListItem).attrs({
   },
   underlayColor: props => (props.unknown ? null : colors.greyLight),
   hideChevron: props => props.unknown,
-});
+})``;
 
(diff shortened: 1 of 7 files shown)


src/components/entity-info.component.js
```diff
@@ -25,7 +25,7 @@ const StyledListItem = styled(ListItem).attrs({
   },
   underlayColor: props => (props.unknown ? null : colors.greyLight),
   hideChevron: props => props.unknown,
-});
+})``;
```
andrewda: ?

src/components/image-zoom.component.js
```diff
@@ -40,7 +40,7 @@ const CloseIcon = styled(Icon).attrs({
   size: 28,
   name: 'x',
   type: 'octicon',
-});
+})``;
```
andrewda: ?

src/components/entity-info.component.js
chinesedfan: @andrewda It is just my [last commit](https://github.com/gitpoint/git-point/pull/612/commits/e00addbaae91defb8e96ea9fdf8e13ff3f0b967b), whose purpose is reverting changes. I removed them by mistake, because `stylelint` complained about empty blocks, as the updated PR description mentioned. But we can't simply do that. "fn``" is different with "fn".

andrewda: Ah ok yea, didn't see that that was a revert commit.

PR #618, opened 2017-11-04
fix(crash): Fix crash in PullDiff screen

### LINKED ISSUE #616: Crash when view pull-diff.screen
Due to #515. If reverted, it is OK.

machour:
SyntaxHighlighter is not stylable as it uses the `style` props for something totally different.

Fixes #616


machour:
Merged this quick one in by myself to fix master (and so I can work on the PullDiff screen)
😘

Issue #584, opened 2017-10-26
Coverall badge not showing up in comments


machour:
![screen shot 2017-10-26 at 11 11 21 am](https://user-images.githubusercontent.com/304450/32047664-fe5779d2-ba46-11e7-88cf-2c0e6e2234be.png)

Here's what it should look like:

<img width="711" alt="screen shot 2017-10-26 at 11 13 10 am" src="https://user-images.githubusercontent.com/304450/32047686-11e14406-ba47-11e7-965d-a1da83be9333.png">

andrewda:
Is it because the badge is an SVG?

machour:
That's probably the cause.

Images handling could benefit from a good rewrite:
- Support SVG & WEBP
- Initially display a loader, then display the images with the appropriate dimensions when they're loaded (currently it's the same height for all images, hence that big white gap in the screenshot)

andrewda:
:thinking: SVGs seem to work fine in the README.md files. Is there any reason why they'd work there but not in comments?

machour:
README files are displayed in a WebView, whereas comments are rendered using https://github.com/jsdf/react-native-htmlview that transforms the HTML code into native views, so we're using <Image /> here, and it doesn't support SVG out of the box.

Should be easy to implement using something like https://github.com/react-native-community/react-native-svg

andrewda:
Gotchya. That sounds like it should work.

PR #621, opened 2017-11-05
feat(markdown): Add coverall badge support

machour:
Implements #584 

I tried to add support for SVG in `GithubHtmlView` but it ended up like this using `react-native-svg`, which seems to be the most popular lib up to now:
screen shot 2017-11-05 at 1 10 58 am

I ended up recreating the badge using styled-components, and it looks quiet realistic thanks to `<LinearGradient />`:
screen shot 2017-11-05 at 3 30 13 am
<sub>(*OCD: it's still 6px wider than the original, cause I had to do some weird stuff to get the rounded borders on only one side*)</sub>

The only problem I'm left with is accurately coloring the right side of the badge.

Fact is, the only information I have is the sentence written below the badge, and this give no hints whether coverage is considered good enough (green) or not (red), in regards of the project-configured threshold.

The only way I see for now, would be to fetch the SVG and decode its content to grab the color from there, but while we're there, we might as well grab the percentage from there (and remove the parsing from GithubHtmlView).

I'm curious for your opinions on this one :)

And of course, we could settle on a neutral lighter gray for now and worry about this in another PR 👯‍♂️

### DIFF
--- a/src/components/coverall-badge.component.js
+++ b/src/components/coverall-badge.component.js
@@ -46,12 +46,14 @@ const ValueContainer = styled(LinearGradient).attrs({
   padding-top: 1;
   align-items: center;
 `;
+
 const ValueBorderContainer = ValueContainer.extend`
   width: 20;
   border-radius: 3;
   position: absolute;
   left: 83;
 `;
+
 const Value = Label.extend`
   font-size: ${props => normalize(props.value === 100 ? 10 : 11)};
 `;
--- a/src/components/coverall-badge.component.js
+++ b/src/components/coverall-badge.component.js
@@ -63,7 +63,7 @@ export const CoverallBadge = ({ coverage = 100 }: Props) => (
     </LabelContainer>
     <ValueBorderContainer />
     <ValueContainer>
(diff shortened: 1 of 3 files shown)


machour:
@andrewda you're totally right, I might have had a case of nose to the grindstone on this one 😬 

The difficulty is that this img is provided by Github as a regular `` which takes an image URL, download it and identify its type through the first magic bytes, before processing it accordingly.

This  might raise some memory management concerns though.. 🤔

housseindjirdeh:
Hmm sort of agree with @andrewda here - I don't know if a single badge warrants its own component 🤔 

For example, should we do the same with Travis CI badges and so forth? Might be too much componentizing here (is that even a word? 😂).

Yeah something like `RemoteImage` makes more sense but like you mentioned, maybe not be the best route?

machour:
@housseindjirdeh yeah, this PR wasn't really well thought (but I had fun with LinearGradient so it was good anyways 😝).
Closing this and opening an issue for a better image support in markdown.

Issue #536, opened 2017-10-21
Input being chopped on issue page


yannbf:
I just installed gitpoint on my Moto G5 Plus, android 7, and noticed that when entering an issue, the comment input gets  chopped(you can easily notice by the placeholder on the bottom). Here's a link to the video showing it: https://drive.google.com/file/d/0B1YWQtOX-9n_dXdUQzFONUcwQ00/view?usp=drivesdk

yannbf:
Np! Actually I mean the section in the bottom where the placeholder "Add a comment here" is. 
See the image below:
![selection_293](https://user-images.githubusercontent.com/1671563/31922157-d72eed1c-b849-11e7-8544-1ac5180c554b.png)

housseindjirdeh:
Ooooh thank you, surprised I didn't catch that from the video.

Thanks for this :) We'll make sure it's resolved before our next patch release 💃

PR #622, opened 2017-11-05
fix(comment-input): Fix the input text being chopped up

machour:
| Version?         | master     |
| Devices tested?  | iPhone 7, Android Nexus (simulators) |
| Related ticket?  | #536       |

## Description

Android had a flickering problem when height started with 0.
Also adapted the height a bit more for the final input.

### DIFF
--- a/src/components/comment-input.component.js
+++ b/src/components/comment-input.component.js
@@ -51,7 +51,7 @@ const StyledPostButtonIcon = styled(PostButtonIcon)`
   color: ${props => (props.disabled ? colors.grey : colors.primaryDark)};
 `;
 
-const inputMaxHeight = Platform.select({
+const inputMinHeight = Platform.select({
   ios: 30,
   android: 37,
 });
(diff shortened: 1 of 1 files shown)


chinesedfan:
@machour So the culprit is the zero height? And the height adjustment is for better UX?

machour:
@chinesedfan It's quiet confusing, my bad.

Originally, I used the inspector to see the height on Android, and found it to be 37.
When I set it, I noticed an improvement, but further testing showed some weird cases where the bug would still happen.

By changing the initial height after that, I was not able to reproduce it anymore, and I figured I'll keep that 37 anyway as it is the actual result.

src/components/comment-input.component.js
```diff
@@ -51,6 +51,11 @@ const StyledPostButtonIcon = styled(PostButtonIcon)`
   color: ${props => (props.disabled ? colors.grey : colors.primaryDark)};
 `;
 
+const inputMaxHeight = Platform.select({
```
chinesedfan: Should be renamed as `inputMinHeight`?

machour: Totally! Got fooled by Math.max() 😸

machour:
@chinesedfan fixed, thanks!

PR #633, opened 2017-11-06
chore: Make travis run eslint first and fail immediately if needed

### LINKED ISSUE #632: Run eslint before coverage in Travis
I'm noticing that if a PR contain eslint problems, coverage is still being ran, and the travis build takes ages to fail.

It would be better to run eslint first, and bail out without running coverage if it fails.

machour:
Should fix #632.

Due to https://github.com/travis-ci/travis-ci/issues/1066, the solution is not as simple as inverting the two lines.
Instead, I made the script a one liner that runs eslint before coverage.

This is not tested as I don't want to create a fake & bogus PR :/

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -6,5 +6,4 @@ cache:
 node_js:
   - 7
 script:
-  - yarn run test:coverage
-  - yarn run eslint
+  - yarn run eslint && yarn run test:coverage


chinesedfan:
@machour We can learn something from [a project](https://github.com/styled-components/stylelint-processor-styled-components/blob/master/.travis.yml) I played recently. My PR [failed](https://travis-ci.org/styled-components/stylelint-processor-styled-components/builds/297494370) in the lint stage first, and no coverage task was triggered.

machour:
Oh, thank you for that! If I understand correctly, they split the various checks across different jobs, and job support stopping on failure.
Notice how prettier was still ran even if eslint failed there? (same problem as described here)

PR #629, opened 2017-11-06
feat(markdown): Add support for <sub>

TheCodeTalker:
@machour can you varify ? this what you want ....??

| Question         | Response    |
| ---------------- | ----------- |
| Version?         | v1.4.0      |
| Devices tested?  | iPhone 6... |
| Bug fix?         | no      |
| New feature?     | yes      |
| Includes tests?  | no      |
| All Tests pass?  | yes      |
| Related ticket?  |  #627      |

---

## Screenshots

after

PR #635, opened 2017-11-06
chore: Upgrade to all-contributors-cli 4.5.3

machour:
## Description

Upgrade to last version to benefit from two bug fixes (by yours truly 😄):

  * All contributors names are bold now (was only the case for first row)
  * Force users to pick at least one contribution type (no more `[]` when new comers add themselves)  

Regenerated the CONTRIBUTORS files as well.


machour:
Waiting for #629 to be merged in before messing up the contributors file :)

PR #637, opened 2017-11-07
refactor: use an enum for subject types

machour:
As noted in #610 review.


housseindjirdeh:
Definitely think we should (when we get time, nothing urgent) to take a step back and think about properly type checking things like this. Utilizing Flow more will definitely help with scenarios just like this throughout our codebase

PR #641, opened 2017-11-15
fix: Dynamic status bar

jouderianjr:
| Version?         | v1.3.1      |
| Devices tested?  | iPhone iOS 11, Android 7 |
| Related ticket?  | #4       |

---

## Screenshots

| Before   | After    |
| -------- | -------- |
| beforedynamic | afterdynamic |
| screen shot 2017-11-15 at 03 34 58 | screen shot 2017-11-15 at 03 36 59 |

## Description

The main idea here is using the `onNavigationStateChange`event for discovering what screen the user is navigating and change the statusBarStyle accordingly with the screen.

### DIFF
--- a/App.js
+++ b/App.js
@@ -47,6 +47,7 @@ class App extends Component {
     this.state = {
       rehydrated: false,
     };
+    this.statusBarHandler = this.statusBarHandler.bind(this);
   }
 
   componentWillMount() {
@@ -67,7 +68,6 @@ class App extends Component {
     );
 
     this.constructor.initLocale();
-    this.getCurrentRouteName = this.getCurrentRouteName.bind(this);
   }
 
   componentDidMount() {
@@ -88,13 +88,35 @@ class App extends Component {
       return null;
     }
     const route = navigationState.routes[navigationState.index];
-    // dive into nested navigators
 
     if (route.routes) {
       return this.getCurrentRouteName(route);
     }
 
     return route.routeName;
+  }
+
+  statusBarHandler(prev, next) {
+    const routeName = this.getCurrentRouteName(next);
+    const lightContentScreens = [
+      'MyProfile',
+      'Profile',
+      'Organization',
+      'Repository',
+      'Login',
+    ];
+
+    if (lightContentScreens.includes(routeName)) {
+      StatusBar.setTranslucent(true);
+      StatusBar.setBackgroundColor(
+        routeName === 'Login' ? colors.transparent : colors.black
+      );
+      StatusBar.setBarStyle('light-content');
+    } else {
+      StatusBar.setTranslucent(false);
+      StatusBar.setBackgroundColor(colors.grey);
+      StatusBar.setBarStyle('dark-content');
+    }
   }
 
   render() {
@@ -108,24 +130,9 @@ class App extends Component {
 
     return (
       <Provider store={configureStore}>
-        <GitPoint
-          onNavigationStateChange={(prev, next) => {
-            const darkScreens = [
-              'MyProfile',
-              'Profile',
-              'Organization',
-              'Repository',
-            ];
-
-            if (darkScreens.includes(this.getCurrentRouteName(next))) {
-              StatusBar.setBarStyle('light-content');
-              StatusBar.setBackgroundColor(colors.primaryDark);
-            } else {
-              StatusBar.setBarStyle('dark-content');
-              StatusBar.setBackgroundColor(colors.white);
-            }
-          }}
-        />
+        <GitPoint onNavigationStateChange={this.statusBarHandler}>
+          <StatusBar />
+        </GitPoint>
       </Provider>
     );
   }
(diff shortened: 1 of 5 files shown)


App.js
```diff
@@ -77,6 +83,42 @@ class App extends Component {
...
+      StatusBar.setTranslucent(true);
+      StatusBar.setBackgroundColor(
+        routeName === 'Login' ? colors.transparent : colors.black
+      );
+      StatusBar.setBarStyle('light-content');
+    } else {
+      StatusBar.setTranslucent(false);
+      StatusBar.setBackgroundColor(colors.grey);
+      StatusBar.setBarStyle('dark-content');
+    }
```
chinesedfan: @jouderianjr Can we config these informations in somewhere? So that we will not forget to update them once new routers were added.

jouderianjr:
Awesome guys, until the end of the week I will fix it 💃

machour:
Finally! 👯‍♂️

App.js
```diff
@@ -77,6 +83,42 @@ class App extends Component {
...
+      'Profile',
+      'Organization',
+      'Repository',
+      'Login',
+    ];
+
+    if (lightContentScreens.includes(routeName)) {
+      StatusBar.setTranslucent(true);
+      StatusBar.setBackgroundColor(
+        routeName === 'Login' ? colors.transparent : colors.black
```
machour: Can you replace `colors.black` by #262626. This would Android look like the iPhone:

Before:
screen shot 2017-11-21 at 3 58 04 pm

After:
screen shot 2017-11-21 at 4 49 17 pm

App.js
```diff
@@ -77,6 +83,42 @@ class App extends Component {
...
+      );
+      StatusBar.setBarStyle('light-content');
+    } else {
+      StatusBar.setTranslucent(false);
+      StatusBar.setBackgroundColor(colors.grey);
```
machour: colors.white look nicer to me (give the same transparent look on Android)

App.js
machour: We should make a new color in config for this. 
The existant dark gray on profile pages is already #262626 but created with a rgba() thingy (that seems unneeded)

App.js
jouderianjr: This transparent look is not a "default" look and feel of material design ( as described [here](https://developer.android.com/training/material/theme.html?hl=pt-br) ). What do you think about @machour? Should we keep the material design thing or it is better to have the same iOS design? 😄

jouderianjr:
@andrewda @chinesedfan Looks better? 💃  If not, I will glad to hear some suggestions, I don't know if is the best implementation for this. 😄

App.js
machour: I completely forgot about material design 😅 
Let's keep it like you did it then

App.js
```diff
@@ -77,6 +83,35 @@ class App extends Component {
...
+    StatusBar.setBackgroundColor(backgroundColor(routeName));
+    StatusBar.setBarStyle(barStyle);
+  };
+
+  statusBarHandler(prev, next) {
+    const routeName = this.getCurrentRouteName(next);
+
+    const statusBarConfig = lightScreens.includes(routeName)
+      ? lightStatusBar
+      : darkStatusBar;
```
chinesedfan: @jouderianjr How about moving these lines into `src/config/status-bar.js` and only exporting 1 function called `getStatusBarConfig(routerName)`? Suppose the 3rd kind of status bar was introduced, only the config file should be updated.

jouderianjr:
@chinesedfan Done! Good suggestion mate

App.js
```diff
@@ -77,6 +83,31 @@ class App extends Component {
...
+
+  statusBarHandler(prev, next) {
+    const routeName = this.getCurrentRouteName(next);
+
+    const { translucent, backgroundColor, barStyle } = getStatusBarConfig(
+      routeName
+    );
+
+    StatusBar.setTranslucent(translucent);
+    StatusBar.setBackgroundColor(backgroundColor(routeName));
```
chinesedfan: @jouderianjr Please return `backgroundColor` value, instead of a function. Except here everything is so good!

jouderianjr: `backgroundColor` in lightStatusBar depends of if you are in `Login` or not. But I have an Idea, I can transform `lightStatusBar` and `darkStatusBar` in a function that receives the routeName and return the object.

Example: 

```
const lightStatusBar = routeName => {
  translucent: true,
  backgroundColor: routeName === 'Login' ? colors.transparent : colors.black,
  barStyle: 'light-content',
};
```

What do you think about?

chinesedfan: Right, that's what I want.

chinesedfan: And you may rename `lightStatusBar` as `getLightStatusBar`.

jouderianjr:
done @chinesedfan 💃

jouderianjr:
I think it good to go now @chinesedfan @machour @andrewda 💃

machour:
Just tested and validated with latest modifications. Merging this baby in!

PR #643, opened 2017-11-21
docs: add acknowledgments section

lex111:
Earlier this month we received licenses for free use of IDE's from JetBrains, I think it worth mentioning this and add a note that the new members of the organization (active contributors) can get it.

### DIFF
--- a/README.md
+++ b/README.md
@@ -123,5 +123,4 @@ Support this project by becoming a sponsor. Your logo will show up here with a l
 
 ## Acknowledgments
 
-Thanks to [JetBrains](https://www.jetbrains.com) for supporting us with [free Open Source License](https://www.jetbrains.com/buy/opensource). If you will actively participate in this project you can receive this one. Feel free to contribute!
-
+Thanks to [JetBrains](https://www.jetbrains.com) for supporting us with a [free Open Source License](https://www.jetbrains.com/buy/opensource).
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -120,4 +121,7 @@ Support this project by becoming a sponsor. Your logo will show up here with a l
 <a href="https://opencollective.com/git-point/sponsor/8/website" target="_blank"><img src="https://opencollective.com/git-point/sponsor/8/avatar.svg"></a>
 <a href="https://opencollective.com/git-point/sponsor/9/website" target="_blank"><img src="https://opencollective.com/git-point/sponsor/9/avatar.svg"></a>
 
+## Acknowledgments
+
+Thanks to [JetBrains](https://www.jetbrains.com) for supporting us with [free Open Source License](https://www.jetbrains.com/buy/opensource). If you will actively participate in this project you can receive this one. Feel free to contribute!
```
machour: .. supporting us with **a** ..

I would personally remove the second sentence, as this is a simple acknowledgment. If someone is actively participating & notice this acknowledgment, he'll know his entitled and will ask for it on gitter

lex111: Hmm, I thought the opposite, thus attracting more contributors. It turns out that the third sentence does not make sense, if the second one does not exist?

machour: I meant second and third sentences sorry :D 

I think it's better to have people attracted to GitPoint because they love GitPoint and want to make it better, rather then participating to have a free IDE license.

lex111: Good, although it could be stimulate them.

PR #647, opened 2017-11-25
docs: remove codesponsor.io

machour:
## Description

Remove codesponsor.io, as the service is [shutting down](http://mailchi.mp/f9b57b5ea377/code-sponsor-is-shutting-down-on-december-8) 😞


lex111:
Merged, since no longer works.

Issue #581, opened 2017-10-25
Notifications badge count has styling issues


chinesedfan:
Separated from #569.

![image](https://user-images.githubusercontent.com/12476932/31921499-ac0b161c-b83d-11e7-96f4-cba1b7dd6494.png)

JSLinn727:
So you want it to be a complete circle? I will take this issue just need to know exactly what you want it to look like.

alejandronanez:
@JSLinn727 yeah it should be a full circle with the number in the middle.
Thanks for your interest!

chinesedfan:
Fixed by #648.

PR #648, opened 2017-11-25
fix: re-enable color prop for the badge component

machour:
| Devices tested?  | iPhone 7, Android Nexus 5 |
| Related ticket?  | #581        |

---

## Screenshots

| Before   | After    |
| -------- | -------- |
| screen shot 2017-11-25 at 2 41 37 pmscreen shot 2017-11-25 at 2 46 08 pm | screen shot 2017-11-25 at 2 48 19 pm screen shot 2017-11-25 at 2 48 12 pm  |

## Description

#535 seems to have lost the `color` prop for the badge text, making it always black (the default).
This fixed the style be reading from that prop again when it's available.

### DIFF
--- a/src/components/badge.component.js
+++ b/src/components/badge.component.js
@@ -19,15 +19,14 @@ const BadgeContainer = styled.View`
   height: 18;
   border-color: ${colors.alabaster};
   border-width: 1;
-  ${({ backgroundColor }) => `background-color: ${backgroundColor};`};
+  background-color: ${({ backgroundColor }) => backgroundColor};
 `;
 
 const BadgeText = styled.Text`
   ${{ ...fonts.fontPrimaryBold }};
   background-color: transparent;
-  ${({ color }) => `color: ${color || colors.black};`};
-  ${({ largeText }) =>
-    `font-size: ${largeText ? normalize(9.5) : normalize(7)};`};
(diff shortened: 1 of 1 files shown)


src/components/badge.component.js
```diff
@@ -25,6 +25,7 @@ const BadgeContainer = styled.View`
 const BadgeText = styled.Text`
   ${{ ...fonts.fontPrimaryBold }};
   background-color: transparent;
+  ${({ color }) => `color: ${color || colors.black};`};
```
chinesedfan: It reminds me a question that, what's the differences between this way with `color: ${{color} => (color || color.black)};`?

chinesedfan: Both styles are OK. But the one I mentioned is more similar with styled-components [official documents](https://www.styled-components.com/docs/basics#passed-props).

machour: @chinesedfan just had the time to test, it's exactly the same.
I switched to the syntax you proposed as it's clearer.

I guess the difference between the two syntaxes, is that with the one I originally used, you can choose to output no style at all based on tests, which is not possible with the other one:

```js
${({ foo }) => foo ? `color: '#FFCC00'` : ''};
```

You could also output more styles in the condition

Issue #604, opened 2017-10-29
Integrate styled components checking tasks


chinesedfan:
It was a pity that we introduced so many styles problems by styled components refactoring. I suggest to integrate 2 more tasks into our Travis CI.

- [`stylelint-processor-styled-components`](https://github.com/styled-components/stylelint-processor-styled-components), which guarantees basic CSS writing.
- [`styled-components-native-code-mod`](https://github.com/styled-components/styled-components-native-code-mod), which enhances unitless checking.
    - Now it adds units to all properties, but I hope to fork it and only rewrite those necessary shorthands properties.

Of course, they will increase our testing time, but the safety should be more important. And I will also fix the current code base by them at first.

machour:
@chinesedfan I wouldn't mind adding `stylelint-processor-styled-components` as part of CI, but I'm not sure about `styled-components-native-code-mod` which looks more like a conversion utility.

Also, unless I'm missing something, current code base is broken because of an [incompatibility](https://github.com/react-native-training/react-native-elements/issues/679) with `react-native-elements`.

alejandronanez:
I second @machour, We should integrate `stylelint-processor-styled-components` But not the other one, it looks like a conversion tool.

@machour what do you think about rolling back any changes that involve SC + `react-native-elements` (or go with a workaround) ?

machour:
@alejandronanez was about to reply from the app but it crashed 😆 (you fixed it in #602)

A big +1 on rolling back failing PRs for now until react-native-elements is updated.

alejandronanez:
Just to clarify, I meant only roll back changes related to RNE.

machour:
Yes yes :)

jamesg1:
What were the actual benefits of using styled-components? Seems to me it has some issues with supporting third party component libraries.

I use https://github.com/css-modules/css-modules with a work project and enjoy it as it keeps the CSS seperate to the JS.

alejandronanez:
Hey @jamesg1 we already had a discussion about it (I can’t find the issue right now - on mobile) 
My 2 cents on CSS modules is that they work _ok_ but have some restrictions that styled-components address in a really good way.

chinesedfan:
@machour @alejandronanez Right, `styled-components-native-code-mod` is not a real test tool. Adding it here is just for reminding us to have the same ability to avoid unitless shorthands properties crash.

[...]

PR #652, opened 2017-12-03
chore(lint): Add stylelint script

chinesedfan:
As discussed in #604, I integrated `stylelint` by following styled-components [tutorial](https://www.styled-components.com/docs/tooling#stylelint).

Please notice that even with the official `stylelint-processor-styled-components`, `stylelint` is not 100% compatible with styled-components. So I added some customized rules.

```
  "rules": {
    // 1
    "block-no-empty": null,
    // Inspired by https://github.com/styled-components/css-to-react-native/blob/master/src/tokenTypes.js#L31
    // to avoid unitless shorthands properties crash
    "declaration-property-value-whitelist": {
      "margin": ["/^(0|[+-]?(\\d*\\.)?\\d+px)( (0|[+-]?(\\d*\\.)?\\d+px))?$/"],
      "padding": ["/^(0|[+-]?(\\d*\\.)?\\d+px)( (0|[+-]?(\\d*\\.)?\\d+px))?$/"],
    },
    // 2
    "declaration-colon-newline-after": null,
    "value-list-max-empty-lines": null
  },
```

- Cases like 1, i.e. [comment-input.component.js](https://github.com/gitpoint/git-point/blob/master/src/components/comment-input.component.js#L42)
```
const PostButtonIcon = styled(Icon).attrs({
  color: props => (props.disabled ? colors.grey : colors.primaryDark),
})``; // here will generate an empty block
```

- Cases like 2, i.e. [badge.component.js](https://github.com/gitpoint/git-point/blob/master/src/components/badge.component.js#L28)
```
const BadgeText = styled.Text`
  ${{ ...fonts.fontPrimaryBold }};
  background-color: transparent;
  ${({ largeText }) =>
    `font-size: ${largeText ? normalize(9.5) : normalize(7)};`};
`;
```
It will be parsed as,
```
const BadgeText = styled.Text`
  ${{ ...fonts.fontPrimaryBold }};
  background-color: transparent;
  -styled-mixin1: dummyValue
/* dummyComment */;
`;
```

### DIFF
--- a/.stylelintrc
+++ b/.stylelintrc
@@ -1,0 +1,8 @@
+{
+  "processors": ["stylelint-processor-styled-components"],
+  "extends": [
+    "stylelint-config-standard",
+    "stylelint-config-styled-components"
+  ],
+  "syntax": "scss"
+}
--- a/.stylelintrc
+++ b/.stylelintrc
@@ -4,5 +4,14 @@
     "stylelint-config-standard",
     "stylelint-config-styled-components"
   ],
+  "rules": {
+    "block-no-empty": null,
+    "declaration-colon-newline-after": null,
+    "declaration-property-value-whitelist": {
+      "margin": ["/^(0|[+-]?(\\d*\\.)?\\d+px)( (0|[+-]?(\\d*\\.)?\\d+px))?$/"],
+      "padding": ["/^(0|[+-]?(\\d*\\.)?\\d+px)( (0|[+-]?(\\d*\\.)?\\d+px))?$/"],
+    },
+    "value-list-max-empty-lines": null
+  },
   "syntax": "scss"
 }
--- a/.travis.yml
+++ b/.travis.yml
@@ -6,4 +6,4 @@ cache:
 node_js:
   - 7
 script:
-  - yarn run eslint && yarn run test:coverage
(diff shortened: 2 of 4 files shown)


machour:
Thanks for looking into this! For the second case, would formatting the rule as you suggested in #648 solve the compatibility issue?

chinesedfan:
@machour No, rules in case 2 will only affect multi-lines declarations. And I replied with my understandings of those 2 styles in #648 directly.

machour:
Applied this PR diff on my macbook, got the following error message:
screen shot 2017-12-04 at 2 31 25 pm

Any clue what might be going on?

chinesedfan:
@machour I searched around and found that `$export is not a function` was usually caused by [circular dependencies](https://stackoverflow.com/questions/43222249/webpack-2-3-3-typeerror-export-is-not-a-function). After this PR applied, 2 different versions of `core-js` were installed.

It is an annoy feature of `yarn` that `yarn add <pkg>` only updates dependencies of pkg, but doesn't update those already in `yarn.lock`. Finally, I fixed it by keeping only one version of `core-js`.(In fact, if we just installed core-js 2.5.1, there were still no errors)

PR #659, opened 2017-12-10
fix(issue): Fix wrong repository was used

### LINKED ISSUE #655: Crash in notifications feed after install & login
| Devices tested?  | Android Nexus 5 |

---

## What Currently Happens?

After a fresh install and login, if you directly click an issue title in the notifications feed, before visiting the repository, the application will crash:

screen shot 2017-11-25 at 2 49 35 pm

## What Do You Expect To Happen?

The issue screen to be displayed.

## My Reproduction Steps

1. Uninstall the app (to wipe the redux store)
1. Compile it & run it
1. Login to github
1. Click the first issue title you see on your feed

chinesedfan:
| Devices tested?  | iPhone 6 simulator, iOS 11.0 |
| Related ticket?  | #655        |

---

## Description

Fixes #655.

This bug was introduced by #438. `getIssueEvents` should not be invoked until `this.props.issue/repository` were updated.

In fact, besides the crash problem, the unfixed codes may also get wrong issue events due to `getRepository` was not returned.

### DIFF
--- a/src/issue/screens/issue.screen.js
+++ b/src/issue/screens/issue.screen.js
@@ -199,28 +199,33 @@ class Issue extends Component {
       .replace(`${v3.root}/repos/`, '')
       .replace(/([^/]+\/[^/]+)\/issues\/\d+$/, '$1');
 
-    const repoName = repository.name;
-    const owner = repository.owner.login;
-
     Promise.all([
       getIssueFromUrl(issueURL),
       getIssueComments(`${issueURL}/comments`),
-    ]).then(() => {
-      const issue = this.props.issue;
+    ])
+      .then(() => {
+        const issue = this.props.issue;
 
-      if (repository.full_name !== issueRepository) {
-        Promise.all([
-          getRepository(issue.repository_url),
-          getContributors(this.getContributorsLink(issue.repository_url)),
-        ]).then(() => {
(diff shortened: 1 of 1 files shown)


jouderianjr:
Hey guys,

I think this piece of code is getting confusing. 

So I have 2 possible suggestions:

1 - send all this piece of code to an action creator, I think this should not be handled on the screen component. ( complex one)

2 - Or just split this into some small functions to improve the readability.

What do you think guys? 😄

machour:
I really don't have an opinion on this, as I still plan to re-propose the big redux refactoring whenever I get my head back in it.

If @chinesedfan is okay and can address this rapidly, why not. I'd hate to see the crash fix (part of 1.4 which is long overdue) being delayed ;)

jouderianjr:
@machour yeah, I totally okay with fix first for the new version then we improve it after 💃

chinesedfan:
@jouderianjr I agree with you that moving those codes to an action creator. Components should not include so complex logics. But it is risky to refactor now. Hope we can improve it later.

housseindjirdeh:
Agreed folks - some refactoring is definitely needed in this and many other screen components. When we get time, we'll hopefully get to them 🙏

Issue #55, opened 2017-07-14
Show last search results


FlorianTaut:
It would be helpful to show the last search results or last watched repos when starting a new search request

housseindjirdeh:
I think this would be awesome to have - will probably want to make sure a decision is made on which search bar tool is being used in #2. Once that's done and merged in, I think this will nice to have.

How I'm envisioning things: 

When you try searching for a repository - some recent searches (up to 5?) show up as a dropdown or something along those lines. When you try searching for a user - the same happens but with users. (And if search for issues gets merged in #12, then the same there as well).

Another alternative is just to show some recent searches in the search screen instead of the placeholder (`Search for any...`). _However_, there's an issue open for showing trending repositories #44 and I think that can be reserved for that instead.

Always open to suggestions 💬

PR #661, opened 2017-12-10
feat(store): blacklist everything but auth from persistent storage

machour:
| Devices tested?  | iPhone 7, Android Nexus 5 |
| Related ticket?  | none        |

---

## Description

This PR removes everything but `auth` from the persistent storage.
Motivation is that there's no point in keeping this information, and it's sometime a pain in the A.
Example:

* Go to an issue that crashes because of some markdown problem
* Restart the app
* Go to another issue that shouldn't crash.
* App crashes because it's still parsing the old markdown from store, before getting the new one

This was partially discussed here: https://github.com/gitpoint/git-point/pull/366#discussion_r141516846

### DIFF
--- a/App.js
+++ b/App.js
@@ -61,14 +61,7 @@ class App extends Component {
       {
         storage: AsyncStorage,
         transforms: [encryptor],
-        blacklist: [
-          'user',
-          'issue',
-          'repository',
-          'organization',
-          'search',
-          'notifications',
-        ],
+        whitelist: ['auth'],
       },
       () => {
         this.setState({ rehydrated: true });
--- a/App.js
+++ b/App.js
@@ -61,7 +61,14 @@ class App extends Component {
       {
         storage: AsyncStorage,
         transforms: [encryptor],
-        blacklist: ['user'],
+        blacklist: [
+          'user',
+          'issue',
+          'repository',
+          'organization',
+          'search',
+          'notifications',
+        ],
       },
       () => {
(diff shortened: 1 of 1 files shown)


machour:
Perfect illustration of my problem: clicked an issue on git-point, the app crashed due to the previous issue visited on one of my repositories:

screen shot 2017-12-10 at 11 57 05 am

screen shot 2017-12-10 at 11 55 11 am

App.js
```diff
@@ -61,7 +61,14 @@ class App extends Component {
...
         transforms: [encryptor],
-        blacklist: ['user'],
+        blacklist: [
+          'user',
+          'issue',
+          'repository',
+          'organization',
+          'search',
+          'notifications',
+        ],
```
chinesedfan: @machour Setting `whitelist: ['auth']` can be simpler.

chinesedfan: And if history search depends on `search`, we should also add it to the whitelist. I am not sure.

machour: Do we have a search history? Oo
Looking into `whitelist` now :)

machour: Fixed to use `whitelist`, thank you for the hint @chinesedfan!
Kept `search` blacklisted as it currently only holds only the result set from the API, and there's no history yet.

chinesedfan: @machour Thanks for your investigation. #55 mentions historical search results but seems not have implemented yet.

Issue #663, opened 2017-12-10
chore: Replace moment.js by date-fns


machour:
As mentionned in #523, we are currently using `moment/min/moment-with-locales.min` which weights 289.5K (63.8K) only to output the time ago information. 

Even "worse", we don't even use the locales from the library for the translations, instead, we override them with our own wordings.

@ocarreterom proposed we switch to `date-fns` which is kind of a `lodash` for date handling, and will have a wayyyy smaller footprint, he will be working on the PR for this change 🎉 

/cc @lex111

ocarreterom:
@machour there is a problem. date-fns doesn't provide format translations for all the languages git-point supports.
[distanceInWords](https://date-fns.org/v1.29.0/docs/distanceInWords), [distanceInWordsStrict](https://date-fns.org/v1.29.0/docs/distanceInWordsStrict) and [distanceInWordsToNow](https://date-fns.org/v1.29.0/docs/distanceInWordsToNow) always returns de units in words. I can't format as `1h`.

It's a problem?

machour:
yes, we definitely need those.
if you check our translations, we have a [common.relative](https://github.com/gitpoint/git-point/blob/master/src/locale/languages/en.js#L326) object with all the strings we currently display.

Looking at `date-fns`, it seems possible to adapt it our way:

```
var distanceInWords = require('date-fns/distance_in_words')
// Require Esperanto locale
var eoLocale = require('date-fns/locale/eo')

var result = distanceInWords(
  new Date(2016, 7, 1),
  new Date(2015, 0, 1),
  {locale: eoLocale} // Pass the locale as an option
)
```

Seems to me like we just need to mimic the `eoLocale` object and override its `formatDistance` property with our own definition.

https://github.com/date-fns/date-fns/blob/master/src/locale/eo/_lib/formatDistance/index.js

ocarreterom:
OK.
This works, but we need to change the translations of relativeTime.
```js

relativeTime: {
  lessThanXSeconds: {
    one: 'less than a second',
    other: 'less than {{count}} seconds'
  },
  xSeconds: {
    one: '1 second',
    other: '{{count}} seconds'
  }
}
```

PR #664, opened 2017-12-10
refactor: replace momentjs by date-fns

ocarreterom:
| Devices tested?  | iPhone 7 iOS 11.2 (simulator) |
| All Tests pass?  | Running infinite      |
| Related ticket?  | #663        |

---

## Description
Use date-fns to represent relative time.

#### Some questions:
- Personally I like the long format (hours) more than the short one (h). Can be replaced easily.
- This PR have changes that affect translations. Only `en.js` has been updated (locales test not pass).
- I can't verify the tests have passed since it seems not work :/
captura de pantalla 2017-12-10 a las 22 31 06


housseindjirdeh:
This is so great, thank you @ocarreterom <3

Yep the only thing I can think of is reverting back to single character representations of time as mentioned by @machour. Otherwise this looks beautiful.

ocarreterom:
Comments addressed. Ready for review again :)

machour:
@ocarreterom had a quick test with the last version of the PR, and it looks like you made it! 🎉 🎉 🎉 

Noticed a tiny glitch though, while testing your code on the simulator and the old one on my device, while looking at the events.

For events older than 1 hour, your PR was showing "1h" while the old code showed "2h".
Checked Github web, and it was also saying 2h. Refreshed several times, and right now, your PR started showing "2h".

Seems like timeago & date-fns don't make the same approximations when it comes to rounding.
date-fns, I think, considers "1h50" to be "1h" when momentjs reports it as "2h".

I don't think that this would be a show stopper, but I'm wondering if date-fns behavior can be fine-tuned with some configuration 🤔 

Also, could you take care of the rest of translations so that the Travis CI build passes? It should be pretty straight forward. Once done, I'll give this PR a wider test to validate it and it will be merged right after 1.4 is released.

Thank you again for your work, this one have been itching me for so long ❤️

machour:
PS: You can reproduce that tiny shift by checking the time for the last comment here: https://github.com/machour/git-point-playground/issues/34#issuecomment-351651675
it's also the case for minutes.

Again, this is no biggie for me, but I'm not sure about the rest of the team.

machour:
Reported this difference in this comparison thread: https://github.com/date-fns/date-fns/issues/275#issuecomment-351668021

ocarreterom:
@machour thanks for your review.

Yes, you are right. The rounding is `floor` by default. I will change it to `round` and add the rest of translation just today. 👍 

I'm not sure it works exactly like moment.js since it uses custom range breakdowns 🙄 : https://momentjs.com/docs/#/displaying/fromnow/

machour:
@ocarreterom Sorry I took so long to get back to you on this PR 😞 

I just gave it a new test and it's working as expected 🎉 🎉 🎉  

Last thing before being able to merge this PR would be to fix the Travis build. It seems that these two locale files needs to be updated:

- zh-tw.js
- eu.js

Could you take care of this please? 🙏 🙏 

Here are a few comparisons between this PR and GitHub web for the other reviewers to be amazed at:

screen shot 2018-01-15 at 10 25 01 pm

screen shot 2018-01-15 at 10 25 39 pm

ocarreterom:
@machour no problem. It's ready :)

machour:
Thank you for doing this so fast ❤️

machour:
@housseindjirdeh ready for approval & merge!

package.json
```diff
@@ -52,14 +52,14 @@
     ]
   },
   "dependencies": {
+    "date-fns": "^1.29.0",
```
housseindjirdeh: 🎉

housseindjirdeh:
@machour Let's definitely explore possibilities to see if we can reduce bundle sizes if possible

PR #673, opened 2017-12-20
fix(crash): don't navigate to issue if it's a release notification

### LINKED ISSUE #667: Crash touching notification release tag
| Devices tested?  | iPhone 7 iOS 11.2.1 |

---

## What Currently Happens?

App crashes when touching the notification release tag in notifications view.

## My Reproduction Steps

1. Go to notifications screen
2. Touch a notification release tag

captura de pantalla 2017-12-19 a las 9 35 20

machour:
| Related ticket?  | #667         |

---

## Description

We now only try to navigate to the issue if the `notification.subject.type` was either a `PullRequest` or an `Issue`.

We still need it to be clickable to mark the notification as read.

Also made a release notification display the accurate icon (tag).

Closes #667.

### DIFF
--- a/src/components/notification-list-item.component.js
+++ b/src/components/notification-list-item.component.js
@@ -57,6 +57,7 @@ const IconStyled = styled(Icon).attrs({
 const SubjectType = {
   commit: 'Commit',
   pull: 'PullRequest',
+  release: 'Release',
 };
 
 export class NotificationListItem extends Component {
@@ -76,6 +77,8 @@ export class NotificationListItem extends Component {
         return 'git-commit';
       case SubjectType.pull:
         return 'git-pull-request';
+      case SubjectType.release:
+        return 'tag';
       default:
         return 'issue-opened';
     }
--- a/src/components/notification-list-item.component.js
+++ b/src/components/notification-list-item.component.js
@@ -87,7 +87,8 @@ export class NotificationListItem extends Component {
   getTitleComponentProps = () => {
     const { notification, navigationAction } = this.props;
 
-    return notification.subject.type === SubjectType.commit
+    return notification.subject.type === SubjectType.commit ||
+      notification.subject.type === SubjectType.release
       ? {}
       : {
           onPress: () => navigationAction(notification),
--- a/src/notifications/screens/notifications.screen.js
+++ b/src/notifications/screens/notifications.screen.js
@@ -346,7 +346,6 @@ class Notifications extends Component {
     const { markAsRead, navigation } = this.props;
 
     markAsRead(notification.id);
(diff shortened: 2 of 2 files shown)


src/notifications/screens/notifications.screen.js
```diff
@@ -346,6 +346,14 @@ class Notifications extends Component {
     const { markAsRead, navigation } = this.props;
 
     markAsRead(notification.id);
+
+    if (
+      notification.subject.type !== 'PullRequest' &&
+      notification.subject.type !== 'Issue'
+    ) {
+      return;
+    }
```
ocarreterom: How about to check the opposite condition? In this case we can add or remove this exceptions in the future.
Also could be worth move `SubjectType` to an other file or export it to be imported in `notifications.screen.js`.
```js
if (notification.subject.type === SubjectType.release) {
  return;
}
```

machour: @ocarreterom checking the opposite means a potential crash at runtime :/
Good idea for SubjectType, but where would you put this enum? 🤔

ocarreterom: We have the problem that the project structure for components' directory is too flat.
Maybe each component should have their own directory: e.g `/components/notification/notification-list-item.js`.
Then, the enum could be in `/components/notification/notification-subject-type.js`.

The quick solution is to export that enum from `notification-list-item.component.js` waiting for a structure refactor. 😞

machour: In a perfect world, these definitions would be part of the API definition, not inside components.. 💭

ocarreterom: I'm agree. 👍

src/components/notification-list-item.component.js
```diff
@@ -76,6 +77,8 @@ export class NotificationListItem extends Component {
         return 'git-commit';
       case SubjectType.pull:
         return 'git-pull-request';
+      case SubjectType.release:
+        return 'tag';
       default:
         return 'issue-opened';
     }
```
chinesedfan: We'd better filter Release events in function `getTitleComponentProps` (just after here, but Github doesn't support to add review comments at unmodified codes). Then it can avoid exporting SubjectType.

I have to admit that this component is not implemented very straightforward. Because the navigationAction 
 prop is not used as press handler directly.

chinesedfan: @machour Can you have a rest during your huge refactoring? I bet you can fix this one in 5min  😏

machour: There you go! :D

PR #677, opened 2017-12-20
test: add tests for NotificationIcon and RepositorySectionTitle

ZahraTee:
Given NotificationIcon is a connected component, I decided just to export and test the undecorated component inside, I can look into testing the connected component too if necessary. There aren't any other examples in the repo yet to indicate what to do in this case. :)

(Part of #518)


src/components/notification-icon.component.js
```diff
@@ -20,7 +20,7 @@ const mapStateToProps = state => ({
   notificationsCount: state.notifications.notificationsCount,
 });
 
-class NotificationIconComponent extends Component {
+export class NotificationIconComponent extends Component {
```
andrewda: Why does this need to be exported? Can you test `NotificationIcon` which is exported at the bottom of this file?

ZahraTee: Given NotificationIcon is this component but connected to a Redux store, the options were to either just test this inner component or to test NotificationIcon itself and mock out the store etc. I can look into the latter option.

ZahraTee: Although exporting and testing the unwrapped Component seems to be recommended by Redux: https://redux.js.org/docs/recipes/WritingTests.html#connected-components

andrewda: Yep, you're right! This should be exported.

andrewda: (not to self: we might want to make this a dumb component w/o access to redux state and do all that state management elsewhere)

.all-contributorsrc
```diff
@@ -655,6 +655,16 @@
...
+    },
+    {
+      "login": "ZahraTee",
+      "name": "Zahra Traboulsi",
+      "avatar_url": "https://avatars0.githubusercontent.com/u/5565340?v=4",
+      "profile": "http://www.zahra.tech",
+      "contributions": [
+        "code",
+        "test"
+      ]
```

PR #682, opened 2017-12-28
refactor: migrate organization screen to styled-components

ZahraTee:
## Screenshots

| Before   | After    |
| -------- | -------- |
|before|after|

Part of #532.

### DIFF
--- a/src/organization/screens/organization-profile.screen.js
+++ b/src/organization/screens/organization-profile.screen.js
@@ -1,7 +1,8 @@
 import React, { Component } from 'react';
+import styled from 'styled-components/native';
 import { bindActionCreators } from 'redux';
 import { connect } from 'react-redux';
-import { StyleSheet, RefreshControl } from 'react-native';
+import { RefreshControl } from 'react-native';
 import { ListItem } from 'react-native-elements';
 import { createStructuredSelector } from 'reselect';
 import ActionSheet from 'react-native-actionsheet';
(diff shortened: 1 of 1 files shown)


machour:
Since this is affecting the UI, could you provide us with two screenshots of the organization screen?
(One without the PR, and one with this PR applied)

src/organization/screens/organization-profile.screen.js
```diff
@@ -47,16 +48,16 @@ const actionCreators = {
 
 const actions = dispatch => bindActionCreators(actionCreators, dispatch);
 
-const styles = StyleSheet.create({
-  listTitle: {
+const DescriptionListItem = styled(ListItem).attrs({
+  titleStyle: {
```
machour: I think this one can be removed (unused)

ZahraTee:
Added the screenshots :)

machour:
LGTM, thanks!

src/organization/screens/organization-profile.screen.js
```diff
@@ -47,16 +48,12 @@ const actionCreators = {
...
-    ...fonts.fontPrimary,
-  },
-  listSubTitle: {
+const DescriptionListItem = styled(ListItem).attrs({
+  subtitleStyle: {
     color: colors.greyDark,
     ...fonts.fontPrimary,
   },
-});
+})``;
```
andrewda: Can we remove this? Was probably just added by `prettier` for some reason, but I think you can safely get rid of it.

machour: It's needed by styled components when using `attrs()`

andrewda: Ahh alright, yes you're right.

PR #683, opened 2017-12-30
refactor: migrate user repo list screen to styled-components

ZahraTee:
## Screenshots

| Before   | After    |
| -------- | -------- |
|before-repos|after-repos|

Part of #532.


src/user/screens/repository-list.screen.js
```diff
@@ -131,9 +130,9 @@ class RepositoryList extends Component {
     return (
       <ViewContainer>
         <View>
```
machour: Do you think we can get rid of this `<View>` ?

PR #684, opened 2017-12-30
refactor: migrate notifications screen to styled-components

ZahraTee:
## Screenshots

| Before   | After    |
| -------- | -------- |
|before-none| after-none|
|before-notifications|after-notifications|

Part of #532.

### DIFF
--- a/src/notifications/screens/notifications.screen.js
+++ b/src/notifications/screens/notifications.screen.js
@@ -1,18 +1,10 @@
 /* eslint-disable no-nested-ternary */
 /* eslint-disable no-shadow */
 import React, { Component } from 'react';
+import styled from 'styled-components/native';
 import { connect } from 'react-redux';
 import { bindActionCreators } from 'redux';
-import {
-  StyleSheet,
-  FlatList,
-  View,
-  ScrollView,
-  Text,
-  TouchableOpacity,
-  Image,
-  Platform,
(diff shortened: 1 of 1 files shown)


src/notifications/screens/notifications.screen.js
```diff
@@ -493,22 +481,17 @@ class Notifications extends Component {
...
                 ListEmptyComponent={
                   !isLoadingNewNotifications && (
-                    <View
-                      style={[
-                        styles.textContainer,
-                        { height: contentBlockHeight },
-                      ]}
-                    >
-                      <Text style={styles.noneTitle}>
+                    <TextContainer height={contentBlockHeight}>
```
machour: Shouldn't `TextContainer` styled definition read and apply the `height` prop?

ZahraTee: I thought that too but I found that you actually don't need to... It seems that styled.Views can apply values (including height) passed in as props directly and handles null values better. I built out a small example here: https://snack.expo.io/Bys9FZumz

I can change it if it's more explicit but I thought leaving it like this is a bit neater.

chinesedfan: > styled-components pass on all their props.

From [documents](https://www.styled-components.com/docs/basics#passed-props), I learned something new. :)

PR #686, opened 2018-01-06
chore(ci): Upgrade CI Node.js version to 8

chinesedfan:
> yarn run v1.3.2
warning You are using Node "7.10.1" which is not supported and may encounter bugs or unexpected behavior. Yarn supports the following semver range: "^4.8.0 || ^5.7.0 || ^6.2.2 || >=8.0.0"

The CI warns that `yarn` is not compatible with Node.js 7. We also complained that the CI is too slow. Let me try to upgrade Node.js version and compare the time.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -4,6 +4,6 @@ cache:
   directories:
     - node_modules
 node_js:
-  - 7
+  - 8
 script:
   - yarn run eslint && yarn run stylelint && yarn run test:coverage


chinesedfan:
Super fast! Now only need 4 minutes, compared with 15~20 minutes in Node 7.

machour:
Wonderful dude!

Issue #521, opened 2017-10-20
Different font for create event


chinesedfan:
Just a little curious. Is it set on purpose?

![createevent](https://user-images.githubusercontent.com/1736154/31806327-8efb3c50-b52c-11e7-930b-7271f611783f.png)

andrewda:
GitHub displays branches in the `code format` in the event list, so this replicates that.

chinesedfan:
OK. The branch styles in Github are much better, while ours is, em, a little wired. 😆

Hope someone can improve it. (cc @housseindjirdeh)

<img width="695" alt="2017-10-20 14 36 59" src="https://user-images.githubusercontent.com/1736154/31808073-4100a4d0-b537-11e7-9b80-27134d457337.png">

housseindjirdeh:
Haha yeah I _tried_ to replicate it :P

Definitely think we can mimic the way it actually looks on the site :+1: 

Will tag this for the 1.5.0 release :+1:

ZahraTee:
Can I pick this up? :)

chinesedfan:
@ZahraTee Of course! Glad to hear that.

PR #687, opened 2018-01-06
refactor: migrate events screen to styled-components

ZahraTee:
| Before   | After    |
| -------- | -------- |
|events-before|events-after|

I also noticed the branch name (see first event) wasn't getting the LinkBranchDescription code style as it does in 'create events' despite the fact GH also has that styling for 'push events'. So I added that.

(Part of #532)

### DIFF
--- a/src/auth/screens/events.screen.js
+++ b/src/auth/screens/events.screen.js
@@ -1,9 +1,10 @@
 /* eslint react/prop-types: 0 */
 /* eslint-disable no-shadow */
 import React, { Component } from 'react';
+import styled from 'styled-components/native';
 import { connect } from 'react-redux';
 import { bindActionCreators } from 'redux';
-import { StyleSheet, Text, FlatList, View } from 'react-native';
+import { Text, FlatList, View } from 'react-native';
 import moment from 'moment/min/moment-with-locales.min';
 
 import { LoadingUserListItem, UserListItem, ViewContainer } from 'components';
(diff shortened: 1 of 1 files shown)


src/auth/screens/events.screen.js
```diff
@@ -29,52 +30,58 @@ const mapDispatchToProps = dispatch =>
...
-    textAlign: 'center',
-    ...fonts.fontPrimary,
-  },
-});
+const DescriptionContainer = styled.Text`
+  justify-content: center;
+  flex: 1;
+  margin-left: 10;
+  color: ${colors.primaryDark};
+  font-family: ${fonts.fontPrimaryLight.fontFamily};
```
chinesedfan: For styled components fonts, try to import from "src/config/styled-fonts.js".

src/auth/screens/events.screen.js
```diff
@@ -206,103 +213,88 @@ class Events extends Component {
...
         return (
-          <Text style={styles.linkDescription}>
+          <LinkBranchDescription>
             {userEvent.payload.ref.replace('refs/heads/', '')}
-          </Text>
+          </LinkBranchDescription>
         );
       case 'ReleaseEvent':
-        return `${userEvent.payload.release.id}`;
+        return `888 ${userEvent.payload.release.id}`;
```
chinesedfan: 888? Seems like you mistyped something.

src/auth/screens/events.screen.js
```diff
@@ -206,103 +213,88 @@ class Events extends Component {
...
+          <LinkDescription
             onPress={() => {
               if (userEvent.action !== 'deleted') {
                 this.navigateToRepository(userEvent);
               }
             }}
           >
-            {userEvent.repo.name}
-          </Text>
+            999 {userEvent.repo.name}
```
chinesedfan: 999

src/auth/screens/events.screen.js
```diff
@@ -29,52 +30,58 @@ const mapDispatchToProps = dispatch =>
...
+const LinkBranchDescription = styled.Text`
+  font-family: ${fonts.fontCode.fontFamily};
+`;
+
+const DeletedLinkBranchDescription = styled.Text`
+  color: ${colors.greyDarkest};
+  font-family: ${fonts.fontCode.fontFamily};
+`;
+
+const Date = styled.Text`
```
chinesedfan: "Date" sounds too general. It usually reminds me "new Date()" in Javascript. How about "DateText" or something else?

src/auth/screens/events.screen.js
```diff
@@ -206,103 +213,88 @@ class Events extends Component {
...
-          </Text>
+          </LinkDescription>
         );
       case 'PushEvent':
         return (
-          <Text style={styles.linkDescription}>
+          <LinkBranchDescription>
             {userEvent.payload.ref.replace('refs/heads/', '')}
-          </Text>
+          </LinkBranchDescription>
```
chinesedfan: Nice catch! Then we can update together to resolve #521.

src/auth/screens/events.screen.js
ZahraTee: Oops. Forgotten remnants of my pro debugging technique... 😅

src/auth/screens/events.screen.js
ZahraTee: I used fonts because I remember seeing conflicting info on whether there was a plan to get rid of styledFonts on some PRs because it was duplicating the font listing... like a comment by @machour in #557 and the discussion in #533.

chinesedfan: @ZahraTee Yes, I see two kinds of codes. How about choosing any one but don't creating the 3rd one? 😄 
- `${fonts.fontPrimaryLight};`
- `font-family: ${styledFonts.fontPrimaryLight};`

machour:
I'm all for dropping styledFonts and using {...fonts.xxxx}. 
A PR switching the remaining uses of styledFonts and getting rid of it would be awesome!

PR #691, opened 2018-01-14
feat(events): enhance styling for events screen branch names

ZahraTee:
| Devices tested?  | iPhone X, FWVGA slider |
| New feature?     | yes...?     |
| Related ticket?  | #521        |

---

## Screenshots

| Before   | After    |
| -------- | -------- |
|screen shot 2018-01-14 at 01 11 08|screen shot 2018-01-15 at 23 59 35|
|screen shot 2018-01-14 at 01 10 52|screen shot 2018-01-15 at 23 58 06|

## Description

Updated the styling for branch links in events to match the GitHub style.
~Also changed the styling for the deleted branch names as well to be red, as I felt the grey was a bit too washed out and didn't match the updated default style.~


machour:
Thank you for taking care of this!

I'm not really sure about coloring in red though, as it looks kind of "aggressive" to the eye. Maybe a darker gray on a light gray background for the box would feel more confortable and appear as disabled right away.

Also, what do you think about changing the Monospace font we're using? I think something like [Roboto Mono Regular](https://fonts.google.com/specimen/Roboto+Mono) would look crisp rendered a little smaller than the rest of the text and with a bit more left & right padding for the box. 🤔

ZahraTee:
@machour 

I changed it to grey and changed the code font to Inconsolata which is pretty similar to Consolas which is used on GitHub web. Although having just pushed that is a fairly significant change in terms of screens throughout the app. 🤔

So, increasing the branch label padding is tricky because you can't control padding or margins of nested <Text> elements. I've been trying to get UserListItem to accept Views for the title prop but it's not really working out so far.

machour:
@ZahraTee this looks so nice! 🎉 🎉 

Had a look at the padding thing, and as you said, it's uncontrollable with styles. But I managed to trick it `&nbsp;`-style 😄  
This does the trick:

```diff
 const Datestamp = styled.Text`
@@ -214,7 +221,8 @@ class Events extends Component {
         ) {
           return (
             <LinkBranchDescription>
-              {userEvent.payload.ref}
+              {' '}{userEvent.payload.ref}{' '}
             </LinkBranchDescription>
           );
         } else if (userEvent.payload.ref_type === 'repository') {
       case 'ReleaseEvent':
```

resulting in this rendering which looks pretty decent to me, what do you think ?
screen shot 2018-01-16 at 12 17 41 am

You will also need to copy the ttf file to `android/app/src/main/assets/fonts/` for it to work on Android (tested it for you):
screen shot 2018-01-16 at 12 31 47 am

/cc @housseindjirdeh for cosmetic changes approval

ZahraTee:
@machour Just FYI I did add the android font but it didn't get added to the commit the first time round because of the .gitignore file added in #401. Probably worth watching out for.

machour:
erm, there might be a logic flaw in #401 as I ran `yarn run link` and the font weren't copied :/
don't hesitate to open a new issue for this problem!

src/auth/screens/events.screen.js
```diff
@@ -219,7 +219,8 @@ class Events extends Component {
         ) {
           return (
             <LinkBranchDescription>
-              {userEvent.payload.ref}
+              {' '}
+              {userEvent.payload.ref}{' '}
```
machour: Same thing should be done for `case 'PushEvent':` line 288

src/config/fonts.js
```diff
@@ -4,5 +4,5 @@ export const fonts = {
   fontPrimaryItalic: { fontFamily: 'Nunito-Italic' },
   fontPrimarySemiBold: { fontFamily: 'Nunito-SemiBold' },
   fontPrimaryBold: { fontFamily: 'Nunito-Bold' },
-  fontCode: { fontFamily: 'Menlo' },
+  fontCode: { fontFamily: 'Inconsolata-Regular' },
```
housseindjirdeh: 👍

Issue #649, opened 2017-11-30
SafeAreaView - iPhone 10


housseindjirdeh:
With the release of iPhone 10 - we'll need to make sure our screens handle the notch :)

[SafeAreaView](https://github.com/react-community/react-native-safe-area-view) should be out in 0.50.0 for us to use where applicable

ZahraTee:
The v3 release of styled-components includes a styled.SafeAreaView as well :)

PR #707, opened 2018-01-29
feat: support iPhone X

machour:
| Devices tested?  | iPhone X, iPhone 8 |
| New feature?     | yes     |
| Related ticket?  | #649        |

## Description

I experimented with the iPhone X and RN SafeAreaView, here are my findings for now:

**1 - Making iPhone X use the full display:** You need to add a correctly sized launch image in Xcode (1125x2436). Without this, the app will be "boxed"

**2 - AreaSafeView:** Turns out we don't really need it as we're using react-native-navigation which seems to handle it for us:

| without SafeAreaView | With SafeAreaView |
|------------------------|---------------------|
simulator screen shot - iphone x - 2018-01-29 at 16 21 31|simulator screen shot - iphone x - 2018-01-29 at 16 05 14|

**3 - Designing for iPhone X:** I think that the main strategy is simple: *avoid having elements placed at the top of the screen (Y: 0)*

The only problematic screens I saw were the one with an element on the top:
  * Notification screen: the button group
  * Search screen: the search bar
  * User/Repo profile: the cog & the header section once parallaxed

Example:
screen shot 2018-01-29 at 4 26 24 pm

Good news, this can be fixed really easily by adding some extra padding-top where needed. 
StatusBar awesomeness from @jouderianjr still work as expected.

I guess the only thing left to decide is: Do we nudge down the headers where needed for all devices, or do we do that only for iPhone X ? 🤔 

And here are some nice screenshots for the road: 🎉 🎉 🎉 

| | | |
|-|-|-|
|simulator screen shot - iphone x - 2018-01-29 at 16 11 08|simulator screen shot - iphone x - 2018-01-29 at 16 21 31|simulator screen shot - iphone x - 2018-01-29 at 16 28 39
simulator screen shot - iphone x - 2018-01-29 at 16 28 41|simulator screen shot - iphone x - 2018-01-29 at 16 28 47


housseindjirdeh:
😍 😍 😍 😍 

Amazing, React Navigation does take care of `SafeAreaView`. Stumbled on this issue (https://github.com/react-navigation/react-navigation/issues/2589) ages ago, glad to see they resolved it and supporting it themselves 😊 

I guess the only two screens are the `search` and `notifications` screen where we've manually inserted header elements correct? 🤔 

I think we can check how the header bar renders for iPhone 8 and under with the extra padding. If it looks just fine, than we should be good. However if the extra padding looks off (and I think it probably would), we can add a utility method that uses [Platform version](https://facebook.github.io/react-native/docs/platform-specific-code.html#detecting-the-ios-version) and on that condition add extra padding - otherwise remain the same.

Let me know if you think that's a good idea mate

machour:
Checked on an older iPhone with that extra padding, and damn, it's as ugly as you predicted :D

Not sure testing the os version is the good way to go though, as any device OS can be upgraded.

Been pocking around `react-native-device-info` which we already use as a dep, it should be helpful:
  https://github.com/rebeccahughes/react-native-device-info/blob/master/RNDeviceInfo/RNDeviceInfo.m#L102

I'll give that a try and finish this PR

housseindjirdeh:
Ah I'm sorry, silly me. I don't know why I suggested OS version 😓 

That's a great idea, we already use `react-native-device-info` for adding a hash for the encrypted store so I think that can work if we find the device number through there :)

machour:
@housseindjirdeh Had to update `react-native-device-info` as the version we used returned `Simulator` for `DeviceInfo.getModel()` which wasn't really helpful :D

Tested all the screens that I could think of, everything looks fine to me now!

(we could improve the @3x.png file, but I can't seem to find the logo SVG)

machour:
Woops, I forgot to mock the library in `testenv.js`

housseindjirdeh:
Thanks fammmmmm 🚀

PR #715, opened 2018-02-03
Refactor StateBadge component

rkpasia:
| Devices tested  | iPhone 6 |
| Related ticket?  | #532        |

---

## Screenshots

| Before   | After    |
| -------- | -------- |
|before|after|
|before|after|

## Description

This PR is related with the update of the StateBadge component. Now it uses styled-components as a driver to style the user interface.

I kept the possibility to override the styles of the StateBadge through the `style` prop. The object that it accept it's a StyleSheet object.

Also I removed from issue-description component the object `styles`. It was only related to the StateBadge component but the styles declared didn't influenced the UI placement of the latter.

### DIFF
--- a/src/components/issue-description.component.js
+++ b/src/components/issue-description.component.js
@@ -1,5 +1,5 @@
 import React, { Component } from 'react';
-import { StyleSheet, ActivityIndicator } from 'react-native';
+import { ActivityIndicator } from 'react-native';
 import { ListItem } from 'react-native-elements';
 import Parse from 'parse-diff';
 import styled from 'styled-components';
@@ -14,13 +14,6 @@ import {
 import { translate, relativeTimeToNow } from 'utils';
 import { colors, fonts, normalize } from 'config';
 import { v3 } from 'api';
-
-const styles = StyleSheet.create({
-  badge: {
-    alignItems: 'flex-end',
-    justifyContent: 'center',
-  },
-});
 
 const HeaderContainer = styled.View`
   flex-direction: row;
@@ -164,7 +157,6 @@ export class IssueDescription extends Component {
             (issue.pull_request &&
               !isPendingCheckMerge && (
                 <StateBadge
-                  style={styles.badge}
                   issue={issue}
                   isMerged={isMerged && issue.pull_request}
                   locale={locale}
--- a/src/components/state-badge.component.js
+++ b/src/components/state-badge.component.js
@@ -1,5 +1,5 @@
 import React from 'react';
-import { Text, View, StyleSheet } from 'react-native';
+import styled from 'styled-components/native';
 
 import { colors, fonts, normalize } from 'config';
 import { translate } from 'utils';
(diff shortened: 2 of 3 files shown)


src/components/state-badge.component.js
```diff
@@ -1,5 +1,5 @@
 import React from 'react';
-import { Text, View, StyleSheet } from 'react-native';
+import styled from 'styled-components/native';
```
machour: no need to use '/native' anymore, just use this:

```
import styled from 'styled-components';
```

src/components/state-badge.component.js
```diff
@@ -58,19 +47,25 @@ export const StateBadge = ({
...
+      issueStateColor = colors.purple;
+      break;
+    case 'open':
+      issueStateColor = colors.green;
+      break;
+    case 'closed':
+      issueStateColor = colors.red;
+      break;
+    default:
+      issueStateColor = colors.gray;
```
machour: could you do something like this instead for a better readability?

```js
const stateColor = {
  merged: colors.purple,
  open: colors.green,
  closed: colors.red,
};
const issueStateColor = stateColor[issueState] ? stateColor[issueState] : colors.gray;
```

src/components/issue-description.component.js
```diff
@@ -15,13 +15,6 @@ import { translate, relativeTimeToNow } from 'utils';
 import { colors, fonts, normalize } from 'config';
 import { v3 } from 'api';
 
-const styles = StyleSheet.create({
```
machour: why was this removed?

src/components/state-badge.component.js
```diff
@@ -58,19 +47,25 @@ export const StateBadge = ({
...
   }
 
   return (
-    <View style={[styles.badge, style, issueStyle]}>
-      <Text style={styles.text}>{issueText}</Text>
-    </View>
+    <Badge color={issueStateColor} style={style}>
```
machour: You are not passing `style` anymore from `issue-description.component.js`, but you're still using the prop here .. Could you give it a second look?

rkpasia: Yeah I thought to get completely rid of the style prop but it could still be useful in case there is a need to override something from a parent component without modifying again this one instead.

But it's up for discussion, don't know if it is really needed or not so I left it there 😆

src/components/issue-description.component.js
rkpasia: It was used just for the Badge styles so I removed it. I don't think is needed here, also styled-components are already imported.
Also the styles defined for the Badge don't change the UI so they are useless for what I can saw.

rkpasia:
Don't really know how the tests work yet 😆 I'll get through them tomorrow 👍

machour:
No worries mate! We use jest for testing, and everything is located under `<root>/__tests__`
You can run the full test suite in your console like this:

```
yarn test
```

or you can run a single test like this (which is waaaayyyy faster when fixing something 😄) :
```
yarn test __tests__/tests/components/StateBadge.js
``` 

happy debugging 🕺

rkpasia:
What should I do to augment the coverage?

src/components/state-badge.component.js
chinesedfan: Prop `style` is also used in `repository-section-title.component.js`, which is a useless component. I remembered @housseindjirdeh confirmed that in somewhere, but I forgot the detail issue/pr number. I will collect feedbacks in #725.

machour: it was on Gitter, I added the relevant screenshot in the issue you opened.

PR #726, opened 2018-02-14
fix(repository): fix repo view when repos return 404

### LINKED ISSUE #690: Wrong info when i click on forked to
When some forked you get the info like „USER forked REPO to REPO2“
When i click REPO2 in only get a view which looks like a repo but without title etc

dhamanutd:
instead of blank view, now will show not found message when error 404

BREAKING CHANGE: fix repository view when the repos 404 not found

fix #690

| Devices tested?  | OnePlus One |
| New feature?     | yes/no      |
| Related ticket?  | #690         |

---

## Screenshots

| Before   | After    |
| -------- | -------- |
|before|after|

## Description

- [x] Change action reducer `getRepository` to return Promise reject when response status code is not 200 OK
- [x] Make `notFoundMsg` props in the repository screen and component to take string from github api when return 404.
- [x] Change repository reducer in `GET_REPOSITORY.ERROR` to return initialState instead of recent state.
  - In purpose to make repository screen not showing any info from latest state, if there are error 404.
- [x] Create logic to show not found messages, hide owner section, and disabled view code button if error 404.


src/api/index.js
```diff
@@ -118,7 +118,9 @@ export const v3 = {
   getJson: async (url, accessToken) => {
     const response = await v3.call(url, v3.parameters(accessToken));
 
-    return response.json();
+    return response.ok
+      ? response.json()
+      : response.json().then(err => Promise.reject(err));
```
chinesedfan: In fact, `response.json().then(err => Promise.reject(err))` is the same with `response.json()`. You handle the error in `then` and throw out identically. Please correct me if I misunderstood your thinking.

src/components/repository-profile.component.js
```diff
@@ -143,13 +145,17 @@ export const RepositoryProfile = ({
...
+          (repository.fork && 'repo-forked') ||
+          'repo'
+        }
         type="octicon"
         size={45}
         color={colors.greyLight}
       />
 
-      <Text style={styles.title}>{repository.name || ' '}</Text>
+      <Text style={styles.title}>{notFoundMsg || repository.name || ' '}</Text>
```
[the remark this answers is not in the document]
dhamanutd: is that mean, you want make "Unkown" or "Not Found" as default, right?

src/api/index.js
dhamanutd: @chinesedfan 
this because when the response status is 4xx it always handled by `.then()`, that's why i give return `Promise.reject'  to make it handled by `.catch()`. But if you think there better way to distinct beetween 200 OK and 4xx Error, let me know.

Another suggestion from me is to implement Axios and Redux-Saga to the app for better handling failures.

src/components/repository-profile.component.js
```diff
@@ -133,7 +135,7 @@ export const RepositoryProfile = ({
         )}
 
       <Text style={[styles.languageInfoTitle]}>
-        {repository.language || ' '}
+        {notFoundMsg || repository.language || ' '}
```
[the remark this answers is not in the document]
dhamanutd: @chinesedfan  can i use google translate for another language?

src/components/repository-profile.component.js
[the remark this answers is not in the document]
chinesedfan: My point is that the current "Not Found" is a message returned by Github API. Setting it as the value of Text directly will make the usage of this Text unclear. We'd better,
- declare a corresponding string constant to represent special repository name, as I said before
- or, create another certain UI component (maybe like a Toast) to show it

src/components/repository-profile.component.js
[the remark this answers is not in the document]
chinesedfan: @dhamanutd No, please just leave them in English. Other contributors will pick them up and translate.

src/api/index.js
chinesedfan: @dhamanutd I see. Then you can modify `getRepository` to call `v3.call`, and check `response.status`. Please refer to [`checkReadMe`](https://github.com/gitpoint/git-point/blob/master/src/repository/repository.action.js#L159).

src/components/repository-profile.component.js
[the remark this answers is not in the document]
dhamanutd: @chinesedfan if want to use Toast, i will install library react-native-simple-toast. What do you think? does it needed?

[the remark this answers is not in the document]
chinesedfan: Sure, I don't mind to introduce a new library as long as it is useful. And we can also wait for some feedbacks from other reviewers, like @machour.

dhamanutd:
@chinesedfan i did some of you requested to changes. you can review it now.

src/components/repository-profile.component.js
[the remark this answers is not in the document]
arthurdenner: Nice feature. Just an observation, the key has a typo, should be `unknownLanguage`.

src/repository/repository.action.js
```diff
@@ -36,7 +36,18 @@ export const getRepository = url => {
...
+      .then(response => {
+        let json = {};
+
+        if (response.status === 200) {
+          json = response.json();
+        } else {
+          json = response.json().then(err => Promise.reject(err));
+        }
+
+        return json;
```
chinesedfan: How about writing in one line, `return response.status === 200 ? response.json() : Promise.reject()`?

src/repository/screens/repository.screen.js
```diff
@@ -39,6 +40,7 @@ const mapStateToProps = state => ({
   starred: state.repository.starred,
   forked: state.repository.forked,
   subscribed: state.repository.subscribed,
+  hasRepoExist: state.repository.hasRepoExist,
```
chinesedfan: Shall we keep `hasRepoExist` and `error` together? Or either is enough?

src/locale/languages/en.js
```diff
@@ -177,6 +177,8 @@ export const en = {
   },
   repository: {
     main: {
+      notFoundRepo: 'Repository is not found',
+      unknowLanguage: 'Unknown',
```
chinesedfan: `unknownLanguage` as @arthurdenner suggested? Sorry I am not a native English speaker.

src/repository/screens/repository.screen.js
dhamanutd: Actually i want to remove `error`. But as you said before, you said maybe it will be used in future.

So, will me to delete that?

chinesedfan: I prefer to keep `error`, because `error` is needed for Toast, while `hasRepoExist` can be replaced by `!error`. What do you think about?

dhamanutd: @chinesedfan i think `hasRepoExist` is only to set state if repo exist or not, but `error` is to get value whatever it is not only about repo exist or not. I think both are used.

chinesedfan: Yes, there are tricky differences between them. It requires us to handle these 2 fields more carefully.

src/repository/repository.reducer.js
```diff
@@ -62,19 +63,21 @@ export const repositoryReducer = (state = initialState, action = {}) => {
...
+        hasRepoExist: true,
+        error: '',
         isPendingRepository: false,
       };
     case GET_REPOSITORY.ERROR:
       return {
-        ...state,
+        ...initialState,
         error: action.payload,
-        isPendingRepository: false,
```
chinesedfan: I just noticed that this line was removed. Do you want to keep pending state when an error occurs?

src/repository/repository.reducer.js
```diff
@@ -62,19 +63,21 @@ export const repositoryReducer = (state = initialState, action = {}) => {
       return {
         ...state,
         issues: [],
+        error: '',
```
chinesedfan: If you persist to keep 2 fields, I suggest to clear `hasRepoExist` when pending and error. If repo1 set it to be `true`, then repo2 will get wrong value during pending or error.

src/repository/repository.action.js
```diff
@@ -36,7 +36,12 @@ export const getRepository = url => {
     dispatch({ type: GET_REPOSITORY.PENDING });
 
     return v3
-      .getJson(url, accessToken)
+      .get(url, accessToken)
+      .then(response => {
+        return response.status === 200
+          ? response.json()
+          : response.json().then(err => Promise.reject(err));
```
chinesedfan: `response.json()` will never throw an error, no matter `response.status === 200` or not, right? So I think `Promise.reject()` is enough. We don't need to call `response.json()` again.

src/repository/repository.reducer.js
dhamanutd: @chinesedfan it's because we take `...initialState` that already had `isPendingRepository` state as `false`.

src/repository/repository.action.js
dhamanutd: @chinesedfan if we just only using `Promise.reject()`, we will not pass any value from that. It will just call `.catch()` but didn't pass any data you get from API response. In this case we will not get the "Not Found" messages from github API.

That's why first i need to get `json()` data from API, then i pass that data in the `Promise.reject(err)`.

Make both if true (equal 200) or false (not equal 200) will send same json formatted data.

src/repository/repository.reducer.js
dhamanutd: @chinesedfan in current changes, i've tried to test in real world. I don't face any problem when openning from one repo to another.

src/repository/repository.action.js
chinesedfan: @dhamanutd Oh, hope I understand your idea this time. Do you mean `response.json().then((json) => response.status === 200 ? json : Promise.reject(json))`?

src/repository/repository.reducer.js
chinesedfan: Yes, as you reset with `initialState`, the error state is cleared. And the pending state is too fast to be noticed.

dhamanutd: Understand, okay i will set `hasRepoExist` when pending too

src/repository/repository.action.js
dhamanutd: @chinesedfan yes, that's what i'm intended to

chinesedfan: Good. Can you update it? Then I think there are no problems from my side. Thanks for your patient revisions!

dhamanutd: @chinesedfan done

.all-contributorsrc
```diff
@@ -740,6 +740,15 @@
       "contributions": [
         "code"
       ]
+    },
+    {
+      "login": "dhamanutd",
```
housseindjirdeh: :tada: 🚀 🚀

Issue #728, opened 2018-02-14
Odd vertical spaces in markdown comments


machour:
| Question         | Response    |
| ---------------- | ----------- |
| Version?         | v1.4.1      |
| Devices tested?  | iPhone 7 |

---

## What Currently Happens?

Markdown comments contains weird vertical spacings:

![simulator screen shot - iphone 6 - 2018-02-14 at 10 55 53](https://user-images.githubusercontent.com/304450/36198347-a28ab3bc-1176-11e8-856a-b6d03c2410ce.png)

## What Do You Expect To Happen?

Not that much spaces! It seems like Github introduced extra spaces in their HTML transformation.

## My Reproduction Steps

Check this PR: https://github.com/theliturgists/app/pull/45

housseindjirdeh:
Just got merged in :)

THANK YOU FAM

PR #729, opened 2018-02-14
fix: remove extra spaces from empty text nodes

machour:
| Related ticket?  | #728         |

---

## Screenshots

| Before   | After    |
| -------- | -------- |
| simulator screen shot - iphone 6 - 2018-02-14 at 10 55 53 |  simulator screen shot - iphone 6 - 2018-02-14 at 11 00 26 |

## Description

Instead of testing that the content of the text chunk if "\n" to replace it with our own vertical spacer, I now make sure that trim(text) is empty and it contains at least one "\n".

GitHub probably changed from "\n" to "_space_\n" or something like that, which made the previous implemented test fail.

### DIFF
--- a/src/components/github-htmlview.component.js
+++ b/src/components/github-htmlview.component.js
@@ -300,7 +300,8 @@ export class GithubHtmlView extends Component {
               <Text
                 style={linkStyle}
                 onPress={() =>
-                  onLinkPress({ ...node, attribs: { href: node.attribs.src } })}
+                  onLinkPress({ ...node, attribs: { href: node.attribs.src } })
+                }
               >
                 [{node.attribs.alt}]
                 {'\n'}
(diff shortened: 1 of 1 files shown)


src/components/github-htmlview.component.js
```diff
@@ -388,13 +389,13 @@ export class GithubHtmlView extends Component {
       };
 
       if (_node.type === 'text') {
-        if (_node.data === '\n') {
+        if (_node.data.trim() === '' && _node.data.includes('\n')) {
```
chinesedfan: @machour Is it possible that `_node.data` contains more than one `\n`? Then we need to render as multiply empty lines.

If not, everything is handled very well.

machour: Markdown only takes a single line break into account, so it should be okay

PR #742, opened 2018-03-01
fix: fix style-components write style and composeWithDevTools use

MrZhang123:
when i read the code i find style-components write not good in App.js , I think maybe forget add semicolon 😄, and other questions is why use `composeWithDevTools` in productions ? so I try to fixed it . Let me know what you think.


App.js
```diff
@@ -21,7 +21,7 @@ import { configureStore } from './root.store';
 
 const Container = styled.View`
   align-items: center;
-  background-color: ${colors.white}
+  background-color: ${colors.white};
```
chinesedfan: Nice catch! Our stylelint script only checks `src/**/*.js`.

root.store.js
```diff
@@ -31,14 +31,11 @@ let store;
...
-    rootReducer,
-    compose(getMiddleware(), ...getEnhancers())
-  );
-} else {
-  store = createStore(
     rootReducer,
     composeWithDevTools(getMiddleware(), ...getEnhancers())
   );
+} else {
+  store = createStore(rootReducer, compose(getMiddleware(), ...getEnhancers()));
```
chinesedfan: > Reactotron is different, it differs from React Native Debugger, and both of these applications do not work together, so composeWithDevTools is not added when Reactotron is enabled.

According to #330, maybe we should fix as,
```
if (__DEV) {
    if (process.env.TRON_ENABLED) {
        // Reactotron.createStore
    } else {
        // composeWithDevTools
    }
} else {
    // createStore
}
```

src/auth/screens/auth-profile.screen.js
```diff
@@ -42,6 +43,8 @@ const mapDispatchToProps = dispatch =>
     dispatch
   );
 
+const hairlineWidth = StyleSheet.hairlineWidth;
```
chinesedfan: Hi @MrZhang123, maybe you should create another PR to change all borders to hair line.

src/auth/screens/auth-profile.screen.js
```diff
@@ -145,6 +149,7 @@ class AuthProfile extends Component {
 
           {hasInitialUser &&
             user.bio &&
+            !isPending &&
```
chinesedfan: If it matters, we need a new PR, too.

src/components/section-list.component.js
```diff
@@ -15,9 +15,9 @@ type Props = {
   noItemsMessage: string,
   buttonAction: Function,
 };
-
+const hairlineWidth = StyleSheet.hairlineWidth;
 const Section = styled.View`
-  margin-top: 15;
+  /* margin-top: 15; */
```
chinesedfan: Do you want to remove this line?

src/components/section-list.component.js
```diff
@@ -35,8 +35,8 @@ const TitleView = styled.View`
 const StyledList = styled(List).attrs({
   containerStyle: {
     marginTop: 0,
-    borderBottomColor: colors.grey,
-    borderBottomWidth: 1,
+    borderBottomColor: colors.greyLight,
```
chinesedfan: Color is changed.

src/components/section-list.component.js
MrZhang123: yes, i think some style need to change, so i change something , but it's no complete

chinesedfan: Sure, friendly reminding, please keep a PR independent. And you can test different kinds of styles locally or in other branches. Then our maintainers will not be notified to review. :)

chinesedfan:
Landed in https://github.com/gitpoint/git-point/commit/738ab3435a8f88f7813e847a7efb7bd0075cde77. And thank @MrZhang123 all the same.

Issue #273, opened 2017-08-22
Bringing translations to the next level.


machour:
Dears,

As mentionned in a previous issue, and on gitter, **I** think that the current system used for handling translations is flawed. 

Here are the major problems I've seen/I predict:

  * keywords doesn't often make a lot of sense (had to check the source code to guess the context/meaning of some of them)
  * contributing becomes "harder". I'm introducing a new feature with some sentences, I have to check : If these sentences exists as keywords somewhere. If not, where should I put them ? under what namespaces ?
  * even reading the code source gets tedious: https://github.com/gitpoint/git-point/blob/master/src/auth/screens/events.screen.js#L491-L502
  * *concatenation is bad*: `user + ' made ' + repository + 'public'`, is perfect for English, but it may not be suited for another language. Let's not speak about the mess it will be when we introduce a first RTL language
  * keeping translations up to date is tedious, as if someone changes the english value of the key, there's a big chance it will be missed by translators : https://github.com/gitpoint/git-point/commit/c5b90dad760f80e678b197b0c3520ab64812e685
  * currently, adding a entry `en.js` break all other languages. Contributors are using Google Translate in order to avoid this: https://github.com/gitpoint/git-point/pull/271#issue-251850959
  * last but not least, making sure that all translations are up to date before pushing next release to App stores is impossible for @housseindjirdeh and will require him to wait for all translators to respond.

  In the past days, and with the help of @Antoine38660 (<3 buddy) I worked on a proof of concept for a better translation system, designed to solve all these problems and make contributors lifes easier.
    
  Its concept is **clearly** inspired **a lot** by a PHP framework called Yii, which handles i18n in a clever way. [1]
  
  The POC is available here: https://github.com/machour/rn-i18n-poc

  (please keep in mind, I only did jQuery in the past and discovered `const`, `import`, `yarn`, `let`, `export` a week ago. Antoine did its best to enhance stuff, but I kept on breaking things.)
  
  Here are the main features of this poorly coded idea :
  
Coding your app
---------------
  
  * The application is developed using english sentences, surrounded by a function, and using placeholders to inject values :
  ```
  t("{user} made {repository} public", { 
  	user: "Houssein",
  	repository: "gitpoint/git-point"
  });
  ```
  
* Injected values can be strings, or ... React Components !
  ```
  t("{user} made {repository} public", { 
    user: "Houssein",
[...]

PR #747, opened 2018-03-22
chore(i18n): Migrate to js-lingui

machour:
| Question               | Response    |
| ----------------- | ----------- |
| Version?               | v1.4.1      |
| Bug fix?               |  no      |
| New feature?      | no      |
| Includes tests?   | no      |
| All Tests pass?   | no      |

## Description

This is an attempt to migrate from react-native-i18n to js-lingui.

### Why change?

I already addressed my concerns about our current translation process in #273. Main points were:

- **from the developer side, translate('www.xxx.yyyy.zzzz') is often cryptic, and you have to go check `src/locale/en.js` to understand what's the sentence is about.** 

Try reading the code of `events.screen.js` and mentally reconstructing sentences, you'll be in for a headache

- **from the translator side, coming up with an accurate translation is almost unfeasible without diving into the code to understand how the concatenations are made**

- **we are currently loosing track of our translations.** 

How many sentences do we have? Are they all used? How many translations are missing in `eo.js`?
Every PR adds more untranslated strings to all languages, and I bet it's a hassle for all contributors.

- **what work in english doesn't always work for other languages**: 

_machour_ opened TYPE "foobar" at "gitpoint/git-point"

currently TYPE is a variable than can be either "issue" or "pull request".
in French, "issue" is a masculine word (**le** ticket), and "pull request" feminine (**la** pull request).
but TYPE is also used in another context where it doesn't have a pronoun. So the french translation is screwed up.

### Why js-lingui

I first attempted to come up with my own lib, but after discovering the amazing work @tricoder42 (👋) have already accomplished, the decision to not re-invent the wheel was obvious to make.

A picture being worth a thousand words, consider this amazing before/after:
screen shot 2018-03-22 at 12 16 46 pm

To me, this _IS_ how i18n should be handled in an application:
* The developer write plain sentences in English in the code
* Translators are provided with meaningful sentences to translate

`js-lingui` comes with a lot of great features for us:

* a [I18nProvider](https://lingui.github.io/js-lingui/ref/lingui-react.html#component-I18nProvider) we wrap our App with.
* a `<Trans>` component to wrap our sentences in JsX (see it as a `<Text>` that will get translated)
* a [withI18n](https://lingui.github.io/js-lingui/ref/lingui-react.html#withI18n) HOC which injects an i18n prop to our components when need to translate a component attribute (usage: ``i18n.t`My sentence` ``) 
* an extraction utility (`yarn i18n:extract`) which extracts all messages into JSON files, and produces a nice/meaningful output:
```
Catalog statistics:
┌──────────┬─────────────┬─────────┐
│ Language  │ Total count  │ Missing  │
├──────────┼─────────────┼─────────┤
│ de        │     194      │    7     │
│ eo        │     194      │    9     │
│ es        │     194      │    8     │
│ eu        │     194      │    9     │
│ fr        │     194      │    7     │
│ gl        │     194      │    9     │
│ nl        │     194      │   10     │
│ pl        │     194      │    8     │
│ pt_br     │     194      │    8     │
│ pt        │     194      │    8     │
│ ru        │     194      │    7     │
│ tr        │     194      │   12     │
│ uk        │     194      │    7     │
│ zh_cn     │     194      │    7     │
│ zh_tw     │     194      │    7     │
└──────────┴─────────────┴─────────┘
```

The JSON files look like this:

```js
{
  "No organizations": {  // <= the original sentence
    "translation": "Aucune organisation", // <= the translations
    "origin": [ // <= where the original sentence was found
      ["src/auth/screens/auth-profile.screen.js", 167],
      ["src/user/screens/profile.screen.js", 223]
    ]
  }
  // ....
}
```

Icing on the cake: When a string become obsolete, it's kept in the JSON object with an additional `obsolete: true` field

* a compilation utility (`yarn i18n:compile`) that transforms the JSON files into minified JS files.

### PR Status

Disclaimer: I originally wanted to try js-lingui and ended up migrating **all the app** 😅. Since js-lingui and react-native-i18n can co-exist, I don't mind breaking this PR into smaller PRs easier to review. 

* There's just one problem left: I don't know how to deal with the last remaining translate() call located in route.js, for the Issue screen title. Tomáš being unavailable at the moment, I didn't bug him for support. This may be a good opportunity for us to separate Issue from PullRequest ? 🤔 

* I had to do a lot of refactoring in `event.screen.js` to tame the code and make it correctly translatable. (This definitely can be done in a small initial PR to introduce js-lingui)

* I ran into https://github.com/lingui/js-lingui/issues/171 so I'm using v2.0.2 instead of the latest v2.0.4

* Didn't adapt tests yet, so they will fail.

Now awaiting @gitpoint/maintainers feedback <3


App.js
```diff
@@ -122,9 +129,11 @@ class App extends Component {
 
     return (
       <Provider store={configureStore}>
-        <GitPoint onNavigationStateChange={this.statusBarHandler}>
-          <StatusBar />
-        </GitPoint>
+        <I18nLoader language={language} catalogs={catalogs}>
```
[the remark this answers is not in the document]
machour: import catalogs from './locale';

catalogs is an object with all translations for all languages:

{
  fr: {
    "Welcome to GitPoint": "Bienvenue à GitPoint",
    // ...
  }
}

src/auth/screens/login.screen.js
```diff
@@ -289,10 +295,13 @@ class Login extends Component {
                 type="octicon"
               />
               <Text style={styles.title}>
-                {translate('auth.login.notificationsTitle', locale)}
+                <Trans>Control notifications</Trans>
```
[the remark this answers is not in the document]
machour: `<Trans>` can be used in JSX, just like you would use `<Text>`:

```
<View>
  <Trans>This will be translated</Trans>
</View>
```

When you need to translate a component prop however, you can't use that as `<Button title={<Trans>foo</Trans>} />` wouldn't make much sense as title expects a string, not a component.

This is when i18n.t comes into play: ```<Button title={i18n.t`foo`} />```

machour:
@housseindjirdeh indeed, migrating languages was kind of a pain in the A. The script was able to migrate automatically ~ 80% of the strings, then I had to tweak it several times to automatically concat stuff with placeholders, this is why it got so ugly that I deleted it right after finishing 😄 

I would really love if you could pull down this PR and give it a go locally. If you like what you see, then please help out with that last translation that's killing me!

housseindjirdeh:
@machour where you seeing this by any chance including js-lingui to the project? 

simulator screen shot - iphone 6 - 2018-03-26 at 20 08 56

Not sure if I've missed anything pulling it down, but clearing my cache and restarting everything doesn't seem to fix this 🤔

machour:
Seems like I was always authenticated when working on this PR 😄
Fixed with last commit

src/issue/screens/issue.screen.js
```diff
@@ -77,7 +77,7 @@ const compareCreatedAt = (a, b) => {
 };
 
 class Issue extends Component {
-  static navigationOptions = ({ navigation }) => {
+  static navigationOptions = ({ navigation, i18n }) => {
```
housseindjirdeh: We can fix the probelm of that single `translate` in the routes file by doing it here. For example:

```
  static navigationOptions = ({ navigation, i18n }) => {
    const { state, navigate } = navigation;
    const { issue, issueURL, isPR } = state.params;

    const issueNumberRegex = /issues\/([0-9]+)(#|$)/;
    const number = issue ? issue.number : issueURL.match(issueNumberRegex)[1];
    const issueType = isPR ? i18n.t`Pull Request` : i18n.t`Issue`;

    const getHeaderIcon = () => {
      // ...
    };

    return { title: `${issueType} #${number}`, headerRight: getHeaderIcon() };
  };
```

With that being said, for some interesting reason nothing defined in the navigationOptions _here_ are rendering in this PR. For example, I don't see the cog on the right hand side where I should 🤔 

simulator screen shot - iphone 6 - 2018-03-30 at 18 09 35

Another thing I wanted to mention (and completely outside this PR and my fault 😅) is I think we need to move all screen specific route configurations to `static navigationOptions` in their component files and not in `routes.js` Things are a bit unwieldy there so I'll create a refactoring ticket where I can get to that sometime soon.

App.js
```diff
@@ -63,8 +68,8 @@ class App extends Component {
         transforms: [encryptor],
         whitelist: ['auth'],
       },
-      () => {
-        this.setState({ rehydrated: true });
+      (err, { auth }) => {
+        this.setState({ rehydrated: true, language: auth.locale });
```
housseindjirdeh: This fails when we're not authenticated and loading the app from scratch:

simulator screen shot - iphone 6 - 2018-03-30 at 17 35 31

Maybe we can default the locale if the store is empty like `this.setState({ rehydrated: true, language: auth.locale || 'en' });`?

src/auth/screens/language-setting.screen.js
```diff
@@ -7,9 +7,11 @@ import { ListItem } from 'react-native-elements';
 import { colors, fonts } from 'config';
 import { changeLocale } from 'auth';
```
housseindjirdeh: Currently, changing the language in the language settings screen doesn't update the language so we'll just need to map it to our new library.

housseindjirdeh:
Playing around with the app (and besides the few things I noticed and left comments on) - things are looking beautiful.

Let me know if you would rather have us not merge anything in until this gets resolved if you want to not deal with merge conflicts 😅 This already took a lot of time for you and don't want to make things harder by merging PRs in that may conflict mate.

src/issue/screens/issue.screen.js
machour: Just noticed the navigationOptions issue. Definitely with you on use static navigationOptions()!

App.js
machour: Definitely :D

machour:
Frankly, the hardest part was converting catalogs from one format to another, which is done and will live on in this PR even if it's discarded.

As you can see, there are still a few problems, and probably some uncovered ones. Working on js-lingui was a great opportunity to see our current i18n flaws, and that they could be solved.

It also made me realize that we're currently not using react-native-i18n as best as we could be doing it:
 * We can use plain English strings as keys
 * We could merge {...en, ...fr} at runtime rather than forcing contributors to introduce untranslated strings in all languages for every PR.

This way, switching to js-lingui would be easier in the future, and we will do that gradually to avoid a big clumsy PR.

Since we started the Redux refactoring, I'd rather close this PR and focus all our efforts on the ongoing migration, instead of potentially introducing more problems.

Issue #732, opened 2018-02-15
new auth mothods


ShalokShalom:
See this whole thread and especially that linked convo:

https://github.com/gitpoint/git-point/issues/674#issuecomment-365206090

housseindjirdeh:
Thanks so much for logging this :)

Quite a few folks are experiencing this - definitely something we need to prioritize for our next release

ShalokShalom:
And it happens in all the Github clients I tried yet. 
So all which are available for 4.3
Here is another discussion ongoing:
https://github.com/slapperwan/gh4a/issues/818
Thanks a lot for addressing this as high priority. :)

ShalokShalom:
Ok, same news: Its in the auth app itself.

This issue happens to me on all different implementations.
And the auth on pages who offer both, Twitter and Github, works for Twitter and not for Github.

ShalokShalom:
Octodroid added password/username login back: https://github.com/slapperwan/gh4a/commit/270e952c968be1b2daf5e526b3e15110b4a2f8a9

ShalokShalom:
Is this also a solution for this project?

ShalokShalom:
Is there any news on this high priority issue? :slightly_smiling_face:

machour:
Hi @ShalokShalom, unfortunately all maintainers have been quiet busy lately.
Do you feel like tackling this one?

ShalokShalom:
I am currently unable to do any JavaScript

machour:
@ShalokShalom could you remove the app and install https://idk.tn/gitpoint-test.apk ?

After clicking "Sign in" you should see a really ugly "Can't sign in" button that will open the GitHub oAuth URL in your browser. Tell me if that works for you 🙏

ShalokShalom:
Nope. Same out-greyed button.

machour:
In your Android browser this time? :/

ShalokShalom:
In my Blackberry browser, which is pretty outdated.
It works in Firefox :D

ShalokShalom:
Sure sure 

Thanks for dealing with me :smiley:

Issue #680, opened 2017-12-27
Stuck on authorization screen


HaLeiVi:
I'm using a Samsung Galaxy S3 running Android 4.4.2. I am able to enter login info. Then I'm taken to the authorizing screen, but it stays there. The authorize button and even 'cancel' is unresponsive.

housseindjirdeh:
Thanks for logging this @HaLeiVi.

Someone just recently showed me this problem on their Android device as well, will look into it soon and find out what's going on.

machour:
Javascript wasn't enabled on this Webview, and the GitHub authorization screen [seems to need it](https://stackoverflow.com/questions/44425233/why-is-the-github-authorization-button-in-the-oauth-flow-grayed-out). Opened #681 to fix it.

However, weirdly enough the auth process always worked on my OnePlus 3T device, without enabling JS. 🤔

housseindjirdeh:
Thanks a million @machour 

@HaLeiVi if you don't mind, I'll generate an APK for you to test on your device if that's okay? You'll get the app working for you sooner that way and we can confirm that this indeed fixes it (especially since I can't reproduce this on any android device)

HaLeiVi:
Sounds great!

housseindjirdeh:
So sorry for the delay @HaLeiVi, but here's the apk with the fix: https://github.com/gitpoint/git-point/releases/tag/hotfix

Give it a shot whenever and please let me know if all is well! We should have a release with this fix in about a week.

linrz:
![screenshot_2018-01-30-18-42-41-989_com gitpoint](https://user-images.githubusercontent.com/33575974/35562149-1b9188d6-0578-11e8-8d36-eae5ec420fef.png)
I download the latest apk.It seems still can not work.

machour:
@linrz hey, did you download the APK from the provided link to debug this issue?
    https://github.com/gitpoint/git-point/releases/tag/hotfix

linrz:
@machour Yes, I download this pre-release APK  which released 12days ago. My android's version is 6.0.1. It seems this webview use IE caused it.
![wechatimg116](https://user-images.githubusercontent.com/33575974/35571164-4a490f16-060c-11e8-9b6f-4d7a8abec435.jpeg)

HaLeiVi:
I got the same results. I was able to enter password, and verify it. But
couldn't go past that screen.

machour:
Ok Houston, we definitely have a problem 😄 
[...]

Issue #674, opened 2017-12-20
Github's Authorize Button Disabled in Android WebView


pallavbakshi:
| Question         | Response    |
| ---------------- | ----------- |
| Version?         | v1.4.1    |
| Devices tested?  | Xiaomi Mi 4 - Android 5.0|

---

## What Currently Happens?
Upon logging in, the inbuilt browser showcases a Github auth page with the green button (Authorize) disabled.

This problem is encountered only when the sign-in flow is interrupted due to some user circumstances. For example, if you download the app for the first time and click sign-in, it will take you to the Github Auth page where you can enter your details to allow git-point to access your profile/repo. Once you successfully sign in, Github asks you to allow the app the required permissions. At this point, the green button is working (if this is a fresh sign-in attempt) and the cancel button is disabled so the user cannot break the flow. However, due to external circumstances, like a phone call or if the user closes the app or anything else, the sign-in process in broken in between. User has already signed in to the Github web page but hasn't authorized the GitPoint app yet. Now when the user reopens the app and tries to sign in, he is directly taken to the authorization page (permission page where he left last time). He is asked to allow the app to access its profile, etc. And at this time the green button (authorize button) is disabled. It shall be noted that scrolling and normal tapping is still enabled. ONLY the authorize button is disabled.

## What Do You Expect To Happen?
Green button to be enabled.

## My Reproduction Steps
Mentioned above.

1. Try sign-in for the first time
2. Enter credentials. Don't approve the app though.
3. Close the app (don't authorize the app yet)
4. Reopen the app and try to sign-in

My personal view of what is happening -
Since the android WebView saves the cache/cookies therefore the app picks up from the last point where the user left the signin process on the github page (that is why user doesn't need to sign in). However, the javascript seems to be disabled which disables the green button. I have only checked it in Android.

machour:
I think this is the same as #680 
Leaving this open for the wonderful details, thank you @pallavbakshi !

pallavbakshi:
Ha ha! I came first though (Mine issue is 674). Fun aside, I will test it and close the issue if it is fixed. However, it will take me some time to test it. Thanks :D

machour:
hahahaha, you just reminded me when I told my grandpa older brother: "oh you look like grandpa" and he shouted at me "no, HE looks like ME" 🤣 

I think @housseindjirdeh will be providing a test APK shortly for #680, I'll make sure to notify you so that you can test as well. Thank you soo much!

[...]

PR #749, opened 2018-03-24
fix(login): Provide an escape hatch for Android

machour:
| Devices tested?  | OnePlus 5t |
| Related ticket?  | #732 #674 #680    |

---

## Screenshots

| Before   | After    |
| -------- | -------- |
| screenshot_20180324-125009 | screenshot_20180324-124806 |

## Description

Some users complained about the "Login" button being disabled in the WebView. 
This is because GitHub disables it from some old/stock android browsers.

This PR simply adds a link to open the oAuth screen in an external browser, so that the user can decide which browser to use and finalize the login step. Tested and confirmed in #732.

### DIFF
--- a/src/auth/screens/login.screen.js
+++ b/src/auth/screens/login.screen.js
@@ -22,7 +22,7 @@ import { ViewContainer, ErrorScreen } from 'components';
 import { colors, fonts, normalize } from 'config';
 import { CLIENT_ID } from 'api';
 import { auth, getUser } from 'auth';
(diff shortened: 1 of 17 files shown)


ShalokShalom:
I guess its easy to overlook that, did you consider to detect the version of the runtime and redirect directly?

chinesedfan:
@ShalokShalom I think the scope of versions is not easy to be determined. But if you insist, feel free to create a new issue to remind it.

Issue #751, opened 2018-03-26
Rest: Migrate paginated calls to the new API


machour:
This issue is here to help keeping track of the first step of our big redux refactoring: migrating paginated calls.

- [x] Events: list (#736)
- [ ] Issue: comments
- [ ] Organization: members
- [ ] Organization: repositories
- [ ] Repository: issues list
- [ ] Repository: pull requests list
- [ ] Repository: collaborators
- [ ] Search: repositories
- [ ] Search: users
- [ ] User: following list
- [ ] User: followers list
- [ ] User: repositories list
- [x] User: starred repositories list (#750)

Feel free to assign yourself for any of these.

PR #750 contains a walk-through for the required changes

chinesedfan:
Just remind: if not migrating `Repository: issues list`, it will crash when no result is found.

PR #752, opened 2018-03-27
refactor: switch organization screen to new Redux api

machour:
| Devices tested?  | iPhone 6 |
| Related ticket?  | #751    |

---

## Description

Switched the Organization screen to the new API, the members list is now paged 🎉 (check the Facebook org for example)

Now we're already familiar with the paginated stuff. The implementation in this PR is similar to the previous ones.

This PR introduces "single entity" retrieval using the new Redux API, in order to fetch the organization information.  This process is simpler than the pagination, as we're not dealing with nextUrl, or anything like that.

PS: Sorry for the 3 first unrelated commits, I managed to mess my fork 😓


src/organization/screens/organization-profile.screen.js
```diff
@@ -199,6 +184,38 @@ class OrganizationProfile extends Component {
...
-);
+const mapStateToProps = (state, ownProps) => {
+  const {
+    auth: { user, locale },
+    pagination: { ORGS_GET_MEMBERS },
+    entities: { orgs, users },
+  } = state;
+
+  const orgId = ownProps.navigation.state.params.organization.login;
+  const org = orgs[orgId] || ownProps.navigation.state.params.organization;
```
machour: **Something important to have in mind:**

This screen is accessed from a User profile, or from the Search results. 

In both cases, `ownProps.navigation.state.params.organization`  contains an incomplete Organization object (left), which is not enough to render the screen, as we need the number of repositories, etc, which is only present after calling the organization endpoint to get the complete object (right):

| initial object | after fetching information |
| ------------ | -------------------------- |
| screen shot 2018-03-27 at 11 49 31 pm | screen shot 2018-03-27 at 11 51 22 pm |

This is why, _with or without this PR_, if you navigate to a user profile, turn off internet, and then click on one of his organizations, you will see a broken incomplete UI.  Usually the GitHub API call is so quick that it goes unnoticed.

---

Normally, `org` would only be read from `state.orgs`, which would already contain the incomplete organization object. Since the User profile & Search screens doesn't use the new API yet, hence, doesn't populate `state.orgs` with the incomplete object, I'm retrieving it from the navigation state.

Reading from navigation will be removed in a future PR, once the new Redux API is used everywhere. Only the organization **id** will be passed in the navigation object, and the Redux API will go get the object for us.

---

If we use this same approach for all our screens (receiving only IDs, no objects, and letting the redux API fetch things), GitPoint will be able to handle all Github.com urls with deep linking 🎉

src/organization/screens/organization-profile.screen.js
```diff
@@ -82,22 +52,24 @@ class OrganizationProfile extends Component {
...
 
   componentDidMount() {
-    const organization = this.props.navigation.state.params.organization;
+    const { org, orgId, getOrgById, getOrgMembers } = this.props;
 
-    this.props.fetchOrganizations(organization.login);
-    this.props.fetchOrganizationMembers(organization.login);
+    if (!org.name) {
+      getOrgById(orgId);
+    }
```
machour: This if does a little caching, to avoid refetching from the REST api if we already have the complete organization object (see comment below).

This is needed for now as single object retrieval is not cached in the new Redux API, and will be automatically handled soon.

src/organization/screens/organization-profile.screen.js
machour: I wouldn't have done that differently as a first approach, but working with this new Redux API & having in mind deep linking, it's now obvious that every screen should be "standalone" and be able to fetch it's data with simple ids.

A pull screen for example would require a repository identifier & the pull number, just like this Pull Request url on GitHub web:
    `https://github.com/**gitpoint/git-point**/pull/**752**`

nahmean?

PR #755, opened 2018-03-29
fix: add a Proxy polyfill to make sure it's available everywhere

machour:
## Description

@MrZhang123 reported on gitter that he wasn't able to run the project anymore on Genymotion, due to the lack of ES6 Proxies in this environment.

This PR adds a polyfill. Tested and validated on a Genymotion 6.0.0 (api 23) player.

I've did a lot of googling about whether or not RN supports proxies in all targets for the Release build, and after reading a bunch of stuff, all I'm certain of is that I'm really uncertain. I would be really thankful if anyone could provide a clear explanation about this. 🙏

### DIFF
--- a/package.json
+++ b/package.json
@@ -64,6 +64,7 @@
     "normalizr": "^3.2.4",
     "opencollective": "^1.0.3",
     "parse-diff": "^0.4.0",
+    "proxy-polyfill": "0.1.6",
     "query-string": "^5.0.1",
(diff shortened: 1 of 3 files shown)


machour:
Wow, just tested on my OnePlus 5T for the first time and a while and got the same error 😨 

I'm going to go ahead and merge this one and go star the `proxy-polyfill` repository a thousand times..

PR #758, opened 2018-03-30
 refactor: migrate search screen to the new API

machour:
| Related ticket?  | #751         |

## Description

Refactored the search screen to make use of the new API. Search results are now paginated & cached!


src/api/client.js
```diff
@@ -149,4 +149,23 @@ export class Client {
...
     },
   };
+
+  search = {
+    repos: async (q, params) => {
+      return this.call(`search/repositories?q=${q}`, params).then(response => ({
+        response,
+        nextPageUrl: this.getNextPageUrl(response),
+        schema: Schemas.REPO_ARRAY,
+        normalizrKey: 'items',
```
machour: This is the first use of `normalizrKey`.

The API response for this call looks like the following:

```
{
  total_count: 645354,
  incomplete_results: false,
  items: [
    { /* first repo object */},
    { /* second repo object */},
    ...
  ]
}
```

We only care for what's inside items, the array of repositories.

`normalizrKey` tells the dispatch proxy to [only treat that property](https://github.com/gitpoint/git-point/blob/master/src/api/proxies/create-dispatch-proxy.js#L105) of the JSON, instead of all the JSON.

src/search/screens/search.screen.js
```diff
@@ -15,25 +20,54 @@ import {
...
-      searchUsers,
-    },
-    dispatch
+import { RestClient } from 'api';
+
+const NAV_QUERY_PARAM = 'q';
+const SearchTypes = {
+  REPOS: 0,
+  USERS: 1,
+};
```
machour: introduced this little enum to make the code more readable

src/search/screens/search.screen.js
```diff
@@ -15,25 +20,54 @@ import {
...
+    auth: { locale },
+    pagination: { SEARCH_REPOS, SEARCH_USERS },
+    entities: { repos, users },
+  } = state;
+
+  const searchQuery = ownProps.navigation.getParam(NAV_QUERY_PARAM);
+
+  const reposSearchResultsPagination = SEARCH_REPOS[searchQuery] || {
+    ids: [],
+    isFetching: false,
```
machour: contrarily to other PRs, when we're loading this screen, we're not fetching yet, so we set `isFetching` to false.

src/search/screens/search.screen.js
```diff
@@ -127,19 +161,22 @@ class Search extends Component {
     const selectedSearchType =
       selectedType !== null ? selectedType : this.state.searchType;
 
-    if (query !== '') {
+    if (query.trim() !== '') {
```
machour: Added a trim() here. Currently, if you search for "    ", the app will crash!

src/search/screens/search.screen.js
```diff
@@ -127,19 +161,22 @@ class Search extends Component {
...
+      const searchedQuery = query.toLowerCase();
```
machour: GitHub search is case insensitive. We lowercase the query string in order to avoid having two paginations for nothing (Gitpoint vs gitpoint)

src/search/screens/search.screen.js
```diff
@@ -127,19 +161,22 @@ class Search extends Component {
...
           ...this.state.currentQuery,
-          [selectedSearchType]: query,
+          [selectedSearchType]: searchedQuery,
         },
-        query,
+        query: searchedQuery,
       });
-      if (selectedSearchType === 0) {
-        searchRepos(query);
+      this.props.navigation.setParams({ [NAV_QUERY_PARAM]: searchedQuery });
```
machour: We save the query string in the navigation, to be able to get it in MapStateToProps and select the corresponding pagination

src/search/screens/search.screen.js
```diff
@@ -172,26 +209,58 @@ class Search extends Component {
...
+
+    if (
+      this.props[
+        searchType === SearchTypes.REPOS
+          ? 'reposSearchResultsPagination'
+          : 'usersSearchResultsPagination'
+      ].nextPageUrl === null
+    ) {
+      return null;
+    }
```
machour: Here, we conditionally select the good pagination. 

I can brake the code to have something less barbaric. This will change anyways when we introduce Code & Wikis search

src/search/screens/search.screen.js
```diff
@@ -249,7 +322,30 @@ class Search extends Component {
...
+                  })
+                }
+                refreshing={isPendingSearchRepos || isPendingSearchUsers}
+                onEndReached={() =>
+                  this.props[
+                    searchType === SearchTypes.REPOS
+                      ? 'searchRepos'
+                      : 'searchUsers'
+                  ](query, { loadMore: true })
+                }
```
machour: same remark as above.

root.reducer.js
```diff
@@ -4,7 +4,6 @@ import { userReducer } from 'user';
 import { repositoryReducer } from 'repository';
 import { organizationReducer } from 'organization';
 import { issueReducer } from 'issue';
-import { searchReducer } from 'search';
```
machour: pew pew pew 🔫

src/api/client.js
housseindjirdeh: NICE ⭐️

src/search/screens/search.screen.js
housseindjirdeh: Just wondering, how are you thinking we can simplify this when code & wikis are introduced?

Also, in the meantime, maybe we can simplify the following to something like:

```
renderFooter = () => {
    const { searchType } = this.state;
    const { reposSearchResultsPagination, usersSearchResultsPagination } = this.props;

    const currResultsPagination = searchType === SearchTypes.REPOS ? reposSearchResultsPagination : usersSearchResultsPagination;

    if (currResultsPagination.nextPageUrl) {
      return (
        <View
          style={{
            paddingVertical: 20,
          }}
        >
          <ActivityIndicator animating size="large" />
        </View>
      );
    }
  };
```

And another thought (if not know) - maybe we can extract this to a separate component (`<LoadingFooter />`?) if this logic is duplicated exactly the same in multiple screens?

machour: I was thinking about having `getActiveResultsPagination()` & `getActiveResults()` instead of using ifs everywhere.

And for the refactoring, I was thinking of:
* a `<PaginatedFlatList>` which would take a `pagination` prop and do everything for us. 
* a `getPagination()` in utils to avoid this chunk of code in mapStateToProps:

```
  const reposSearchResultsPagination = SEARCH_REPOS[searchQuery] || {
    ids: [],
    isFetching: false,
 }
```

was just holding off on changing stuff until I got a few PRs merged. I think I'll do that next

machour:
@housseindjirdeh refactoring done, now our render() method doesn't check for the search type anymore (aside for a translation) 🙌

Adding a new search type should be as simple as declaring it in the enum and the switch/case in the new methods.

I also spotted and fixed a minor UI problem, where the Loading list footer was displayed during the first search, resulting in two Loaders being displayed.

machour:
By the way, noticed a bad UX design. Having a single FlatList to handle the various search type results may not be a good idea: 

If you search for "react", display results for Repositories and Users, then scroll down on Repositories, when you get back to the Users tab, the list will have been scrolled.

Something to address in another PR.

housseindjirdeh:
And aaah I've never caught that before :( That really isn't great behaviour - opened #764 for this

PR #765, opened 2018-04-02
feat: Add LoadingEventListItem

machour:
| Devices tested?  | iPhone 7, OnePlus5T |
| Related ticket?  | none        |

---

## Screenshots

| Before   | After    |
| -------- | -------- |
| simulator screen shot - iphone 6 - 2018-04-02 at 13 05 11|simulator screen shot - iphone 6 - 2018-04-02 at 12 57 10|

## Description

Created a new `<LoadingEventListItem />` component used as a placeholder in the events screen.

Prior to the Redux refactoring, the events screen was using `<LoadingUserListItem />` which was missing the icon placeholder. During the refactoring, this got lost due to a silly mistake on my side.

Also, when reaching the list bottom, instead of displaying the boring `<ActivityIndicator />`, we're now using this new loader, which matches the expected content :

screen shot 2018-04-02 at 12 52 06 pm

---

While setting this new component up, I noticed that after a page or two, it wasn't animating anymore. This is due to the current implementation of `loadingAnimation` that only animated 10 times, then stopped.

Since we'll be seeing this component a lot of time in the list (every time we hit bottom), we needed a truly infinite animation. This is implemented by `infiniteAnimation`, which takes a callback so that we can start the animation again when it ends.

---

@chinesedfan since you're our styled-component expert, I'd really love if you could give `<LoadingEventListItem />` a deeper look. I wasn't able to handle the `opacity` prop using SC, so I'm setting it from the `style` prop.

Once we agree on how this should be done, I think we can use this new component as an example to migrate other loading indicators (adding them to #532 now)

### DIFF
--- a/src/auth/screens/events.screen.js
+++ b/src/auth/screens/events.screen.js
@@ -3,9 +3,9 @@
 import React, { Component } from 'react';
 import styled from 'styled-components';
 import { connect } from 'react-redux';
-import { Text, FlatList, View, ActivityIndicator } from 'react-native';
+import { Text, FlatList, View } from 'react-native';
 
-import { LoadingUserListItem, UserListItem, ViewContainer } from 'components';
+import { LoadingEventListItem, UserListItem, ViewContainer } from 'components';
 import { colors, fonts, normalize } from 'config';
 import { emojifyText, translate, relativeTimeToNow } from 'utils';
 import { getNotificationsCount } from 'notifications';
@@ -508,15 +508,7 @@ class Events extends Component {
       return null;
     }
 
-    return (
-      <View
-        style={{
-          paddingVertical: 20,
-        }}
-      >
-        <ActivityIndicator animating size="large" />
-      </View>
-    );
+    return <LoadingEventListItem />;
   };
 
   render() {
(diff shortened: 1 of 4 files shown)


src/auth/screens/events.screen.js
```diff
@@ -530,10 +522,10 @@ class Events extends Component {
     const linebreaksPattern = /(\r\n|\n|\r)/gm;
     let content;
 
-    if (userEventsPagination.isFetching && !userEvents) {
+    if (userEventsPagination.isFetching && userEvents.length === 0) {
```
machour: the silly mistake I was talking about

src/components/loading-indicators/loading-event-list-item.component.js
```diff
@@ -0,0 +1,79 @@
+import React, { Component } from 'react';
+import { Animated } from 'react-native';
+import styled from 'styled-components';
+import { colors } from 'config';
+import { infiniteAnimation } from 'utils';
+
+const Container = styled.View`
+  padding: 10px 10px 10px 0;
```
housseindjirdeh: can also be simplified to `padding: 10px 10px 0;`

src/auth/screens/events.screen.js
housseindjirdeh: Easy mistake to miss 😅

Issue #761, opened 2018-03-31
Redux refactoring steps


machour:
Sharing here what I have in mind for the big Redux refactoring.

### Step 1 : Pagination

Migrate all the lists fetched from the API to use pagination
This is the currently ongoing step and is tracked in #751 

### Step 2 : Single entity retrieval

Migrate all the single entities retrieval to the new API.
This have been done for organizations in #752 for example, and will be tracked once #751 is closed.

### Step 4 : Counts

Migrate counting operations to the new API.

For example: Getting the number of notifications.

### Step 5  : PUT / PATCH / DELETE

Migrate these operations on single entities to the new API.
Example: Create/Update/Delete comment

### Step 6 : Source tree cleaning

When step 5 is closed, all our directories will be looking like below since we will have removed all *.actions.js *.reducer.js *.types.js files:

```
src/organization/
├── index.js
└── screens
    ├── index.js
    └── organization-profile.screen.js
``` 

A PR will be made to get rid of the `screens` folder:

```
src/organization/
├── index.js
└── organization-profile.screen.js
``` 

### Step 7 : Nested schemas in Normalizr & attributes clean up

For now, every Normalizr schema is independent. So a "repository" looks like:

```
{
  "id": 86202845,
  "name": "git-point",
  "full_name": "gitpoint/git-point",
  "owner": {
    "login": "gitpoint",
    "id": 30082377,
    "avatar_url": "https://avatars0.githubusercontent.com/u/30082377?v=4",
    "gravatar_id": "",
    "url": "https://api.github.com/users/gitpoint",
    "html_url": "https://github.com/gitpoint",
    "followers_url": "https://api.github.com/users/gitpoint/followers",
    "following_url": "https://api.github.com/users/gitpoint/following{/other_user}",
    "gists_url": "https://api.github.com/users/gitpoint/gists{/gist_id}",
    "starred_url": "https://api.github.com/users/gitpoint/starred{/owner}{/repo}",
    "subscriptions_url": "https://api.github.com/users/gitpoint/subscriptions",
    "organizations_url": "https://api.github.com/users/gitpoint/orgs",
    "repos_url": "https://api.github.com/users/gitpoint/repos",
    "events_url": "https://api.github.com/users/gitpoint/events{/privacy}",
    "received_events_url": "https://api.github.com/users/gitpoint/received_events",
    "type": "Organization",
    "site_admin": false
  },
  "private": false,
  "html_url": "https://github.com/gitpoint/git-point",
[...]

PR #770, opened 2018-04-12
[dont merge] - Refactor proxy client & migrate issue screen

machour:
## Description

Hi team,

I've started looking at Step 4 of #761 (handle put/patch/delete) and experimented on the Issue screen.

While doing that, I came to the conclusion that the Proxy <=> Client interactions needed to be changed a bit, in order to give the proxy more power over all kinds of operations, and how the resulting data is reduced in store.

As a result, the client/proxy code is way clearer than before 🎉

You can follow the train of thought that lead to this PR in this gist, and I really encourage you to do so in order to completely understand this Proxy thing: https://gist.github.com/machour/698e91e075ac35ad64531d03c80a99bf

If you pull this PR and test it locally, you will see the following stuff:
  * We retrieve the comments for an issue
  * If we add a comment, a `POST` call is triggered, and its result is automatically appended to the list of comments (without refetching)
  * If we edit a comment, a `PATCH` call is triggered, and the changed fields are automatically updated in store (without refetching)
  * If we delete a comment, a `DELETE` call is triggered, and the comment reference is removed from the comments list in store (again, no refetching)
  * If you lock/unlock the issue, a `PUT` call is triggered, and the Issue object we already have in store is patched so that the `locked` property is up to date (instead of re-fetching the whole issue).

Same thing goes for issue labels and assignees.

To say thing differently: *Our Store is 100% up to date with GitHub, by doing precisely what needs to be done. Nothing more. Nothing less.*

---

This PR won't be merged as a whole. 
Instead, I'll be opening a smaller PR with the subset of changes needed for the refactoring so that we can confirm that everything still work without touching the migrated screens code.

Once this first PR is merged, I'll open a new one to unleash the PUT/DELETE/PATCH/POST features.

---

I've already did a 1 on 1 with @housseindjirdeh to walk him through this PR, and I'll be happy to do this with any of you girls and guys!


machour:
Closing this demo.

PR #775, opened 2018-04-17
refactor(repository-screen): switch to GraphQL

machour:
| All Tests pass?  | no      |
| Related ticket?  | #761         |

---

## Description

Introducing GraphQL in the new Redux API 🎉 

Rewrote the repo screen to be loaded with only two requests:
* A GraphQL query to load all the displayed information on the screen
* A REST list query to load the contributors (not provided by GitHub v4 yet)

I also ported star/unstar watch/unwatch & fork to the new API.

For now, the GraphQL query also returns 50 issues & 50 pull requests, to avoid making any changes to the issues/pulls list screens. We will get rid of that once these two screens are rewritten to the new API.

Even with that little performance annoyance, the new screen is blazing fast!

## UX changes

* Starring/Unstarring or Watching/Unwatching are immediately reflected on the UI :

| watching | unwatching | not watching  |
| -- | -- | -- |
|simulator screen shot - iphone 7 - 2018-04-17 at 02 58 52|simulator screen shot - iphone 7 - 2018-04-17 at 02 58 56|simulator screen shot - iphone 7 - 2018-04-17 at 02 58 58|

This is possible with the new Client.delete() and Client.put() methods, that allow us to patch the entity in our Store. This logic already showcased for REST in #770, is also compatible with GraphQL.

* Error Page (to be discussed)

When a repository doesn't exist, we currently display a Toast with an error message, and an incomplete Repository screen.

I wanted to simplify things, a create a nicer user experience, so I tweaked this part a bit.

My first approach was to immediately navigate back & display the toast, but the toast length was really limited.

So for now I'm displaying this error page:

|  |
|  -- |
|simulator screen shot - iphone 7 - 2018-04-17 at 02 48 29|

Come to think about it, we already have an ErrorPage component, but with a hardcoded message for now. We could definitely refactor it a bit:
- Allow for a custom message to be passed as a prop
- Give it some UI ❤️ 
- Allow the user to provide us some feedback about how he got to the error (if it's a GitPoint app error) and have that feedback posted as a new issue in the GitPoint repository ✨ 

And icing on the cake, we would use that from the new `componentDidCatch` React lifecycle method and spread that in all our screens.

Any opinions on that ? If accepted, this would have to be done in a separate PR.

NB: For now the new API displays an additional (ugly) alert message meant for debugging.

### DIFF
--- a/src/api/client.js
+++ b/src/api/client.js
@@ -20,6 +20,12 @@ type FetchParameters = {
   body?: Object,
 };
 
+type UpdateEntity = {
+  type: string,
+  id: string,
+  updater: Function,
+};
+
 export type CallParameters = {
   endpoint: string,
   schema: Schema,
@@ -28,6 +34,7 @@ export type CallParameters = {
   normalizrKey?: string,
   paginationArgs?: Array<string | number | boolean>,
   entityId?: String | number,
+  updateEntity?: UpdateEntity,
 };
 
 export type CallType = CallParameters & {
@@ -99,12 +106,45 @@ export class Client {
     return fetch(finalUrl, parameters);
   };
 
+  put = ({ fetchParameters, ...config }: CallParameters): CallType => ({
+    type: 'put',
+    params: {},
+    ...config,
+    fetchParameters: merge(
+      {
+        method: this.Method.PUT,
+        headers: { Accept: this.Accept.JSON, 'Content-Length': 0 },
+      },
+      fetchParameters
+    ),
+  });
+
   get = ({ fetchParameters, ...config }: CallParameters): CallType => ({
     type: 'get',
     params: {},
     ...config,
     fetchParameters: merge(
       { method: this.Method.GET, headers: { Accept: this.Accept.JSON } },
+      fetchParameters
+    ),
+  });
+
+  delete = ({ fetchParameters, ...config }: CallParameters): CallType => ({
+    type: 'delete',
+    params: {},
+    ...config,
+    fetchParameters: merge(
+      { method: this.Method.DELETE, headers: { Accept: this.Accept.JSON } },
+      fetchParameters
+    ),
+  });
+
+  create = ({ fetchParameters, ...config }: CallParameters): CallType => ({
+    type: 'create',
+    params: {},
+    ...config,
+    fetchParameters: merge(
+      { method: this.Method.POST, headers: { Accept: this.Accept.JSON } },
       fetchParameters
     ),
   });
@@ -198,10 +238,11 @@ export class Client {
   };
 
   graphql = {
-    getRepo: (repoId: string) => {
+    getRepo: (repoId: string, params: SpecialParameters = {}) => {
       const [owner, name] = repoId.split('/');
 
       return this.query({
+        params,
         query: repoQuery,
         variables: { owner, name },
         schema: Schemas.GQL_REPO,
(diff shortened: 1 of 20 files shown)


src/api/client.js
```diff
@@ -200,6 +277,68 @@ export class Client {
...
+            subscribed: true,
+          },
+        },
+        schema: Schemas.GQL_REPO,
+        updateEntity: {
+          type: 'gqlRepos',
+          id: repoId,
+          updater: gqlRepo => ({
+            isSubscribed: true,
+            watchersCount: gqlRepo.watchersCount + 1,
```
housseindjirdeh: Good stuff :)

src/api/proxies/create-dispatch-proxy.js
```diff
@@ -103,14 +103,82 @@ export const createDispatchProxy = (Provider: Client) => {
...
+                    entities: {
+                      [type]: {
+                        [id]: updater(entity),
+                      },
+                    },
+                  });
+                }
+
+                return Promise.resolve();
+              }
```
housseindjirdeh: No worries about doing it now, we can chat about this much later - but because this single method is getting so long we can spend a little time separating these to separate methods

src/api/queries/repo.query.js
```diff
@@ -0,0 +1,122 @@
...
+    pullRequests: pullRequests(first: 50, orderBy: {field: CREATED_AT, direction: DESC}) {
+      totalCount
+      edges {
+        ...PullRequestsList
+      }
+    }
+  }
+}
+
+`;
```
housseindjirdeh: 👏 👏 👏 👏

src/api/reducers/entities.js
```diff
@@ -7,6 +7,7 @@ export const entities = (
     orgs: {},
     repos: {},
     users: {},
+    gqlRepos: {},
```
housseindjirdeh: Why are we separating it into it's own entity instead of just using `repos`? I'm assuming this is because other parts of the app still reference the original `repos` object (and if that's the case -> not a big deal for now)

src/repository/repository.action.js
```diff
@@ -152,221 +136,6 @@ export const getIssues = url => {
...
-      })
-      .catch(error => {
-        dispatch({
-          type: FORK_REPO_STATUS.ERROR,
-          payload: error,
-        });
-      });
-  };
-};
-
```
housseindjirdeh: Love it <3

src/api/client.js
machour: this was almost already in #770, but changed it to take a callback with the modified elements, for way more flexibility.

src/api/reducers/entities.js
machour: `repos` here is from our new store architecture. It holds v3 repositories found in the Search screen, or starred by a user for example.

I needed to separate repositories retrieved via GraphQL , from the one retrieved from the Rest API, because they don't share the same structure / field names.

---

If you get back to #761, you'll see that I've mentioned adapting GraphQL entities structure to match the ones from Rest. 

But now that I've experimented a bit with GraphQL, I think that we should be doing exactly the opposite: Adapt REST entities to the GraphQL structure in our store. We will do that once we migrated everything to the new API, because it will require updates in most of our components (avatar_url => avatarUrl).

---

Furthermore, since we're retrieving **exactly** what we need from GraphQL, I think that we should be storing the response structure as-is in the store. I'll be updating this PR accordingly to reflect that (and loose the ugly code in  gqlRepos schema)

src/api/schemas/gql-repos.js
```diff
@@ -0,0 +1,79 @@
...
+          true
+        ),
+        issues: formatIssues(repository.issues, repository.nameWithOwner),
+        pullRequests: formatIssues(
+          repository.pullRequests,
+          repository.nameWithOwner
+        ),
+        contents_url: `https://api.github.com/repos/${
+          repository.nameWithOwner
+        }/contents`,
```
machour: This is the file I just called ugly. We're receiving a structure that correspond exactly to the current GitHub data model & relation, and I'm just breaking everything to try to mimic the old Rest structure.  Sounds totally illogic.

machour:
Thanks for the feedback buddy. As mentioned in the comments, I think I'll be tweaking the data structure a bit more before merging this one.

Currently experimenting with GraphQL on the Issue/Pull Request screen to gain more hindsight about how we should store the data from v4.

machour:
Reverted to a nicer error screen (the one that was put in place by #726).

Also, as mentioned in the conversation, I've changed the structure of our data in store to closely match the data structure of GitHub & the new Repository screen accept only a `repoId` as entry parameter.

machour:
@housseindjirdeh merged this PR as I had your okay. Changes since your comment only enhanced it.

PR #777, opened 2018-04-19
refactor: translate() to t()

machour:
## Description

Wrote a babel-traverse [script](https://gist.github.com/machour/0739cda4e5905f18c2061287a9b0bc40) that took care of migrating all calls from `translate()` to `t()` swapping keys with strings everywhere.

Then did some polishing for the relativeTimeFromNow() function & fixed a few things in t().

Also removed untranslated english sentences from translation files. 

Should be good to go!

Next thing will be to go over some of the old concatenations we're currently doing and turn those into plain sentences (like I did for the events screen)


src/utils/localization-helper.js
```diff
@@ -21,18 +19,22 @@ export const t = (key, locale, interpolation = null) => {
...
+  let key = 0;
 
   /* eslint-disable no-cond-assign */
   while ((match = componentPlaceholdersReg.exec(translation))) {
+    key += 1;
     ongoing += translation.substring(lastIndex, match.index);
-    if (typeof interpolation[match[2]] === 'undefined') {
-      ongoing += this.config.missingPlaceholder;
-    } else if (typeof interpolation[match[2]] === 'object') {
+    const value = interpolation[match[2]];
```
chinesedfan: You can also add a local variable to save `typeof value`. And how about to replace the while loop with `translation .replace(componentPlaceholdersReg, (match) => { /* blabla */ })`?

machour: Done for the local `typeof value` const.

I didn't know that replace() also accepted a callback as a second parameter, named parameters greatly improve code readability, thank you!

machour:
@chinesedfan could you pull this PR and check if everything is okay for zh? 🙏 
Since this PR touches a lot of files, I'd love to get it merged to avoid breaking it with other merges.

housseindjirdeh:
Beautiful :)

If you would like me to pull it down and review locally as well, let me know - I can do that shortly if need be 🙏

chinesedfan:
@machour I noticed a bug due to `I18n.defaultSeparator === '.'`. Sentences include the dot (like `You may have to request approval for them.`, which appears in the bottom of `auth-profile.screen`) can not be translated correctly.

machour:
@chinesedfan Thank you for spotting this!

Fixed by setting I18n.defaultSeparator = 'GITPOINT-DISABLED';

machour:
@housseindjirdeh yes please! Absolutely nothing should change on the UI side with this PR.

This kind of renders react-native-i18n useless (see #783) and will make switching to lingui-js or other more translator friendly libraries painless in the future :)

I'll add some documentation for translation handling once this is merged.

housseindjirdeh:
Perused through the app, and couldn't find any translation files missing/broken 👍

PR #779, opened 2018-04-22
fix(loading): Fix margins of loading event list item

chinesedfan:
| Devices tested?  | iPhone 6 Simulator |
| Related ticket?  | #765        |

---

## Screenshots

| Before   | After    |
| -------- | -------- |
|||

## Description

@machour Sorry for my late response of #765. Everything is OK, except for a little problem of UI styles and code styles.

According to [styled-components/styled-components#892](https://github.com/styled-components/styled-components/issues/892) and [stackoverflow](https://stackoverflow.com/questions/45121540/animation-not-working-on-release-mode/45127430#45127430), I think the current animated SC implementation is the best.

### DIFF
--- a/src/components/loading-indicators/loading-event-list-item.component.js
+++ b/src/components/loading-indicators/loading-event-list-item.component.js
@@ -5,7 +5,7 @@ import { colors } from 'config';
 import { infiniteAnimation } from 'utils';
 
 const Container = styled.View`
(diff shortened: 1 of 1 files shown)


machour:
Thank you for fixing this glitch!
You also reminded me that we were supposed to be unitless, I'll keep that in mind 🙏

PR #781, opened 2018-04-22
refactor(events): simplify icons code

machour:
## Description

refactored `getIcon()` to keep the code tight

### DIFF
--- a/src/auth/screens/events.screen.js
+++ b/src/auth/screens/events.screen.js
@@ -180,9 +180,11 @@ class Events extends Component {
       } else if (userEvent.action === 'closed') {
         return 'issue-closed';
       }
+
+      return 'issue-opened';
     }
 
-    return 'issue-opened';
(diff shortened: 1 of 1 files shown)


src/auth/screens/events.screen.js
```diff
@@ -155,46 +155,34 @@ class Events extends Component {
...
+
+    if (eventType === 'IssuesEvent') {
+      if (userEvent.action === 'reopened') {
+        return 'issue-reopened';
+      } else if (userEvent.action === 'closed') {
+        return 'issue-closed';
+      }
     }
+
+    return 'issue-opened';
```
chinesedfan: Should the final return statement be in `if (eventType === 'IssuesEvent')`, and the default return value is `null`?

machour: You're absolutely right sir!

Issue #640, opened 2017-11-13
Make user location listitem trigger a search


machour:
## What Currently Happens?

Currently, when you click on the `Location` field in a user profile, a Google Map is opened with a pin to that location.

## What Would You Like To Happen?

I don't think the current behavior is really interesting. Would we consider instead triggering a search for other users in that location?

andrewda:
Hey, that's a really good idea! For anyone who wants to take this on: location searches can be performed with the following query: `location:"San Francisco, CA"` [(try it out)](https://github.com/search?q=location%3A"San+Francisco%2C+CA").

t-laird:
Hi! I previously commented on a different issue but it looked like the person who had claimed it was still in the process of working through a solution. I would love to take a crack at this one, though!

housseindjirdeh:
Thanks @t-laird! Please feel free to take this on <3

With that being said - definitely interested in chatting about the functionality a bit. Is the idea that when we press the location field - it navigates to a separate screen that shows "Other users in the same location"?

t-laird:
Great! My sense of the solution in mind was that clicking the location would navigate to the search screen where there would be an auto populated query with the location as listed under profile info for the user page. That's what I gathered from the previous conversation with @andrewda's suggestion. Here's a screenshot of how it might look, what do you think? 
![simulator screen shot - iphone 6 - 2018-01-31 at 15 20 31](https://user-images.githubusercontent.com/28363671/35650764-7b6acd80-069a-11e8-9232-afaccbe3c48b.png)

housseindjirdeh:
@t-laird I think thats an absolutely great idea :) Definitely nicer than creating a new screen entirely 🙌

skenne21:
Hi there! I’m a first-time contributor and was hoping to help out with this issue. I noticed nobody was assigned to it for a while, but if there’s already a solution in progress I’m happy to try helping out elsewhere. Thanks!

PR #802, opened 2018-05-10
feat: clicking entity location searches for github users matching that location

JSweet314:
| Devices tested?  | iPhone 6/7, Samsung S8 |
| Bug fix?         | yes/**no**      |
| New feature?     | **yes**/no      |
| Includes tests?  | yes/**no**      |
| All Tests pass?  | **yes**/no      |
| Related ticket?  | #640  |

---

## Screenshots

| Before   | After    |
| -------- | -------- |
|before|after|

## Description

Per the suggested solution, we changed the functionality of clicking on an entity's location to navigate to the search screen and initiate a search for users corresponding to the entity's location. This eliminates the need for the Communications.web method to direct the user outside of the application to google maps (deemed not very useful/interesting). Subsequently, we pass down the search query to the search screen from the navigation object's params to display the search term to the user in the RNSearchBar and MNSearchBar components. Rather than TabNavigator, an additional shared StackNavigator route was needed in order to allow for the passing of the params (upgrading the React-Navigation dependency to the current version would eliminate this need, however, that refactor would fall well outside the scope of this issue). You will note the new shared route titled 'SearchUsersByLocation' in the routes.js file. Navigating to this feature in this way also gives the user the added benefit of having a back arrow to return to the entity from which the location was derived.


routes.js
```diff
@@ -206,6 +206,12 @@ const sharedRoutes = {
       title: navigation.state.params.title,
     }),
   },
+  SearchUsersByLocation: {
```
chinesedfan: In order to share this screen for other places, I suggest to define it as `Search` and replace the one in `SearchStackNavigator` with `MainSearch`.

src/components/entity-info.component.js
```diff
@@ -96,7 +97,7 @@ export const EntityInfo = ({ entity, orgs, locale, navigation }: Props) => {
               type: 'octicon',
             }}
             subtitle={entity.location}
-            onPress={() => Communications.web(getLocationLink(entity.location))}
+            onPress={() => getUsersByLocation(entity.location, navigation)}
```
chinesedfan: Let's write the implementation here directly. `getUsersByLocation` sounds like returning users but it is not.

src/search/screens/search.screen.js
```diff
@@ -155,6 +154,19 @@ class Search extends Component {
...
+  componentDidMount() {
+    this.setQueryAsLocation();
+  }
+
+  setQueryAsLocation = () => {
+    const { params } = this.props.navigation.state;
+
+    if (params) {
+      this.switchQueryType(1);
+      this.search(`location:${params.location}`, 1);
```
chinesedfan: Please use `SearchTypes.USERS`, instead of `1`. And you'd better check `params && params.location`.

src/search/screens/search.screen.js
```diff
@@ -294,11 +306,20 @@ class Search extends Component {
       this.getSearchResults().length === 0 &&
       !this.getSearchPagination().isFetching;
 
+    let isLocation = '';
+
+    if (params) {
+      if (params.location) {
+        isLocation = `location:${params.location}`;
```
chinesedfan: `isLocation` usually means a boolean. How about naming it as `searchText`? And we can check in only one if statement, `if (params && params.location)`.

__tests__/tests/components/EntityInfo.js
```diff
@@ -20,13 +20,6 @@ describe('<EntityInfo />', () => {
     expect(wrapper.find({ title: 'Location' }).prop('subtitle')).toEqual(
       organization.location
     );
-    wrapper.find({ title: 'Location' }).simulate('press');
-    expect(Communications.web).toBeCalledWith(
-      `https://www.google.com/maps/place/${organization.location.replace(
-        / /g,
-        '+'
-      )}`
-    );
```
chinesedfan: Try to test your new press action. Mock `navigation.navigate` and check whether it is called with right arguments.

android/gradle/wrapper/gradle-wrapper.properties
```diff
@@ -2,4 +2,4 @@ distributionBase=GRADLE_USER_HOME
 distributionPath=wrapper/dists
 zipStoreBase=GRADLE_USER_HOME
 zipStorePath=wrapper/dists
-distributionUrl=https\://services.gradle.org/distributions/gradle-2.14.1-all.zip
+distributionUrl=https\://services.gradle.org/distributions/gradle-4.3-rc-2-all.zip
```
chinesedfan: Is it really needed? If not, please do not commit, as well as several files like `Cound/Failed/project.pbxproj`.

chinesedfan (changes requested):
Good teamwork! Left some comments for you. Feel free to tell me if you have different opinions.

chinesedfan:
> upgrading the React-Navigation dependency to the current version would eliminate this need, however, that refactor would fall well outside the scope of this issue

By the way, if it didn't introduce too much changes, maybe you can upgrading the dependency.

src/search/screens/search.screen.js
```diff
@@ -155,6 +154,19 @@ class Search extends Component {
...
   }
 
+  componentDidMount() {
+    this.setQueryAsLocation();
+  }
+
+  setQueryAsLocation = () => {
+    const { params } = this.props.navigation.state;
+
+    if (params) {
```
machour: We are already using navigation state in this screen to store/retrieve the searched keyword (see line 38 for example).
Here, it seems that as soon as there's a parameter, you assume that it's related to location, which seems wrong.

machour (changes requested):
Hi @JSweet314! Thank you and other contributors for this PR 🎉 

Aside from what @chinesedfan already said, I left you a comment to further explain a little problem.

Also, I'm noticing two new empty files "Could" and "Failed" which shouldn't have been included in this PR. (joys of bash scripting, heh)

Don't hesitate to reach us on gitter if you need any help at all!

PR #811, opened 2018-09-18
fix: pass props into super() in constructor function

whitedogg13:
| Related ticket?  | #...        |

---

## Description

As in [67a17a](https://github.com/gitpoint/git-point/commit/67a17aedec6195f36997cb24431ae2a72f10a01b), the react components doesn't pass props to super call, which results in TypeError if we need to use props in constructor.

From official [react document](https://reactjs.org/docs/state-and-lifecycle.html#adding-local-state-to-a-class): `Class components should always call the base constructor with props`.  This PR enforces all class components in project to follow this guideline, to avoid potential future bugs.

### DIFF
--- a/src/components/comment-input.component.js
+++ b/src/components/comment-input.component.js
@@ -60,8 +60,8 @@ export class CommentInput extends Component {
     height: number,
   };
 
-  constructor() {
-    super();
+  constructor(props) {
+    super(props);
 
     this.state = {
       text: '',
--- a/src/components/image-zoom.component.js
+++ b/src/components/image-zoom.component.js
@@ -49,8 +49,8 @@ export class ImageZoom extends Component {
     style: Object,
   };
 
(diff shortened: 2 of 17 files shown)


housseindjirdeh:
Thanks for this @whitedogg13 🚀 

But I wouldn't really say it's wrong to not have `props` included in constructors unless we plan on accessing `this.props` 🤔. But if you both think it's nice to just have this always included as a pattern, I'm okay with that!

whitedogg13:
@housseindjirdeh Totally understand 👍 . Just think this simple pattern might be helpful to avoid future potential issues, since React devs normally consider props to be available all over the place in a component.

BTW, thanks you guys for building GitPoint app and make it OSS! Really love this app and use it everyday :)

lex111:
I think we can merge, because there are 2 endorsements from team members.

Issue #528, opened 2017-10-20
Markdown: Add support for <details>


machour:
It would be nice to add support for `<details>` tag in GithubHtmlView:

## Title
<details>
  <summary>
    The following subheadings are hidden by default.
    Click to expand them.
  </summary>

### Sub header 1

Hello there.
</details>

ArvinH:
Hi, I want to take care of this one!
First time for React-native but I'm trying 😄 
![demo](http://g.recordit.co/1EIxv3Kyfr.gif)

machour:
@ArvinH awesome, thank you!! 🎉 
You may find our `<ToggleView />` component handy: https://github.com/gitpoint/git-point/blob/master/src/components/toggle-view.component.js
Could you also try to use the same arrow as github?  ► ▼

housseindjirdeh:
@ArvinH thank you <3

Any questions about anything - don't hesitate to ask 🙌

PR #609, opened 2017-10-30
feat(markdown): Add support for details tag

ArvinH:
feat(markdown): Add support for details tag

| Question         | Response    |
| ---------------- | ----------- |
| Version?         | v1.4.0      |
| Devices tested?  | iPhone 6 |
| Bug fix?         | no    |
| New feature?     | yes      |
| Includes tests?  | no      |
| All Tests pass?  | yes      |
| Related ticket?  | #528      |

---

## Screenshots

| Before   | After    |
| -------- | -------- |
| before | after |

More test cases in [here](https://github.com/machour/git-point-playground/issues/33):

demo

## Description

Add support for github `<details>` markdown tag.
Modify `ToggleView` component for supporting state-based render

PR #819, opened 2018-10-02
feat: support <details> tag for GithubHtmlView (#528)

whitedogg13:
| Devices tested?  | iPhone X |
| Related ticket?  | #528   |

---

## Screenshots

Before (not support details tag)
image

After 1 (detail not expanded)
image

After 2 (detail expanded)
image

## Description

Support `<details>` tag in `GithubHtmlView`, as mentioned in #528 .

First try to find `<summary>` tag inside `<details>`, then render a toggle-able view if found, otherwise do a rollback rendering. Also try to prettify the text in `<summary>` if we got only one text child inside it.

Thanks!

### DIFF
--- a/src/components/toggle-view.component.js
+++ b/src/components/toggle-view.component.js
@@ -6,6 +6,7 @@ export class ToggleView extends Component {
   props: {
     children: any,
     TouchableView: any,
+    renderTouchable: Function,
   };
 
   state: {
@@ -28,7 +29,9 @@ export class ToggleView extends Component {
     return (
       <View>
         <TouchableOpacity onPress={() => this._toggle()}>
-          {this.props.TouchableView}
+          {this.props.renderTouchable
+            ? this.props.renderTouchable(this.state.collapsed)
+            : this.props.TouchableView}
         </TouchableOpacity>
         <Collapsible collapsed={this.state.collapsed}>
           {this.props.children}
--- a/__tests__/tests/components/GithubHtmlView.js
+++ b/__tests__/tests/components/GithubHtmlView.js
@@ -5,7 +5,7 @@ import { GithubHtmlView } from 'components';
 import { Icon } from 'react-native-elements';
 
 describe('<GithubHtmlView />', () => {
-  it('correctly renders <details/> tag in GithubHtmlView', () => {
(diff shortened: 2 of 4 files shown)


lex111:
@whitedogg13 it's awesome! :clap: 

What do you think can write a small test for this component?

whitedogg13:
@lex111 Sure, love to do so. I will update it later.

whitedogg13:
Hi @chinesedfan , thanks for the review!

I made some enhancement over your first and third suggestions:
* enhance toggle-view.component.js to replace WithToggle
* add different edge cases to tests

It's always good to test and reuse code.

However, about the handling for `<summary/>` tag, I have checked #609, some points listed:
* it adds a extra `<hidden/>` tag in order to parse `<summary/>`, since `<hidden/>` is not a valid HTML tag, so more or less like a hack to me. 
* it renders `<summary/>` with `<View>{defaultRenderer(node.children, parent)}</View>`, however, this basically means we don't even need to handle this case since it simply do the default behavior.

Because of above reasons, I think current version makes more sense to me.

Thanks!

whitedogg13:
@chinesedfan Thanks for suggestions, I will update it later :)

whitedogg13:
Hi @chinesedfan , I have pushed the changes, thanks!

lex111:
@machour hi, can we merge this PR?

machour:
Hey @lex111, sorry I've been away from computer these days. Have this PR been tested on Android? If it's working there too then you have my blessing :)

whitedogg13:
Hi @lex111 I have just tested this on Nexus 5 using Genymotion Emulator, and it works correctly. For further PR, I will test it in Android as well :)

src/components/toggle-view.component.js
```diff
@@ -28,7 +29,9 @@ export class ToggleView extends Component {
     return (
       <View>
         <TouchableOpacity onPress={() => this._toggle()}>
-          {this.props.TouchableView}
+          {this.props.renderTouchable
+            ? this.props.renderTouchable(this.state.collapsed)
+            : this.props.TouchableView}
```
housseindjirdeh: 👏

Issue #823, opened 2018-10-04
Languages not in order in settings


SpaceLenore:
## What Currently Happens?

The languages in settings are in a seemingly unordered list and it makes it more difficult to find the language you're looking for.

## What Would You Like To Happen?

The languages should be sorted alphabetically based on their ISO 639-1 code.

## Resolving this issue
I'd like to fix this, I'll be submitting a Pull Request shortly

SpaceLenore:
Merged

PR #824, opened 2018-10-04
Language list in order

SpaceLenore:
Sort languages alphabetically based on ISO 639-1 codes 

| Devices tested?  | Android 8|
| Bug fix?         | no     |
| New feature?     | no       |
| Improvement? | yes |
| Related ticket?  | #823       |

---

## Screenshots

| Before   | After    |
| -------- | -------- |
|ul1|ol1|


src/auth/screens/language-settings.js
```diff
@@ -1,23 +1,48 @@
 export default [
```
chinesedfan: Instead of sorting all languages by hard coding, how about invoking `Array.prototype.sort` at `language-setting.screen.js`? Then we can insert new languages without keeping sorting in mind.

SpaceLenore: Good idea! Added with a new commit.

SpaceLenore:
New commit added that programmatically sorts on name. Thanks for the feedback!

lex111:
@SpaceLenore thank you! :100:

PR #838, opened 2018-10-24
refactor: switch LoadingMembersList to styled components

Chandrasekar-G:
| Devices tested?  | iPhone 6 (iOS 10.3) |
| Related ticket?  | #532        |

---

## Screenshots

| Before   | After    |
| -------- | -------- |
|before|after|

## Description

Migrated LoadingMembersList to styled-components

### DIFF
--- a/src/components/loading-indicators/loading-members-list.component.js
+++ b/src/components/loading-indicators/loading-members-list.component.js
@@ -1,31 +1,33 @@
 import React, { Component } from 'react';
-import { StyleSheet, View, Text, Animated } from 'react-native';
+import { Animated } from 'react-native';
+import styled from 'styled-components';
 
 import { colors, fonts } from 'config';
 import { loadingAnimation } from 'utils';
 
-const styles = StyleSheet.create({
-  wrapper: {
-    marginTop: 15,
-    padding: 15,
-  },
-  avatarContainer: {
-    flex: 1,
-    flexDirection: 'row',
-  },
-  avatar: {
-    borderRadius: 15,
-    height: 30,
-    width: 30,
-    backgroundColor: colors.grey,
-    marginRight: 5,
-  },
-  sectionTitle: {
-    color: colors.black,
-    ...fonts.fontPrimaryBold,
-    marginBottom: 10,
-  },
-});
+const Wrapper = styled.View`
+  margin-top: 15;
+  padding: 15px;
+`;
+
+const SectionTitle = styled.Text`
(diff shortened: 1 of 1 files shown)


src/components/loading-indicators/loading-members-list.component.js
```diff
@@ -47,20 +49,20 @@ export class LoadingMembersList extends Component {
...
 
-        <View style={styles.avatarContainer}>
+        <AvatarContainer>
           {[...Array(10)].map((item, index) => {
             return (
-              <Animated.View
+              <Avatar
                 key={index} // eslint-disable-line react/no-array-index-key
-                style={[styles.avatar, { opacity: this.state.fadeAnimValue }]}
+                style={{ opacity: this.state.fadeAnimValue }}
```
chinesedfan: How about designing `opacity` as a prop of `Avatar`? Then,
```
const Avatar = styled(Animated.View)`
    opacity: ${prop => prop.opacity};
    // blabla
`
```

Chandrasekar-G: @chinesedfan  I pass the prop from the parent component 
`<LoadingMembersList title={t('MEMBERS', locale)} opacity={new Animated.Value(0)}/>`

I use it in Avatar using props
`opacity: ${props => props.opacity};`

I'm able to get the props in the constructor and it is of type _AnimatedValue_

But inside the styled component the props seems to be undefined and hence I get the below exception in the device. Any idea why?

simulator screen shot 26-oct-2018 1 03 26 pm

Chandrasekar-G: @chinesedfan any other changes required here?

chinesedfan: The space between section CONTRIBUTORS and section SOURCE are larger before. Is it normal?

Chandrasekar-G: I tried reverting to master and took a screenshot again. 
before

Not sure what was causing the extra space in the first place. The spacing right now is what is present in the master.

chinesedfan: At first, 

- with master codes, it was always larger space
- with your codes, it was smaller space at the first render. And if visited the second repository, it became larger space

Then I removed all `node_modules` and `yarn` again. It is always larger space, no matter with master or your codes. Well, it's no a big problem, so don't mind about it.

Issue #826, opened 2018-10-06
Some organization page not displayed correctly


garywill:
| Question         | Response    |
| ---------------- | ----------- |
| Version?         | v1.6.0      |
| Devices tested?  | iPhone SE... |

---

## What Currently Happens?

Some problems opening some organization page, like [this page](https://github.com/SimpleNexus), on gitpoint.
1. Opening this page from home screen cause white blank freeze.
1. Opening this page from search won't display repo number.

### 1
1. Open gitpoint and see a news item about [SimpleNexus](https://github.com/SimpleNexus)
1. Touch SimpleNexus
1. White blank freeze

### 2
1. Search for user "SimpleNexus" on gitpoint
1. Touch SimpleNexus
1. Infomation not displayed correctly on that organization page

chinesedfan:
This problem is due to the schema id of orgs is converted to lower case, see `api/shcemas/orgs.js`. I am not sure whether there is something similar existed. Because other entities, like repos/users, are also converted to lower case. A simple fix of this case may like,

```
--- a/src/organization/screens/organization-profile.screen.js
+++ b/src/organization/screens/organization-profile.screen.js
@@ -189,7 +189,7 @@ const mapStateToProps = (state, ownProps) => {
   } = state;
 
   const orgId = ownProps.navigation.state.params.organization.login;
-  const org = orgs[orgId] || ownProps.navigation.state.params.organization;
+  const org = orgs[orgId.toLowerCase()] || ownProps.navigation.state.params.organization;
```

BTW, we need to set `noMembersMessage` of `MemberList` in `organization-profile.screen.js` as well as `repository.screen.js`.

chinesedfan:
@machour Did you remember why those entities idAttribute were converted to lower cases?

machour:
@chinesedfan here's the origin of this piece of code: https://github.com/reduxjs/redux/blob/master/examples/real-world/src/middleware/api.js#L52

machour:
Ideally the navigation parameter would already be in lowercase (it would be the key of the org entity in our store).

chinesedfan:
@machour Thanks for your link. I found Github API accepts case-insensitive arguments and returns case-sensitive logins. For example, following 3 calls will receive the same login `SimpleNexus`.

- https://api.github.com/orgs/SimpleNexus
- https://api.github.com/orgs/simpleNexus
- https://api.github.com/orgs/simplenexus

[...]

PR #856, opened 2018-12-02
feat: Support pagination in issue-list screen and pull-list screen

chinesedfan:
| Version?         | v1.6.0      |
| Devices tested?  | iPhone 6 Simulator |
| Related ticket?  | #751        |

---

## Screenshots

issue-list

## Description

- Support pagination in issue-list screen and pull-list screen
- Clean useless old codes
- Fix bugs caused by store refactoring
  - one is during fetching pull request details, caused by #775
  - stop converting entity ids to lower cases, see https://github.com/gitpoint/git-point/issues/826#issuecomment-434563624, cc @machour

Sadly, the ability to show different colors between closed and merged pull requests, see #852, is broken. Because the search API always returns state `closed`. I will open an issue to record it and hope someone can implement later.


__tests__/tests/reducers/issue.reducer.js
```diff
@@ -23,68 +20,6 @@ describe('Issuer Reducer', () => {
     expect(issueReducer()).toEqual(initialState);
   });
 
-  describe('GET_ISSUE_COMMENTS', () => {
```
housseindjirdeh: Thanks for cleaning this up 🚀

src/api/actions/index.js
```diff
@@ -12,6 +12,7 @@ export const ACTIVITY_WATCH_REPO = createActionSet('ACTIVITY_WATCH_REPO');
 export const ACTIVITY_UNWATCH_REPO = createActionSet('ACTIVITY_UNWATCH_REPO');
 export const SEARCH_REPOS = createPaginationActionSet('SEARCH_REPOS');
 export const SEARCH_USERS = createPaginationActionSet('SEARCH_USERS');
+export const SEARCH_ISSUES = createPaginationActionSet('SEARCH_ISSUES');
```
housseindjirdeh: 👍

src/api/schemas/repos.js
```diff
@@ -48,8 +48,7 @@ export const repoSchema = new schema.Entity(
   'repos',
   {},
   {
-    idAttribute: repo =>
-      (repo.full_name ? repo.full_name : repo.name).toLowerCase(),
+    idAttribute: repo => (repo.full_name ? repo.full_name : repo.name),
```
housseindjirdeh: I'm assuming there's no harm to just keeping the repo name as is (without lowercase) as the idAttribute right? (definitely better this way but just want to double check)

machour: I remember some edge cases where the case was a problem, but definitely can't tell what it was now. 
Let's drop the lowercase and see what happens :)

chinesedfan: @housseindjirdeh  @machour Thanks for your concerns. @machour was inspired from [redux examples](https://github.com/reduxjs/redux/blob/master/examples/real-world/src/middleware/api.js#L52). It reads user input as keyword to fetch Github data. And when finding entities, it also converts the key to lower cases, see [link](https://github.com/reduxjs/redux/blob/master/examples/real-world/src/containers/UserPage.js#L76). Because Github APIs accept case-insensitive arguments and return case-sensitive data, as I commented in https://github.com/gitpoint/git-point/issues/826#issuecomment-434563624.

I think our use cases are a bit different. We always use case-sensitive ids. So we should stop converting entity ids to lower cases.

housseindjirdeh: That clears things up nicely, thank you makes perfect sense :)

PR #875, opened 2019-05-05
chore: Upgrade eslint config

G2Jose:
### Changes
Updated the following packages related to linting: 
* eslint
* babel-eslint

Adds the following packages:
* prettier-eslint

Reformat files that do not comply with linting configuration. 
* The main change is the later version of eslint applies bracket spacing rules to `async` functions too - [Example here](https://github.com/gitpoint/git-point/compare/master...G2Jose:upgrade-eslint-config?expand=1#diff-284817f1b6aad47c0ac6d61bd72f474cR32). 

Rename `.eslintrc` -> `.eslintrc.js` so it can be automatically formatted too.

### Motivation
* Versions of eslint, babel-eslint were outdated
* https://github.com/gitpoint/git-point/pull/874 requires later versions of eslint & babel-eslint
* Using prettier + eslint separately sometimes causes weird conflicts, where eslint formats files one way, followed by prettier formatting it another way while running lint-staged. This was annoying in https://github.com/gitpoint/git-point/pull/874. `prettier-eslint` was created exactly to solve this problem. See https://github.com/prettier/prettier-eslint

---

| Devices tested?  | N/A |
| Related ticket?  | N/A    |


src/repository/screens/commit-list.screen.js
```diff
@@ -24,29 +24,30 @@ class CommitList extends Component {
     return item.id;
   };
 
-  renderItem = ({ item }) =>
-    <CommitListItem commit={item} navigation={this.props.navigation} />;
+  renderItem = ({ item }) => (
+    <CommitListItem commit={item} navigation={this.props.navigation} />
+  );
```
chinesedfan: Can we do some prettier/eslint configurations so that these changes will not be made? As well as the space after `async` and extra empty line between every let/const.

chinesedfan:
Updates,

- Now `prettier-eslint` only supports eslint@4, https://github.com/prettier/prettier-eslint-cli/issues/143
- We need upgrade `eslint-config-airbnb` to use eslint@5 (and remove `eslint-config-airbnb-base`?)
- Some rules are [deprecated](https://eslint.org/docs/rules/#deprecated) or needed to be updated, i.e.

```
-    'newline-after-var': ['error', 'always'],
-    'newline-before-return': 'error',
+    'padding-line-between-statements': [
+        2,
+        { 'blankLine': 'always', 'prev': '*', 'next': 'return' },
+        { 'blankLine': 'always', 'prev': ['var', 'let', 'const'], 'next': '*' },
+        { 'blankLine': 'any', 'prev': ['var', 'let', 'const'], 'next': ['var', 'let', 'const'] },
+    ],
     'newline-per-chained-call': 'off',
     'no-confusing-arrow': 'off',
     'no-mixed-operators': [
@@ -71,7 +75,14 @@ module.exports = {
         allowForLoopAfterthoughts: true,
       },
     ],
-    'space-before-function-paren': ['error', 'never'],
+    'space-before-function-paren': [
+      'error',
+      {
+        anonymous: 'never',
+        named: 'never',
+        asyncArrow: 'always'
+      }
+    ],
```

chinesedfan:
`prettier-eslint-cli@5` released today, with `eslint@5`.

@G2Jose Can you update this PR or would you mind me taking over it?

chinesedfan:
Close in favor of #882. Still many thanks to @G2Jose.

PR #896, opened 2019-10-17
fix: Update locale text for pt-Br

nersoh:
| Version?         |  |
| Devices tested?  | |
| Related ticket?  |         |

---

## Description
Some translation keys for **ptBr** updated.

### DIFF
--- a/src/locale/languages/ptBr.js
+++ b/src/locale/languages/ptBr.js
@@ -14,7 +14,7 @@ module.exports = {
   'Are you sure?': 'Tem certeza?',
   'Assign Yourself': 'Atribua a você',
   Assignees: 'Atribuída a',
-  'Author: ': 'Autor',
+  'Author: ': 'Autor: ',
   BIO: 'BIO',
   CANCEL: 'CANCELAR',
   CONTACT: 'CONTATO',
@@ -164,7 +164,7 @@ module.exports = {
   Users: 'Usuários',
   'View All': 'Ver Todos',
   'View Code': 'Ver Código',
-  'View Commits': '',
+  'View Commits': 'Ver Commits',
   'View and control all of your unread and participating notifications':
     'Ver e controlar todas as suas notificações',
   Watch: 'Acompanhar',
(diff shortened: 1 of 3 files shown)


src/repository/screens/commit.screen.js
```diff
@@ -143,7 +143,7 @@ class Commit extends Component {
         {!!commit.author && (
           <View style={styles.header}>
             <Text>
-              <FieldTitle>{t('Author: ', locale)}</FieldTitle>
+              <FieldTitle>{`${t('Author', locale)}: `}</FieldTitle>
```
chinesedfan: Hi, can you explain a little more why it should be changed? I see keys are 'Author: ' and 'Committer: ' in all local files.

nersoh: Currently, for example, it looks for `'Author: '` (including colon and space) instead of 'Author' , so it doesn't return the correct translation.

nersoh: Changes in this file reverted. 😉

src/locale/languages/ptBr.js
```diff
@@ -14,7 +14,7 @@ module.exports = {
   'Are you sure?': 'Tem certeza?',
   'Assign Yourself': 'Atribua a você',
   Assignees: 'Atribuída a',
-  'Author: ': 'Autor',
+  'Author: ': 'Autor: ',
```
chinesedfan: Yes, this should be the root case. Can you do me a favor to add similar colon and space for 'Author: ' in it/pt/sv and 'Committer: ' in sv?

src/repository/screens/commit.screen.js
chinesedfan: Sorry, seems that this file didn't be reverted.

src/locale/languages/ptBr.js
nersoh: Sure

src/repository/screens/commit.screen.js
nersoh: Please, check if I didn't mess up again 😅

chinesedfan: Everything is good now. 🎉

PR #598, opened 2017-10-28
wip(test): add text normalization test

andrewda:
| Question         | Response    |
| ---------------- | ----------- |
| Version?         | all      |
| Devices tested?  | n/a |
| Bug fix?         | no      |
| New feature?     | no      |
| Includes tests?  | yes      |
| All Tests pass?  | yes*      |
| Related ticket?  | #518        |

---

## Description

Adds a test for the `text-normalize` file. Currently WIP.

PR #899, opened 2019-10-26
test: add tests for normalize-text.js

chinesedfan:
Thanks to the initial work in #598 by @andrewda. This PR uses the same device list and test cases.

And,

- Fixes screens data by,
    - https://www.mydevice.io/
    - https://material.io/resources/devices/
    - https://google.com
- Uses the great [jest.isolateModules(fn)](https://jestjs.io/docs/en/jest-object#jestisolatemodulesfn) to simulate different devices without resetting all mocked modules
- Updates to consider pixelRatio as intervals
    - In fact, `normalize-text.js` was copyed from [react-native-elements/react-native-elements@v1.2.6](https://github.com/react-native-elements/react-native-elements/blob/v1.2.6/src/helpers/normalizeText.js
    ). So this is a sync operation.

Note that not all code branches are tested. I don't know how @andrewda finds those test devices. Maybe they are the most popular ones.

### DIFF
--- a/__tests__/data/screens.js
+++ b/__tests__/data/screens.js
@@ -19,11 +19,11 @@ export const Nexus6P = { width: 411, height: 731 };
 export const Nexus5X = { width: 411, height: 731 };
 export const GooglePixel = { width: 411, height: 731 };
 export const GooglePixelXL = { width: 411, height: 731 };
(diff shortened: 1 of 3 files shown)


andrewda:
Whoo 🎉🎉 thanks for the work on this. I ripped those devices from another repo, but I completely forget where now... Probably should have marked it at the time.

PR #911, opened 2019-11-05
fix: provide real safe area view

chinesedfan:
| Version?         | v1.7.1      |
| Devices tested?  | iPhone 6/Xʀ, Nexus 5X, Pixel 3 XL |
| Related ticket?  | #884, 8839062, 5bdd37e, 161bd78       |

---

## Screenshots

| Nexus 5X (no notch)   | Pixel 3 XL (with notch)    |
| -------- | -------- |
|nexus|pixel|
|nexus3|pixel3|

## Description

- Control all safe area view related screens by `src/config/status-bar.js`
- Force to use react-native-safe-area-view@1.0.0, as it checks by native
    - Good news is that it is compatible with our react-navigation@^1.0.0
    - Use my fork of its peerDependency, react-native-safe-area-context, to support pre-AndroidX

### DIFF
--- a/App.js
+++ b/App.js
@@ -9,7 +9,7 @@ import {
 } from 'react-native';
 import codePush from 'react-native-code-push';
 import { PersistGate } from 'redux-persist/integration/react';
-import { SafeAreaProvider } from 'react-native-safe-area-context';
+import { SafeAreaProvider } from 'react-native-safe-area-context'; // eslint-disable-line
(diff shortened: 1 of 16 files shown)


chinesedfan:
As the fork repository doesn't contain built files, eslint will throw a warning. In fact, never mind it because we only require its `react-native` field, which points to `src/index.tsx`.

